PR #2, opened 2022-01-22
Onboarding Debug Fix

alvind1:


CyStorm:
seems about right. In the real situation I would also check about the return value of the task create functions. Usually, a return status of 0 means success.

PR #3, opened 2022-01-23
updated onboarding challenge

QuintonMak:


CyStorm:
@QuintonMak not quite correct, you missed some things, make sure you read the comments and see what the code is supposed to do

CyStorm:
@QuintonMak take a look at .gitignore files and how to use them, you shouldnt be pushing vscode settings to source control, and be careful of what files you commit. Since you use vs code, i would suggest using the source control GUI it provides, instead of blindly git add *

CyStorm:
very good, but there is even a better way to do this. In the future where adding to .gitignore is not possible, you can place a .gitignore file inside your .vscode folder, and have the content in the gitignore be" * " (without quotes), the star means everything in its directory, that includes itself, this would be the best way to do it. just FYI for next coop or something.. Otherwise it is good.

PR #4, opened 2022-01-31
Fixing 5 bugs

wuda17:


main.c
```diff
@@ -37,7 +39,7 @@ void blinkGIOB1( void * arg )
     for(;;)
     {
         // Toggle on and off at a set speed
-        gioToggleBit(gioPORTB, 1);
+        gioToggleBit(gioPORTA, 1);
```
CyStorm: @wuda17 what is the rationale behind this change? the LED is mapped to GIO port B pin 1.

wuda17: Oh I thought the first LED was mapped to a different port, I didn't know the second argument represented the pin number

CyStorm: yea, it a little bit unclear of the exact usage of these functions without prior knowledge of the systems, but for the purposes of this where it is just looking for C programming fundamentals, it is fine

CyStorm:
Apart of above comment, everything looks about right. the purpose of this exercise is not for exact correctness against some solution, but for general C programming knowledge.

CyStorm:
looks good

PR #7, opened 2022-06-27
Onboarding Challenge Attempt

ziedmasmoudi:


dgobalak (commented):
Looks pretty good so far. There are 2 more errors to find. 

If `BLINK2_DELAY_TICKS` is 400, your program's behaviour for the first 2 cycles is correct. However, for the 3rd and 4th cycles, the LEDs will toggle at the same frequency because they'll have the same delay (400*2 = 800). For those cycles, LED B should be toggled at 4x the frequency of LED A.

The final error is in the include section at the top of the file.

PR #11, opened 2022-09-02
OBC Firmware Onboarding Challenge - roemvaar

roemvaar:
# Purpose
Code implementation for the [OBC Firmware Onboarding Challenge](https://uworbital.notion.site/OBC-Firmware-Onboarding-Challenge-7668fe1436c54bc18a7919ad084ba836)*

**New Changes**

**Things done:**
- HAL: updated the HAL using the provided files on Drive
- Controller: light timer configuration and callback (write to queue)
- Light service: queue creation and handling light message event
- Serial printing: printing to serial 2

**Pending:**
- Light service: read light values from ADCs, send read light values to serial console
- Better error handling 

**Testing (pending)** - once code is completed, builds correctly, and reviewed, I'll test it on the ARM dev board

* This is still work in progress, but I opened so you can have a look at the code that was already implemented. Feedback is always well received.


drivers/source/serial_io.c
```diff
@@ -24,29 +23,43 @@ void sciMutexInit(void) {
     }
     /* USER CODE BEGIN */
     // Create mutex to protect SCI2/SCILin module here.
+    if(sci2Mutex == NULL) {
+        sci2Mutex = xSemaphoreCreateBinary();
```
dgobalak, line 27: I'd suggest using `xSemaphoreCreateMutex` here for a few reasons:
- To stay consistent with `sciMutex`
- As per the [FreeRTOS docs](https://www.freertos.org/xSemaphoreCreateBinary.html#:~:text=Binary%20semaphores%20and,simple%20mutual%20exclusion.), mutexes are usually preferred over binary semaphores when dealing with simple mutual exclusion.

onboarding/source/amb_light_service.c
```diff
@@ -6,34 +6,80 @@
...
+    }
+
+    if (lightServiceQueueHandle == NULL) {
+        lightServiceQueueHandle = xQueueCreate(LIGHT_SERVICE_QUEUE_LENGTH,
+                                               LIGHT_SERVICE_QUEUE_ITEM_SIZE);
+    }
 
     /* USER CODE END */
-    return 1;
+    return 0;
```
dgobalak, line 44: For function returns, we generally use 0 to represent a failure and 1/non-zero to represent success

PR #13, opened 2022-09-12
Initial Onboarding Challenge Submission

nathantg:
# Purpose
My initial submission for the onboarding challenge

Please let me know if there is anything I need to fix :)


onboarding/include/amb_light_service.h
```diff
@@ -15,6 +15,10 @@ typedef enum {
 
 /* USER CODE BEGIN */
 // Define light service queue config here
+#define QUEUE_LENGTH 10
```
dgobalak: You should make these macros more descriptive.

onboarding/source/amb_light_service.c
```diff
@@ -4,9 +4,23 @@
...
+
+
 /* USER CODE BEGIN */
 // Include any additional headers and global variables here
+#include <FreeRTOS.h>
+#include <os_task.h>
+#include <os_queue.h>
+
+static TaskHandle_t lightServiceTaskHandle = NULL;
+static QueueHandle_t xQueue1 = NULL;
```
dgobalak: The name for this variable should be more descriptive.

onboarding/source/amb_light_service.c
```diff
@@ -4,9 +4,23 @@
...
+
 /* USER CODE BEGIN */
 // Include any additional headers and global variables here
+#include <FreeRTOS.h>
+#include <os_task.h>
+#include <os_queue.h>
+
+static TaskHandle_t lightServiceTaskHandle = NULL;
+static QueueHandle_t xQueue1 = NULL; 
+#define Light_Sensor  6U
```
dgobalak: This macro name should be in CAPITAL_SNAKE_CASE.

onboarding/source/amb_light_service.c
```diff
@@ -26,14 +59,58 @@ uint8_t initLightService(void) {
...
     // Wait for MEASURE_LIGHT event in the queue and then print the ambient light value to the serial port.
+    ASSERT(lightServiceTaskHandle != NULL);
+
+    light_event_t event;
+    while(1) {
+        if (xQueueReceive(xQueue1, &event, (TickType_t) 0) == pdPASS) {
+            if(event == MEASURE_LIGHT) {
+                // if event recieved is MEASURE_LIGHT, get value from adc and send over serial
+                uint8_t data = getAmbientLightData();
+                sciPrintText(scilinREG, &data, sizeof(uint8_t));
```
dgobalak: I believe this wouldn't print the correct value. You would be printing the ASCII character that's represented by the number `data`.

For example, let's say we measure the ambient light to be 72. In this case, we want to print 72 over serial. What'll actually happen is you'd print H. Here's an ASCII conversion chart: https://web.alfredstate.edu/faculty/weimandn/miscellaneous/ascii/ascii_index.html

onboarding/source/amb_light_service.c
```diff
@@ -26,14 +59,58 @@ uint8_t initLightService(void) {
...
+    }
 
     /* USER CODE END */
 }
 
 uint8_t sendToLightServiceQueue(light_event_t *event) {
     /* USER CODE BEGIN */
     // Send the event to the queue.
-    
+    if (xQueueSend(xQueue1, (void *) &event, (TickType_t) 0) == pdPASS) {
```
dgobalak: xQueueSend expects a pointer to the event. Since `sendToLightServiceQueue` takes this pointer as an argument, you can just pass `(void *) event` into `xQueueSend`

onboarding/source/controller.c
```diff
@@ -94,6 +120,9 @@ static void ledTimerCallback(TimerHandle_t xTimer) {
 static void lightTimerCallback(TimerHandle_t xTimer) {
     /* USER CODE BEGIN */
     // Send light event to light service queue
+    ASSERT(xTimer != NULL);
+
+    sendToLightServiceQueue(MEASURE_LIGHT);
```
dgobalak: Deal with the case where the event isn't sent to the queue. Printing a message is good enough for this challenge.

onboarding/source/amb_light_service.c
```diff
@@ -26,14 +59,58 @@ uint8_t initLightService(void) {
...
+    adcStartConversion_selChn(adcREG1, Light_Sensor, 6, adcGROUP1);
+
+    while(!adcIsConversionComplete(adcREG1, adcGROUP1));
+
+    adcGetSingleData(adcREG1, adcGROUP1, adc_data_ptr);
+
+    return(adc_data_ptr->value);
+}
+
+void adcGetSingleData(adcBASE_t *adc, unsigned group, adcData_t *data) {
```
dgobalak, line 104: If you're interested, an alternate way to do this ADC conversion is using the included ADC HAL functions which can be found in `hal/source/adc.c`. Some functions that would have been useful are:
- `adcStartConversion`
- `adcGetData`

I'm mentioning this for future reference since it's easier to use the HAL functions than to basically re-implement them. Also, the adc.c file is pretty well-commented, so if you'd like to understand some of this ADC code a bit more, you can check it out. For example, the reason we do `buf & 0xFFFU` is because the RM46 ADC modules are 12-bit.

dgobalak (changes requested):
Looks great! Just have a few minor comments.

onboarding/source/controller.c
dgobalak: Also, you should be passing in a pointer to the event. So have something like `light_event_t event = MEASURE_LIGHT;` and then pass in `&event`.

onboarding/source/amb_light_service.c
dgobalak: Also, the ADC is 12-bit. I.e. you should store the ambient light value in a `uint16_t`.

PR #14, opened 2022-09-15
Second Onboarding Challenge Submission

nathantg:
# Purpose
Second onboarding challenge submission

comments starting with "CHANGE" outlines the changes I made in reaction to the change request comments given in my initial pull request for the challenge


OBC-Firmware-Onboarding.env
```diff
@@ -0,0 +1 @@
+CC_FOLDER_ROOT=C:/Users/natha/GNU-ARM-Embedded-Toolchain/arm-gnu-toolchain-11.3.rel1-mingw-w64-i686-arm-none-eabi
```
dgobalak, line 1: This file should be called ".env", not "OBC-Firmware-Onboarding.env".

The reason for this is that the `Makefile` is dependent on `.env`

onboarding/source/amb_light_service.c
```diff
@@ -4,9 +4,23 @@
...
+#include <os_task.h>
+#include <os_queue.h>
+
+static TaskHandle_t lightServiceTaskHandle = NULL;
+static QueueHandle_t eventQueue = NULL; // CHANGE: more descriptive name for variable
+#define LIGHT_SENSOR  6U // CHANGE: capital
 
+// Helper functions for ADC conversion
+uint16_t getAmbientLightData(void);
+void adcGetSingleData(adcBASE_t *adc, unsigned group, adcData_t *data);
```
dgobalak, line 22: Typically, a function like this would be useful even outside `amb_light_service.c`, so it'd make sense to declare it in a separate header file that can be included by any source files that'd like to use the function. However, this isn't required for the challenge.

If this was a function that was only going to be used in this file (I.e. it's never called outside `amb_light_service.c`), then it should be declared a static function. Again, not necessary to do this for the challenge.

onboarding/source/amb_light_service.c
```diff
@@ -26,14 +59,60 @@ uint8_t initLightService(void) {
...
+    ASSERT(lightServiceTaskHandle != NULL);
+
+    light_event_t event;
+    while(1) {
+        if (xQueueReceive(eventQueue, &event, (TickType_t) 0) == pdPASS) {
+            if(event == MEASURE_LIGHT) {
+                // if event recieved is MEASURE_LIGHT, get value from adc and send over serial
+                uint16_t data = getAmbientLightData(); // CHANGE: data now uint16_t since ADC is 12-bit
+                unsigned char adc_value[10]; 
+                sprintf(adc_value, "%d", data); // CHANGE: convert ADC integer value to char
```
dgobalak, line 71: For future reference, the `snprintf` function is more secure than `sprintf` and should be preferred.
 
The problem with `sprintf` is that it's possible to write more data to the `adc_value` buffer than the size of the buffer itself. `snprintf` makes sure you don't exceed the limit of the buffer. 

However, in this case, your buffer size is 10 and `data` should never be more than 5 digits. So it will still work.

dgobalak (changes requested):

dgobalak:
LGTM. Feel free to pick a task from the CDH task board to work on.

PR #15, opened 2022-09-15
onboarding tasks

Gjjjiang:
# Purpose

# New Changes
- Explain new changes

# Testing

# Outstanding Changes


onboarding/source/amb_light_service.c
```diff
@@ -18,22 +26,91 @@ static void lightServiceTask(void * pvParameters);
...
+                                LIGHT_SERVICE_PRIORITY,     /* Priority at which the task is created. */
+                                &lightServiceTaskHandle);   /* Used to pass out the created task's handle. */
+    }
+
+    if(xReturned == pdFAIL) return 0;
 
+    if (lightServiceQueueHandle == NULL) {
+        lightServiceQueueHandle = xQueueCreate(LIGHT_SERVICE_QUEUE_SIZE,
+                                                LIGHT_SERVICE_ITEM_SIZE);
+    }
```
dgobalak, line 56: Need to check if the queue was created successfully.

onboarding/source/amb_light_service.c
```diff
@@ -18,22 +26,91 @@ static void lightServiceTask(void * pvParameters);
...
+        xReturned = xQueueReceive(lightServiceQueueHandle,
+                                &event,
+                                0);
+        if (xReturned == pdPASS && event == MEASURE_LIGHT) {
+            adcStartConversion(adcREG1, adcGROUP1);
+            adcStartConversion_selChn(adcREG1, LIGHT_SENSOR_PIN, 1, adcGROUP1);
+            while(!adcIsConversionComplete(adcREG1, adcGROUP1));
+            adcGetSingleData(adcREG1, adcGROUP1, lightDataPtr);
+            char *msg = malloc(20 * sizeof(char));
+            sprintf(msg, "light level: %d", lightData.value);
```
dgobalak: `snprintf` is preferred over `sprintf` since it's more secure. It ensures you don't exceed the buffer size.

onboarding/source/amb_light_service.c
```diff
@@ -18,22 +26,91 @@ static void lightServiceTask(void * pvParameters);
...
+    while(1) {
+        xReturned = xQueueReceive(lightServiceQueueHandle,
+                                &event,
+                                0);
+        if (xReturned == pdPASS && event == MEASURE_LIGHT) {
+            adcStartConversion(adcREG1, adcGROUP1);
+            adcStartConversion_selChn(adcREG1, LIGHT_SENSOR_PIN, 1, adcGROUP1);
+            while(!adcIsConversionComplete(adcREG1, adcGROUP1));
+            adcGetSingleData(adcREG1, adcGROUP1, lightDataPtr);
+            char *msg = malloc(20 * sizeof(char));
```
dgobalak: We tend to avoid dynamically allocated memory wherever possible since we want our system to be deterministic. Since `malloc` can fail if we run out of space, it introduces nondeterministic behaviour.

dgobalak: Change `msg` to a stack-allocated array.

dgobalak: For future reference, `malloc` returns a void pointer, so you should cast it to the pointer type you want.

onboarding/source/amb_light_service.c
```diff
@@ -18,22 +26,91 @@ static void lightServiceTask(void * pvParameters);
...
+    BaseType_t xReturned = pdFAIL;
+
+    adcData_t lightData;
+    adcData_t *lightDataPtr = &lightData;
+    while(1) {
+        xReturned = xQueueReceive(lightServiceQueueHandle,
+                                &event,
+                                0);
+        if (xReturned == pdPASS && event == MEASURE_LIGHT) {
+            adcStartConversion(adcREG1, adcGROUP1);
```
dgobalak: I'd recommend creating a helper function to do the ADC conversion. I.e. move this code to its own function.

onboarding/source/amb_light_service.c
```diff
@@ -18,22 +26,91 @@ static void lightServiceTask(void * pvParameters);
...
+
 
     /* USER CODE END */
 }
 
 uint8_t sendToLightServiceQueue(light_event_t *event) {
     /* USER CODE BEGIN */
     // Send the event to the queue.
-    
+    xQueueSend(lightServiceQueueHandle,
```
dgobalak: Return 1 if `xQueueSend` is successful.

onboarding/source/amb_light_service.c
dgobalak: It's a good idea to print a new line after each measurement so you can differentiate between them.

onboarding/source/amb_light_service.c
```diff
@@ -6,7 +6,15 @@
 
 /* USER CODE BEGIN */
 // Include any additional headers and global variables here
+#include "FreeRTOS.h"
+#include "os_task.h"
+#include "os_queue.h"
 
+void adcStartConversion_selChn(adcBASE_t *adc, unsigned channel, unsigned fifo_size, unsigned group);
```
dgobalak: It's clearer if you write the types as `unsigned int` instead of `unsigned`. It's even better if you use the appropriate fixed-size integer types for these arguments (such as `uint8_t` or `uint32_t`).

onboarding/source/controller.c
```diff
@@ -70,15 +76,15 @@ static void controllerTask(void * pvParameters) {
     if (lightServiceStatus == 0) {
         /* USER CODE BEGIN */
         // Deal with error when initializing light service task and/or queue
-
+        sciPrintText(scilinREG, "Error", 6);
```
dgobalak: Add a `\n` at the end of the error message.

onboarding/source/controller.c
```diff
@@ -94,7 +100,9 @@ static void ledTimerCallback(TimerHandle_t xTimer) {
 static void lightTimerCallback(TimerHandle_t xTimer) {
     /* USER CODE BEGIN */
     // Send light event to light service queue
-
+    ASSERT(xTimer != NULL);
+    
+    sendToLightServiceQueue(MEASURE_LIGHT);
```
dgobalak: Print an error message if the `sendToLightServiceQueue` failed

onboarding/source/amb_light_service.c
dgobalak: No need to call both `adcStartConversion` and `adcStartConversion_selChn`. It looks like you're following the example code by TI, so I'd just stick with `adcStartConversion_selChn`.

drivers/source/serial_io.c
```diff
@@ -24,7 +24,9 @@ void sciMutexInit(void) {
     }
```
dgobalak, line 24: There should be some modified files in `hal` once you complete the HALCoGen steps in the challenge

dgobalak (changes requested):
Looks good so far. Just a few changes are needed.

onboarding/source/amb_light_service.c
Gjjjiang: Done

onboarding/source/amb_light_service.c
Gjjjiang: Done

onboarding/source/amb_light_service.c
Gjjjiang: Done

onboarding/source/amb_light_service.c
Gjjjiang: Done

onboarding/source/amb_light_service.c
Gjjjiang: Done

onboarding/source/amb_light_service.c
Gjjjiang: Done

onboarding/source/amb_light_service.c
Gjjjiang: Done

onboarding/source/controller.c
Gjjjiang: Done

onboarding/source/controller.c
Gjjjiang: Done

onboarding/source/amb_light_service.c
Gjjjiang: Done

drivers/source/serial_io.c
Gjjjiang: Oh, I just didn't commit these files.

onboarding/source/amb_light_service.c
```diff
@@ -6,7 +6,19 @@
...
 // Include any additional headers and global variables here
+#include "FreeRTOS.h"
+#include "os_task.h"
+#include "os_queue.h"
 
+#include <string.h>
+#include <stdio.h>
+
+void adcStartConversion_selChn(adcBASE_t *adc, uint8_t channel, uint8_t fifo_size, uint8_t group);
+void adcGetSingleData(adcBASE_t *adc, unsigned group, adcData_t *data);
```
dgobalak: Change this `unsigned` to the appropriate fixed-size integer type as well.

onboarding/source/amb_light_service.c
```diff
@@ -6,7 +6,19 @@
...
+uint16_t getADCConversion();
```
dgobalak: It'd make more sense to call this function something like `getLightValue`.

dgobalak: Since this function is only intended to be used by this source file, declare it a `static` function

dgobalak: In C, when a function takes no arguments, you should write `void` as the argument. For example, something like `uint16_t getADCConversion(void);`

Take a look at this: https://stackoverflow.com/questions/693788/is-it-better-to-use-c-void-arguments-void-foovoid-or-not-void-foo

onboarding/source/amb_light_service.c
```diff
@@ -18,22 +30,101 @@ static void lightServiceTask(void * pvParameters);
...
+    ptr->id    = (unsigned short)((buf >> 16U) & 0x1FU); // int to unsigned short
+
+    adc->GxINTFLG[group] = 9U;
+
+    /**   @note The function canInit has to be called before this function can be used.\n
+    *           The user is responsible to initialize the message box.
+    */
+}
+
+uint16_t getADCConversion(){
```
dgobalak: Write `void` as the argument

dgobalak (changes requested):

onboarding/source/amb_light_service.c
Gjjjiang: Done

onboarding/source/amb_light_service.c
Gjjjiang: Done

Gjjjiang: Done

Gjjjiang: Done

onboarding/source/amb_light_service.c
Gjjjiang: Done

dgobalak:
Great job! I'll close the PR. I'd say you're ready to take on a task. 

If you want, you can try flashing this program onto the RM46 LaunchPad at the weekly meeting or work session.

PR #16, opened 2022-09-20
Fix errors

mangoqwq:
Hope this is correct...


dgobalak:
Looks good! You can work on level2 (the full challenge) now.

PR #19, opened 2022-10-01
Main

prajval-malhotra:
# Purpose

# New Changes
- Explain new changes

# Testing

# Outstanding Changes


drivers/source/serial_io.c
```diff
@@ -24,13 +24,17 @@ void sciMutexInit(void) {
...
-    
+    if (sciLinMutex == NULL) {
+        sciLinMutex = xSemaphoreCreateMutex();
+    }
     /* USER CODE END */
 }
 
 uint8_t sciPrintText(sciBASE_t *sci, unsigned char *text, uint32_t length) {
     /* initSciMutex must be called before printing is allowed */
+    sciMutexInit();
```
dgobalak: This is already called in the main function.

log
```diff
@@ -0,0 +1,142 @@
+hal/source/sys_selftest.c: In function 'fmcECCcheck':
```
dgobalak, line 1: If this is the most up-to-date output when you run `make`, then you'll have to fix all the warnings/errors after line 100.

dgobalak: Also, typically, this file wouldn't be included in the PR. However, it's actually a good idea for the onboarding challenge.

onboarding/include/amb_light_service.h
```diff
@@ -15,7 +17,9 @@ typedef enum {
 
 /* USER CODE BEGIN */
 // Define light service queue config here
-
+// Settings
+static const uint8_t event_queue_len = 10;
```
dgobalak: Instead of storing the parameters in a const variable, define them as a macro like we did for the light service task config above.

onboarding/include/amb_light_service.h
```diff
@@ -15,7 +17,9 @@ typedef enum {
...
+static QueueHandle_t event_queue;
```
dgobalak: This should probably go in the `amb_light_service.c` file.

onboarding/source/amb_light_service.c
```diff
@@ -3,10 +3,17 @@
...
+#include <string.h>
+#include <stdio.h>
+
 
 /* USER CODE BEGIN */
 // Include any additional headers and global variables here
+#include <os_task.h>
+#include <FreeRTOS.h>
 
+#define Light_Sensor 6U
```
dgobalak: This should be in CAPITAL_SNAKE_CASE

onboarding/source/amb_light_service.c
```diff
@@ -15,25 +22,101 @@
  */
 static void lightServiceTask(void * pvParameters);
 
+
+void adcStartConversion_selChn(adcBASE_t *adc, unsigned channel, unsigned fifo_size, unsigned group)
```
dgobalak: Instead of `unsigned`, use `unsigned int` or the appropriate fixed-size integer types (like `uint8_t`)

dgobalak: Our function names are in camelCase

dgobalak: This applies to the rest of your PR as well.

onboarding/source/amb_light_service.c
```diff
@@ -15,25 +22,101 @@
...
+    ptr->id    = (unsigned short)((buf >> 16U) & 0x1FU); // int to unsigned short
+
+    adc->GxINTFLG[group] = 9U;
+
+    /**   @note The function canInit has to be called before this function can be used.\n
+    *           The user is responsible to initialize the message box.
+    */
+}
+
+uint32 getLightSensorData(void) {
```
dgobalak: We typically use `uint32_t`, so I'd use the same here to stay consistent. It's the exact same as `uint32`

onboarding/source/amb_light_service.c
```diff
@@ -15,25 +22,101 @@
...
     return 1;
 }
 
 static void lightServiceTask(void * pvParameters) {
     /* USER CODE BEGIN */
     // Wait for MEASURE_LIGHT event in the queue and then print the ambient light value to the serial port.
-
+    light_event_t ignore = 0;
+    if(xQueueReceive(event_queue, (void *)&ignore, 0) == pdTRUE) {
+        // new event received
```
dgobalak, line 104: Check if the received event was `MEASURE_LIGHT`

onboarding/source/amb_light_service.c
```diff
@@ -15,25 +22,101 @@
...
+        uint32 light_sensor_data = getLightSensorData();
+        unsigned char* ambient_light = (unsigned char*)"";
+        sprintf(ambient_light, "%ld", (long)light_sensor_data);
```
dgobalak: `snprintf` is more secure than `sprintf`

dgobalak: Print a new-line character after the light value. It'll make it easier to read in the serial terminal.

onboarding/source/controller.c
```diff
@@ -92,9 +110,11 @@ static void ledTimerCallback(TimerHandle_t xTimer) {
 }
 
 static void lightTimerCallback(TimerHandle_t xTimer) {
+    ASSERT(xTimer != NULL);
     /* USER CODE BEGIN */
     // Send light event to light service queue
-
+    light_event_t *event = MEASURE_LIGHT;
```
dgobalak: You'll need to first create a variable of type `light_event_t` with a value of `MEASURE_LIGHT`. `sendToLightServiceQueue` expects a pointer to that variable.

`light_event_t event = MEASURE_LIGHT;` and then pass in `&event`

dgobalak (changes requested):
Overall, good work. Just requested a few changes. Also, make sure the program builds without any errors or warnings (aside from the ones in sys_selftest.c).

onboarding/source/amb_light_service.c
```diff
@@ -3,10 +3,18 @@
...
+
 
 /* USER CODE BEGIN */
 // Include any additional headers and global variables here
+#include <os_task.h>
+#include <FreeRTOS.h>
 
+#define LIGHT_SENSOR 6U
+static TaskHandle_t lightServiceTaskHandle = NULL;
+static QueueHandle_t event_queue;
```
dgobalak, line 17: Variable should be camelCase.

onboarding/source/amb_light_service.c
```diff
@@ -15,25 +23,103 @@
...
 
 static void lightServiceTask(void * pvParameters) {
     /* USER CODE BEGIN */
     // Wait for MEASURE_LIGHT event in the queue and then print the ambient light value to the serial port.
-
+    light_event_t event;
+    if(xQueueReceive(event_queue, &event, 0) == pdTRUE) {
+        // new event received
+        if(event == MEASURE_LIGHT) {
+            uint32 Light_Sensor_data = getLightSensorData();
```
dgobalak, line 106: Variable should be camelCase. Also, use uint32_t

onboarding/source/amb_light_service.c
```diff
@@ -15,25 +23,103 @@
...
+            char ambient_light[100];
+            snprintf(ambient_light, strlen(ambient_light), "%lu\n", Light_Sensor_data);
```
dgobalak, line 108: The 2nd argument of snprintf is the size of the buffer (100 in this case). Although, since you're only printing the value, your buffer can have a smaller size, like 10.

onboarding/source/amb_light_service.c
```diff
@@ -15,25 +23,103 @@
...
+            sciPrintText(scilinREG, (unsigned char *)ambient_light, strlen((char *)ambient_light));
+        }
+    }
     /* USER CODE END */
 }
 
 uint8_t sendToLightServiceQueue(light_event_t *event) {
     /* USER CODE BEGIN */
```
dgobalak, line 116: Function should return 1 on success. 0 on fail.

dgobalak (changes requested):
Looks really good! Just a few more changes. Also, consider printing a \n after each error message, so it's easier to read in the serial terminal.

onboarding/source/amb_light_service.c
```diff
@@ -15,25 +23,103 @@
...
     /* USER CODE END */
     return 1;
 }
 
 static void lightServiceTask(void * pvParameters) {
     /* USER CODE BEGIN */
     // Wait for MEASURE_LIGHT event in the queue and then print the ambient light value to the serial port.
-
+    light_event_t event;
+    if(xQueueReceive(event_queue, &event, 0) == pdTRUE) {
```
dgobalak, line 103: You should use a non-zero block time for xQueueReceive. If it's 0, then we'll basically be polling the queue which would result in task starvation if we had lower-priority tasks.

dgobalak (changes requested):

dgobalak:
I'm closing this for now. Send a message in the CDH #onboarding channel if you'd like to get this reviewed again.

PR #23, opened 2022-10-21
Alston's Onboarding Submission

als10:
Onboarding submission for the CDH subteam


onboarding/include/controller.h
```diff
@@ -16,7 +16,9 @@
 
 /* USER CODE BEGIN */
 // define config for the light timer
-
+#define LIGHT_TIMER_NAME          "light_timer"
+#define LIGHT_TIMER_PERIOD        pdMS_TO_TICKS(2500)
```
mangoqwq: We need to make a measurement [every second](https://www.notion.so/uworbital/OBC-Firmware-Onboarding-Challenge-7668fe1436c54bc18a7919ad084ba836#fb4c355b001b495492279b6f9063ec5d).

onboarding/source/controller.c
```diff
@@ -94,7 +110,8 @@ static void ledTimerCallback(TimerHandle_t xTimer) {
 static void lightTimerCallback(TimerHandle_t xTimer) {
     /* USER CODE BEGIN */
     // Send light event to light service queue
-
+    ASSERT(xTimer != NULL);
+    sendToLightServiceQueue(MEASURE_LIGHT);
```
mangoqwq: `sendToLightServiceQueue` expects a pointer, so you should do something like `light_event_t event = MEASURE_LIGHT;` and then pass `&event`.

Also, make sure that the event is successfully added to the queue. Print an error message if it fails!

onboarding/include/amb_light_service.h
```diff
@@ -15,7 +15,8 @@ typedef enum {
 
 /* USER CODE BEGIN */
 // Define light service queue config here
-
+#define QUEUE_SIZE 2
```
mangoqwq: Not sure if this queue size is large enough. Might wanna check with @dgobalak.

onboarding/include/amb_light_service.h
```diff
@@ -15,7 +15,8 @@ typedef enum {
...
+#define QUEUE_ITEM_SIZE sizeof(light_event_t)
```
mangoqwq: Macro names can be more descriptive.

onboarding/source/amb_light_service.c
```diff
@@ -6,7 +6,14 @@
 
 /* USER CODE BEGIN */
 // Include any additional headers and global variables here
+#include <stdio.h>
+#include <math.h>
```
mangoqwq: You should change the import statement ordering to follow the style guide.

> -   Import statments should be grouped in the following order:
>     1.  Local imports (e.g. `#include "cc1120_driver.h`)
>     2.  External library imports (e.g. `#include <semphr.h>`)
>     3.  Standard library imports (e.g. `#include <stdint.h>`)

onboarding/source/amb_light_service.c
```diff
@@ -18,22 +25,63 @@ static void lightServiceTask(void * pvParameters);
...
+    }
     /* USER CODE END */
 }
 
 uint8_t sendToLightServiceQueue(light_event_t *event) {
     /* USER CODE BEGIN */
     // Send the event to the queue.
-    
+    if (lightQueueHandle != NULL) {
+        if (xQueueSend(lightQueueHandle, MEASURE_LIGHT, 1000) != pdPASS) {
```
mangoqwq: Even though `MEASURE_LIGHT` is the only light event we have so far, it would make sense for this to work for any light event, in case we add more in the future.

mangoqwq (commented):
Good work! This is my first review, but I tried to catch everything. Please take a look at the comments.

onboarding/include/amb_light_service.h
dgobalak: A size of 2 should be fine because we're only sending 1 event to the queue every second. As a result, it's unlikely that the queue will be full. However, this does depend on a few things in your implementation. 

For example, let's say we try `xQueueReceive` with a block time of 10 seconds. Then, if there's nothing in the queue, the light service task will enter the blocked state for 10 seconds. In those 10 seconds, the controller task will try and send an event at least 9 times. After 2, the queue will be full. Although, this is an exaggeration since your block time will likely be set to <1 second.

Just to be safe, I'd make the size bigger (like 10).

onboarding/source/amb_light_service.c
dgobalak: I'd recommend using the `pdMS_TO_TICKS` macro to convert the block time from ms to ticks, instead of using the number of ticks directly.

dgobalak (changes requested):

onboarding/source/amb_light_service.c
dgobalak: To add to this, the `xQueueSend` function expects a pointer to the queue item. If you pass in `MEASURE_LIGHT` directly, what'll actually happen is it'll treat that as an address. 

Since `MEASURE_LIGHT` = 0, the program will think the queue item is stored at address 0 which isn't the case. If you try to dereference 0, you'll get a segmentation fault.

As a side note: you never want to implicitly convert between a pointer and integer (like `int *p = 5;`). Most compilers will say that's an error. If you ever need to set a pointer to a specific memory address (like address 0x3011FFE4), then you should do `int *p = (int *) 0x3011FFE4;`.

dgobalak: You could even define this as a macro

als10:
@dgobalak @mangoqwq I've resolved your comments,  could you please take another look at the PR?

dgobalak:
@als10 Looks good to me. You can close this PR now. Can you send me the email you use for GitHub and Notion, so I can add you to our GitHub org and Notion workspace? Just dm me on Discord. My username is `Daniel G#9053`

Once that's done, you can pick up a task from the CDH task board.

PR #30, opened 2023-01-23
Onboarding Challenge

Jac2125:
# Purpose

# New Changes
- Explain new changes

# Testing

# Outstanding Changes


onboarding/source/amb_light_service.c
```diff
@@ -18,22 +28,86 @@ static void lightServiceTask(void * pvParameters);
...
+/**
+ * @brief Send a light service to the queue
+ * @param light_event_t *event
+ **/
+uint8_t sendToLightServiceQueue(light_event_t *event) { 
     /* USER CODE BEGIN */
     // Send the event to the queue.
-    
+    event = &light_event;
+    lightServiceTask(event);
```
Gjjjiang: You should be sending the event to the queue with xQueueSend.

onboarding/source/amb_light_service.c
```diff
@@ -18,22 +28,86 @@ static void lightServiceTask(void * pvParameters);
...
+    */
+	return (adc_data_ptr->value);
+}
+
 static void lightServiceTask(void * pvParameters) {
     /* USER CODE BEGIN */
     // Wait for MEASURE_LIGHT event in the queue and then print the ambient light value to the serial port.
-
+    ASSERT(lightServiceTaskHandle != NULL);
+    uint8_t lightServiceStatus = initLightService();
```
Gjjjiang: this should only be called in the controller task.

onboarding/source/amb_light_service.c
```diff
@@ -18,22 +28,86 @@ static void lightServiceTask(void * pvParameters);
...
+    
+    if(lightServiceStatus == 0){
+        
+        printf("Failed to create a queue");
+        
+    }else{
+
```
Gjjjiang: Need to wait for light event in queue with xQueueReceive

onboarding/source/amb_light_service.c
```diff
@@ -18,22 +28,86 @@ static void lightServiceTask(void * pvParameters);
...
+ 	while(!adcIsConversionComplete(adcREG1, adcGROUP1)); 
+
+	/** - Read the conversion result
+	*     The data contains the Light sensor data
+    */
+	adcGetData(adcREG1, adcGROUP1, adc_data_ptr);
+	
+	/** - Transmit the Conversion data to PC using SCI
+    */
+	return (adc_data_ptr->value);
```
Gjjjiang: Mismatch in type, value is a uint16_t but function type is a ptr.

onboarding/source/amb_light_service.c
```diff
@@ -18,22 +28,86 @@ static void lightServiceTask(void * pvParameters);
...
+    uint8_t lightServiceStatus = initLightService();
+    
+    if(lightServiceStatus == 0){
+        
+        printf("Failed to create a queue");
+        
+    }else{
+        
+        uint32 Light_Sensor_Data = Get_Light_Sensor_data();
+        int length = snprintf(NULL, 0, "&#37;u", Light_Sensor_Data);
```
Gjjjiang: &#37 ; is unicode code for % is used. make sure to use %.

drivers/source/serial_io.c
```diff
@@ -24,7 +24,10 @@ void sciMutexInit(void) {
     }
```
dgobalak, line 24: Looks like there are some build warnings that need to be fixed. Ignore the ones related to the `sys_selftest.c` file.

dgobalak (changes requested):

Gjjjiang (changes requested):
Great start to the onboarding!
Really liked your Get_Light_Sensor_data function
You'll need to revisit some things on tasks.
Just for some clarification, a task is not a function. They are segments of code that run in "parallel" with other tasks, essentially a thread. Information between tasks is communicated over queues.
When you initialize a task (ie TaskCreate), you let FreeRtos know which function (code segment) is going to run in parallel. You only need to do this once.
Make sure to call initLightService in the controller task.

onboarding/source/amb_light_service.c
Gjjjiang, line 91: These lines should be inside an infinite loop.

Gjjjiang (changes requested):

onboarding/source/amb_light_service.c
```diff
@@ -18,22 +28,86 @@ static void lightServiceTask(void * pvParameters);
...
+        char *data = malloc(length + 1);
+        
+        snprintf(data, "&#37;u", Light_Sensor_Data);
+        sciPrintText(scilinREG, data, length + 1);
+        
+        free(data);
+    }
```
Gjjjiang: These lines should be inside a loop

hal/include/queue.h
```diff
@@ -0,0 +1,1277 @@
+/*
```
dgobalak, line 1: Where'd this file come from? Was it generated by HALCoGen?

hal/include/FreeRTOS.h
```diff
@@ -1,1063 +1,1063 @@
-/*
```
dgobalak, line 1: Did you save your HALCoGen project after generating the code? It should modify the `onboarding.dil` and `onboarding.hcg` files.

hal/include/mpu_wrappers.h
```diff
@@ -0,0 +1,134 @@
+/*
```
dgobalak, line 1: Was this file also generated by HALCoGen?

dgobalak (changes requested):

onboarding/include/amb_light_service.h
```diff
@@ -15,7 +15,8 @@ typedef enum {
 
 /* USER CODE BEGIN */
 // Define light service queue config here
-
+#define LIGHT_EVENT_SIZE (sizeof (light_event_t))
+#define QUEUE_SiZE 1
```
dgobalak: This macro name should be more descriptive. Remember that you can include multiple headers in a `.c` file, so if another header defines a macro called `QUEUE_SIZE`, you have a naming conflict. 

To solve this, you either make your macro names very descriptive (like you can enforce a coding standard where all macro names are prefixed by the file name; this would make it unlikely for macros to have the same name) or make them local to a `.c` file (this is how it probably _should_ be done in this case, but don't worry too much about it for this onboarding). For now, just make the macro name more descriptive.

onboarding/source/amb_light_service.c
```diff
@@ -6,7 +6,17 @@
 
 /* USER CODE BEGIN */
 // Include any additional headers and global variables here
+#include <FreeRTOS.h>
+#include <os_task.h>
+#include <stdio.h>
+#include <queue.h>
```
dgobalak, line 12: Where does this header come from? Is it produced by HALCoGen?

onboarding/source/amb_light_service.c
```diff
@@ -18,22 +28,89 @@ static void lightServiceTask(void * pvParameters);
...
+                                &lightServiceTaskHandle);
 
+    }
+
+    if(xReturned == pdFAIL){
+        printf("Unable to create a light service task\n");
+    }
+	
+    if(xLightServiceQueue == NULL){
+        xLightServiceQueue = xQueueCreate(QUEUE_SiZE, LIGHT_EVENT_SIZE);
```
dgobalak: You should check if the queue was created successfully.

onboarding/source/amb_light_service.c
```diff
@@ -18,22 +28,89 @@ static void lightServiceTask(void * pvParameters);
...
+                                LIGHT_SERVICE_NAME,
+                                LIGHT_SERVICE_STACK_SIZE,
+                                NULL,
+                                LIGHT_SERVICE_PRIORITY,
+                                &lightServiceTaskHandle);
 
+    }
+
+    if(xReturned == pdFAIL){
+        printf("Unable to create a light service task\n");
```
dgobalak: `printf` won't work on our microcontroller since there's no way to easily view the output. You need to print via the serial port which is what the `serial_io` module you worked on is for.

dgobalak: No point in trying to create the queue if we failed to create the task. Return 0 on failure.

onboarding/source/amb_light_service.c
```diff
@@ -6,7 +6,17 @@
...
+#include <os_task.h>
+#include <stdio.h>
+#include <queue.h>
+#include <os_queue.h>
+#include <reg_adc.h>
+#include <stdlib.h>
 
+static TaskHandle_t lightServiceTaskHandle = NULL;
+static QueueHandle_t xLightServiceQueue = NULL;
+light_event_t light_event;
```
dgobalak: Variable names are camelCase. You can find the style guide in the README.

onboarding/source/amb_light_service.c
```diff
@@ -18,22 +28,89 @@ static void lightServiceTask(void * pvParameters);
...
+    }
     /* USER CODE END */
     return 1;
 }
 
+/**
+ * @brief Get light data from adcData
+ * @param void
+ **/
+uint16_t Get_Light_Sensor_data(void)
```
dgobalak: Function names are camelCase

onboarding/source/amb_light_service.c
```diff
@@ -18,22 +28,89 @@ static void lightServiceTask(void * pvParameters);
...
+	
+    if(xLightServiceQueue == NULL){
+        xLightServiceQueue = xQueueCreate(QUEUE_SiZE, LIGHT_EVENT_SIZE);
+    }
     /* USER CODE END */
     return 1;
 }
 
+/**
+ * @brief Get light data from adcData
```
dgobalak: We generally add function comments like this in the header file, not the `.c` file.

onboarding/source/amb_light_service.c
```diff
@@ -18,22 +28,89 @@ static void lightServiceTask(void * pvParameters);
...
+ * @param void
```
dgobalak: No point in writing that the parameter is void.

onboarding/source/amb_light_service.c
```diff
@@ -18,22 +28,89 @@ static void lightServiceTask(void * pvParameters);
...
+ **/
+uint16_t Get_Light_Sensor_data(void)
+{
+
+	adcData_t *adc_data_ptr = pvPortMalloc(sizeof(adcData_t));
```
dgobalak: variable name should be camelCase.

dgobalak: We don't use dynamic memory allocation (like `malloc` and even `pvPortMalloc`) in our codebase. We stick with statically allocated code. The main reason is dynamic memory allocation introduces non-deterministic behaviour into the system. `malloc` generally has an unpredictable execution time and it's possible for it to fail (what if we've run out of memory?).

You might be wondering why we're dynamically allocating tasks, queues, timers, and mutexes. That's just how we've left it for our onboarding. In our actual codebase, we use the statically allocated variants of them.

onboarding/source/amb_light_service.c
```diff
@@ -18,22 +28,89 @@ static void lightServiceTask(void * pvParameters);
...
+    adc_data_ptr->id = 6;
+    //adc_data_ptr->value;
```
dgobalak: Remove comment

onboarding/source/amb_light_service.c
```diff
@@ -18,22 +28,89 @@ static void lightServiceTask(void * pvParameters);
...
+}
+
 static void lightServiceTask(void * pvParameters) {
     /* USER CODE BEGIN */
     // Wait for MEASURE_LIGHT event in the queue and then print the ambient light value to the serial port.
-
+    ASSERT(lightServiceTaskHandle != NULL);
+    light_event_t queueTask;
+	
+    while (xQueueReceive(xLightServiceQueue, &queueTask, 0) == pdTRUE) {
```
dgobalak: Although using a block time of 0 here is fine since the tasks have equal priority. Consider the potential issues if the tasks did not have equal priority or if there were other tasks that needed to run.

onboarding/source/amb_light_service.c
```diff
@@ -18,22 +28,89 @@ static void lightServiceTask(void * pvParameters);
...
+	    
+        if (queueTask == MEASURE_LIGHT){
+		
+        	uint16 Light_Sensor_Data = Get_Light_Sensor_data();
```
dgobalak: Use `uint16_t` instead to stay consistent.

onboarding/source/amb_light_service.c
```diff
@@ -18,22 +28,89 @@ static void lightServiceTask(void * pvParameters);
...
+        	int length = snprintf(NULL, 0, "%c;", (char)Light_Sensor_Data);
+        	char *data = malloc(length + 1);
```
dgobalak: You shouldn't use `malloc`. Instead, take a look into using `snprintf` and a statically allocated buffer.

onboarding/source/amb_light_service.c
```diff
@@ -18,22 +28,89 @@ static void lightServiceTask(void * pvParameters);
...
-
+    ASSERT(lightServiceTaskHandle != NULL);
+    light_event_t queueTask;
+	
+    while (xQueueReceive(xLightServiceQueue, &queueTask, 0) == pdTRUE) {
+	    
+        if (queueTask == MEASURE_LIGHT){
+		
+        	uint16 Light_Sensor_Data = Get_Light_Sensor_data();
+        	int length = snprintf(NULL, 0, "%c;", (char)Light_Sensor_Data);
```
dgobalak: This `(char)Light_Sensor_Data` doesn't do what you think. For example, if `Light_Sensor_Data` is 65, then `(char)Light_Sensor_Data` is actually `A`. I'd suggest researching about ASCII codes. You can also play around with casting integers to a `char` in an online C compiler: https://www.programiz.com/c-programming/online-compiler/

onboarding/source/amb_light_service.c
```diff
@@ -18,22 +28,89 @@ static void lightServiceTask(void * pvParameters);
...
+/**
+ * @brief Send a light service to the queue
+ * @param light_event_t *event
+ **/
+uint8_t sendToLightServiceQueue(light_event_t *event) { 
     /* USER CODE BEGIN */
     // Send the event to the queue.
-    
+    event = &light_event;
+    xQueueSend(xLightServiceQueue, event, LIGHT_EVENT_SIZE);
```
dgobalak: Check if the event was sent to the queue successfully.

onboarding/source/amb_light_service.c
```diff
@@ -18,22 +28,89 @@ static void lightServiceTask(void * pvParameters);
...
-uint8_t sendToLightServiceQueue(light_event_t *event) {
+/**
+ * @brief Send a light service to the queue
+ * @param light_event_t *event
+ **/
+uint8_t sendToLightServiceQueue(light_event_t *event) { 
     /* USER CODE BEGIN */
     // Send the event to the queue.
-    
+    event = &light_event;
```
dgobalak: `event` is passed into the function as an argument. You shouldn't set it equal to anything here.

onboarding/source/amb_light_service.c
```diff
@@ -18,22 +28,89 @@ static void lightServiceTask(void * pvParameters);
...
 
-uint8_t sendToLightServiceQueue(light_event_t *event) {
+/**
+ * @brief Send a light service to the queue
+ * @param light_event_t *event
+ **/
+uint8_t sendToLightServiceQueue(light_event_t *event) { 
     /* USER CODE BEGIN */
     // Send the event to the queue.
-
```
dgobalak, line 36: Check if `event` is NULL. In general, if you have a function parameter that's a pointer, then you should be checking if it's NULL inside the function.

onboarding/source/controller.c
```diff
@@ -56,7 +56,18 @@ uint8_t initController(void) {
...
+        // Create light timer
+        lightTimerHandle = xTimerCreate(LIGHT_TIMER_NAME,
+                                        LIGHT_TIMER_PERIOD,
+                                        LIGHT_TIMER_AUTORELOAD,
+                                        (void *) 0,
+                                        lightTimerCallback);
+    }
+    
+    if(xReturned == pdFAIL){
+        printf("Unable to create a task");
```
dgobalak: Use `sciPrintText` and also remember to return 0 if the function failed.

onboarding/source/controller.c
```diff
@@ -70,15 +81,19 @@ static void controllerTask(void * pvParameters) {
     if (lightServiceStatus == 0) {
         /* USER CODE BEGIN */
         // Deal with error when initializing light service task and/or queue
-
+        printf("Failed to create a light service task");
```
dgobalak: Use sciPrintText

onboarding/source/controller.c
```diff
@@ -70,15 +81,19 @@ static void controllerTask(void * pvParameters) {
...
         /* Light service task and queue created successfully */
         BaseType_t xReturned = xTimerStart(ledTimerHandle, 0);
         
         /* USER CODE BEGIN */
         // Start light timer
-
+        xReturned = xTimerStart(lightTimerHandle, 0);
+        
+        if(xReturned == pdFAIL){
+            printf("Timer start failed.");
```
dgobalak: Use sciPrintText

onboarding/source/controller.c
```diff
@@ -94,7 +109,7 @@ static void ledTimerCallback(TimerHandle_t xTimer) {
 static void lightTimerCallback(TimerHandle_t xTimer) {
     /* USER CODE BEGIN */
     // Send light event to light service queue
-
+    sendToLightServiceQueue(MEASURE_LIGHT);
```
dgobalak: The `sendToLightServiceQueue` function expects a pointer. You need to pass in a pointer to a variable that has a value of `MEASURE_LIGHT`.

dgobalak (changes requested):

onboarding/source/amb_light_service.c
Jac2125: Yeah they were generated by HALCoGen

hal/include/queue.h
Jac2125: It was generated by HALCoGen

hal/include/mpu_wrappers.h
Jac2125: It was generated by HALCoGen too

hal/include/FreeRTOS.h
Jac2125: I saved the project after generating the code but I'm not sure. Should I redo the code-generating process so that it can modify the files?

hal/include/mpu_wrappers.h
Jac2125: It was generated by HALCoGen

hal/include/FreeRTOS.h
dgobalak: That's weird lol. Yeah maybe try it again.

onboarding/source/amb_light_service.c
```diff
@@ -18,22 +27,99 @@ static void lightServiceTask(void * pvParameters);
...
+    }
     /* USER CODE END */
     return 1;
 }
 
+uint16_t getLightSensorData(void)
+{
+
+	adcData_t lightData;
+    	lightData.id = 6;
```
dgobalak: So the line `adcStartConversion(adcREG1, adcGROUP1);` actually starts an ADC conversion for all channels in group 1. Remember that you configured the channels in group 1 during the HALCoGen setup. When you call `adcGetData(adcREG1, adcGROUP1, adcDataPtr);`, your buffer `lightData` will be set accordingly.

TLDR: Setting `id` doesn't do anything

onboarding/source/amb_light_service.c
```diff
@@ -18,22 +27,99 @@ static void lightServiceTask(void * pvParameters);
...
+	adcData_t *adcDataPtr = &lightData;
```
dgobalak: You don't need this pointer. You can just pass `&lightData` into `adcGetData` and then at the end return `lightData.value`

onboarding/source/amb_light_service.c
```diff
@@ -18,22 +27,99 @@ static void lightServiceTask(void * pvParameters);
...
     // Wait for MEASURE_LIGHT event in the queue and then print the ambient light value to the serial port.
+    ASSERT(lightServiceTaskHandle != NULL);
+    light_event_t queueTask = NULL_EVENT;
+	
+        if (xQueueReceive(xLightServiceQueue, &queueTask, portMAX_DELAY) == pdTRUE) {
+		
+            if (queueTask == MEASURE_LIGHT){
+            
+            	uint16_t lightSensorData = getLightSensorData();
+            	char data[sizeof(lightSensorData)];
```
dgobalak: `sizeof(lightSensorData)` returns 2 because a `uint16_t` is 2 bytes. Since we'll be printing the actual value, we care about the number of digits in `lightSensorData`. The max value of a `uint16_t` is 65535 (2^16 - 1), so 5 digits. Therefore, to print the number, your buffer should be at least 6 characters (5 for the digits and 1 for the `\0` character). However, you should also be printing a new line and carriage return (`\r\n`) so the values appear on separate lines. So your buffer should be size 8 at minimum. Although, no harm making it a bit bigger (like 10 or 15).

dgobalak: You can initialize the buffer to have all zeros by setting it equal to `{0}`

onboarding/source/amb_light_service.c
```diff
@@ -18,22 +27,99 @@ static void lightServiceTask(void * pvParameters);
...
+            	int length = snprintf(data, sizeof(data), "%c", lightSensorData);
```
dgobalak: You should be checking if this returned value is valid

dgobalak: Since you want to print the number, not the character corresponding to an ASCII code, use `%lu` instead of `%c`. This tells `snprintf` that you want to print a number.

onboarding/source/amb_light_service.c
```diff
@@ -18,22 +27,99 @@ static void lightServiceTask(void * pvParameters);
...
 
+            	sciPrintText(scilinREG, (unsigned char*)data, length + 1);
```
dgobalak: This can lead to u accessing memory u haven't allocated. `snprintf` returns:
```
The number of characters that would have been written if n had been sufficiently large, not counting the terminating null character.
If an encoding error occurs, a negative number is returned.
Notice that only when this returned value is non-negative and less than n, the string has been completely written.
```
`length + 1` can be bigger than the size of your buffer. You don't want that.

dgobalak (changes requested):

onboarding/source/amb_light_service.c
```diff
@@ -18,22 +28,96 @@ static void lightServiceTask(void * pvParameters);
...
+    */
+	return lightData.value;
+}
+
 static void lightServiceTask(void * pvParameters) {
     /* USER CODE BEGIN */
     // Wait for MEASURE_LIGHT event in the queue and then print the ambient light value to the serial port.
+    ASSERT(lightServiceTaskHandle != NULL);
+    light_event_t queueTask = NULL_EVENT;
+
```
dgobalak: Tasks generally shouldn't exit (Notice how the `controllerTask` has a `while(1)` loop at the end). I.e. the function should run forever. 

In this case, imagine the queue is not empty and you've received a MEASURE_LIGHT event. The way this task is implemented, you'll just print the value once and then the function returns. The `controllerTask` is repeatedly sending MEASURE_LIGHT events, so you need to go back to check the queue for events.

onboarding/source/amb_light_service.c
```diff
@@ -18,22 +28,96 @@ static void lightServiceTask(void * pvParameters);
...
+	/** - Transmit the Conversion data to PC using SCI
+    */
+	return lightData.value;
+}
+
 static void lightServiceTask(void * pvParameters) {
     /* USER CODE BEGIN */
     // Wait for MEASURE_LIGHT event in the queue and then print the ambient light value to the serial port.
+    ASSERT(lightServiceTaskHandle != NULL);
+    light_event_t queueTask = NULL_EVENT;
```
dgobalak: This variable name isn't correct. The variable does not store a task. It's just an event sent to the queue. I'd call it `event` or `lightEvent`.

onboarding/source/amb_light_service.c
```diff
@@ -18,22 +28,96 @@ static void lightServiceTask(void * pvParameters);
...
+	
+        if (xQueueReceive(xLightServiceQueue, &queueTask, portMAX_DELAY) == pdTRUE) {
+		
+            if (queueTask == MEASURE_LIGHT){
+            
+            	uint16_t lightSensorData = getLightSensorData();
+            	char data[6] = {0};
```
dgobalak: Create a macro for this buffer size named `LIGHT_DATA_MSG_SIZE` or something along those lines. We usually don't use magic numbers. Either declare a macro or a constant variable. Explan why the size is 6 in a comment or something.

onboarding/source/amb_light_service.c
```diff
@@ -18,22 +28,96 @@ static void lightServiceTask(void * pvParameters);
...
+            	snprintf(data, 8, "%u\r\n", lightSensorData);
```
dgobalak: The size of your buffer is 6, but you're saying you can write up to 8 bytes to the buffer.

onboarding/source/amb_light_service.c
```diff
@@ -18,22 +28,96 @@ static void lightServiceTask(void * pvParameters);
...
     /* USER CODE BEGIN */
     // Send the event to the queue.
-    
+    if(event == NULL){
+        
+        return 0;
+    }
+    BaseType_t xReturned = xQueueSend(xLightServiceQueue, event, LIGHT_EVENT_SIZE);
+
+    if(xReturned == pdPASS){
```
dgobalak, line 119: No need to change this, but another way this could've been done is with a ternary operator. It's basically a 1-line if-else statement. `return (xReturned == pdPASS) ? 1 : 0;`

dgobalak: This is how it works -> `(condition) ? <value if true> : <value if false>`

dgobalak (changes requested):
Just a few more changes to make. The HALCoGen stuff is really weird tho. We'll have to figure out why your generated files are different from what you're supposed to get. You downloaded HALCoGen from here right? https://www.ti.com/tool/HALCOGEN#downloads

onboarding/source/amb_light_service.c
```diff
@@ -18,22 +28,99 @@ static void lightServiceTask(void * pvParameters);
...
+		int dataSize = (int)(ceil(log(lightSensorData))+1);
+            	char data[dataSize];
+            	snprintf(data, dataSize, "%u\r\n", lightSensorData);
 
+            	sciPrintText(scilinREG, (unsigned char*)data, dataSize-1);
+        }  
+      
+    }
+	
+    while(1);
```
dgobalak: Now, if you receive a MEASURE_LIGHT event, you'll print the value once and then enter an infinite loop. Instead, you want to go back to waiting for another MEASURE_LIGHT event. You'll need to place this infinite loop around your if statement.

dgobalak (changes requested):

onboarding/source/amb_light_service.c
```diff
@@ -18,22 +28,99 @@ static void lightServiceTask(void * pvParameters);
...
+    ASSERT(lightServiceTaskHandle != NULL);
+    light_event_t lightEvent = NULL_EVENT;
+	
+        if (xQueueReceive(xLightServiceQueue, &lightEvent, portMAX_DELAY) == pdTRUE) {
+		
+            if (lightEvent == MEASURE_LIGHT){
+            
+            	uint16_t lightSensorData = getLightSensorData();
+		int dataSize = (int)(ceil(log(lightSensorData))+1);
+            	char data[dataSize];
```
dgobalak, line 94: You can't use a non-constant variable to declare an array like this. If you wanted to have a variable-size array, you'd use `malloc`. However, we don't use `malloc` because it introduces some non-deterministic behaviour. Instead, the size of the array should be known at compile time. I.e. `dataSize` must be a constant value. You can just make it big enough to store a 4-digit number with the `\r\n` characters. So a size of 6 is enough. 

Since we don't use magic numbers, you should declare a macro for the size or use a constant variable:
```
#define LIGHT_DATA_STR_LEN 6U
...
char data[LIGHT_DATA_STR_LEN];
```
or
```
const uint8_t lightDataStrLen = 6;
char data[lightDataStrLen];
```

dgobalak (changes requested):

onboarding/source/amb_light_service.c
```diff
@@ -18,22 +28,99 @@ static void lightServiceTask(void * pvParameters);
...
     // Wait for MEASURE_LIGHT event in the queue and then print the ambient light value to the serial port.
+    ASSERT(lightServiceTaskHandle != NULL);
+    light_event_t lightEvent = NULL_EVENT;
+    while(1){
+        if (xQueueReceive(xLightServiceQueue, &lightEvent, portMAX_DELAY) == pdTRUE) {
+		
+            if (lightEvent == MEASURE_LIGHT){
+            
+            	uint16_t lightSensorData = getLightSensorData();
+		const uint8_t dataSize = 6;
```
dgobalak, line 93: I was wrong when I said the length should be 6. Good thing the compiler caught me and gave a warning:
```
onboarding/source/amb_light_service.c: In function 'lightServiceTask':
onboarding/source/amb_light_service.c:95:45: warning: '
   ' directive output may be truncated writing 2 bytes into a region of size between 1 and 5 [-Wformat-truncation=]
   95 |                 snprintf(data, dataSize, "%u\r\n", lightSensorData);
      |                                             ^~~~
onboarding/source/amb_light_service.c:95:17: note: 'snprintf' output between 4 and 8 bytes into a destination of size 6
   95 |                 snprintf(data, dataSize, "%u\r\n", lightSensorData);
```

`dataSize` should actually be 8. A `uint16_t` has a max value of 65535 which is 5 digits (also 5 characters). The `\r\n` is 2 more characters. And then all strings have a null character, `\0`, at the end so 1 more character. `5+2+1=8`, so the `data` array needs to have a size of at least 8.

No need to make this change tho. Just remember to look through all the warnings that the compiler gives in the future when you try to build. It's pretty good at catching some common errors.

Other than that, this PR lgtm. Great job completing the challenge!

PR #32, opened 2023-02-05
Level1 Challenge

SahirRK:


challenge.c
```diff
@@ -50,10 +54,14 @@ uint16_t q3(uint8_t x, uint8_t y) {
 // Note: The array contains 8-bit unsigned integers.
 //-------------------------------------------------------------------------
 int32_t q4(uint8_t * array, uint32_t arrayLength) {
-    for (uint8_t i = 0; i <= arrayLength; i++) {
-        int32_t sum = 0;
+    if (!array) {
+        return -1;
+    }
+    int32_t sum = 0;
+    for (uint8_t i = 0; i < arrayLength; i++) {
```
dgobalak: Your tests were stuck in an infinite loop because of an error on this line. It's subtle, but try and find it. Hint: What's the max value `i` can go to here? What's the max value of `arrayLength`?

dgobalak (changes requested):

challenge.c
SahirRK: Oh I see sorry my bad.

challenge.c
```diff
@@ -101,7 +120,15 @@ int32_t q4(uint8_t * array, uint32_t arrayLength) {
 // Now, x = 10 and y = 5
 //-------------------------------------------------------------------------
 int q9(int *a, int *b) {
+    if (!a || !b) {
+        return 1;
```
dgobalak: At the top of this PR, there's a tab called "Checks". There's a section inside called "Run Tests" which shows the output of all the tests that we run. 
image

There were 3 failed tests. Once you get those tests to pass, you can continue with the rest of the onboarding.

dgobalak (changes requested):

dgobalak:
Great job! You can continue with the rest of the onboarding. I'll close this PR.

PR #34, opened 2023-02-07
Onboarding challenge

SarimMuqeet:
# Purpose

# New Changes
- Explain new changes

# Testing

# Outstanding Changes


onboarding/source/amb_light_service.c
```diff
@@ -18,22 +26,76 @@ static void lightServiceTask(void * pvParameters);
...
+                                LIGHT_SERVICE_STACK_SIZE,
+                                NULL,
+                                LIGHT_SERVICE_PRIORITY,
+                                &xLightServiceHandle
+                            );
+    }
+
+    // Create queue
+    if (xQueue == NULL) {
+        xQueue = xQueueCreate(LIGHT_QUEUE_LENGTH,
```
kiransuren: You likely want to check if the queue was created succesfully (check if the queue is NULL, indicating it failed to allocate the required memory)  https://www.freertos.org/a00116.html

onboarding/source/amb_light_service.c
```diff
@@ -18,22 +26,76 @@ static void lightServiceTask(void * pvParameters);
...
+
+    
     return 1;
 }
 
 static void lightServiceTask(void * pvParameters) {
     /* USER CODE BEGIN */
     // Wait for MEASURE_LIGHT event in the queue and then print the ambient light value to the serial port.
+    adcData_t adc_data;
+    adcData_t *adc_data_ptr = &adc_data;
```
kiransuren: this pointer variable is not necessary, as you can just pass in &adc_data directly

onboarding/source/amb_light_service.c
```diff
@@ -18,22 +26,76 @@ static void lightServiceTask(void * pvParameters);
...
+        if (xQueueReceive(xQueue, NULL, (TickType_t) 11) == pdPASS) {
+            adcStartConversion(adcREG1, 
+                                adcGROUP1);
+
+            while (!adcIsConversionComplete(adcREG1, 
+                                            adcGROUP1));
+
+            adcGetData(adcREG1, 
+                        adcGROUP1, 
+                        adc_data_ptr);
```
kiransuren: Here you need to actually check if you have received the MEASURE_LIGHT event from the queue, instead of just passing NULL to the 2nd argument. I would look at https://www.freertos.org/a00118.html for an example of how this might be implemented.

Only if the event == MEASURE_LIGHT you should complete the adc read and print text to the serial.

onboarding/source/amb_light_service.c
```diff
@@ -18,22 +26,76 @@ static void lightServiceTask(void * pvParameters);
...
 
 static void lightServiceTask(void * pvParameters) {
     /* USER CODE BEGIN */
     // Wait for MEASURE_LIGHT event in the queue and then print the ambient light value to the serial port.
+    adcData_t adc_data;
+    adcData_t *adc_data_ptr = &adc_data;
+
+    // Wait for MEASURE_LIGHT event in the queue and then print the ambient light value to the serial port.
+    while (1) {
+        if (xQueueReceive(xQueue, NULL, (TickType_t) 11) == pdPASS) {
```
kiransuren: Also try not to use magic numbers, instead use macros for things light timeouts. (for this challenge, you can use this existing macro from FreeRTOS called portMAX_DELAY instead of 11)

onboarding/source/amb_light_service.c
```diff
@@ -18,22 +26,76 @@ static void lightServiceTask(void * pvParameters);
...
+    
     /* USER CODE END */
 }
 
 uint8_t sendToLightServiceQueue(light_event_t *event) {
     /* USER CODE BEGIN */
     // Send the event to the queue.
-    
+    // don't need to dereference pointer to event as it takes ptr
+    xQueueSend(xQueue, (light_event_t *) event, (TickType_t) 10);
```
kiransuren: same comment as above for using macros instead of magic numbers

kiransuren: Also in this case, you wouldn't need to cast the event as light_event_t* as the parameter is passed into the function as that already (it would catch any type issues during compilation)

onboarding/source/controller.c
```diff
@@ -56,7 +57,18 @@ uint8_t initController(void) {
...
     /* USER CODE BEGIN */
     // Create light timer and check if task/timers were created successfully
+    if(lightTimerHandle == NULL) {
+        // Create light timer
+        lightTimerHandle = xTimerCreate(LIGHT_TIMER_NAME, 
+                                        LIGHT_TIMER_PERIOD, 
+                                        LIGHT_TIMER_AUTORELOAD, 
+                                        (void *) 0, 
+                                        lightTimerCallback);
+    }
```
kiransuren, line 73: you might want to check if the timer failed to create (check if lightTimerHandle is NULL)

kiransuren (changes requested):
Overall great job! There are a couple things to address but once you've fixed that you'll be on your way!

onboarding/source/amb_light_service.c
```diff
@@ -18,22 +26,76 @@ static void lightServiceTask(void * pvParameters);
...
+                                            adcGROUP1));
+
+            adcGetData(adcREG1, 
+                        adcGROUP1, 
+                        adc_data_ptr);
 
+            unsigned char text[2];
+
+            // LSB by masking with 0xFF
+            uint8_t first_half = (uint16_t) adc_data_ptr->value & 0xFF;
```
dgobalak: `adc_data_ptr->value` returns the ADC value we're interested in. It's a `uint16_t`. It looks like you're separating the upper and lower bytes of the value, casting them to chars, and printing them. For example, imagine `adc_data_ptr->value` is 10. Then, `first_half = 0b1010` and `second_half = 0`. Then, `text[0] = '\n'` and `text[1] = '\0'` as those numbers correspond to these characters in the ASCII table. You'll be printing `\n\0` instead of `10`.

onboarding/source/amb_light_service.c
```diff
@@ -18,22 +26,76 @@ static void lightServiceTask(void * pvParameters);
...
+            uint8_t second_half = ((uint16_t) adc_data_ptr->value >> 8) & 0xFF;
+
+            text[0] = (unsigned char) first_half;
+            text[1] = (unsigned char) second_half;
+
+            sciPrintText(scilinREG,
```
dgobalak: You should also print a new-line after each value so it's easier to read in the terminal.

onboarding/source/amb_light_service.c
dgobalak: Look into using `snprintf` instead.

onboarding/source/controller.c
```diff
@@ -56,7 +57,18 @@ uint8_t initController(void) {
...
 
+    if(xReturned == pdFAIL) {
+        printf("Error - could not create controller task");
```
dgobalak: `printf` doesn't work on our system as we wouldn't be able to see the output. Use the `sciPrintText` function you created instead.

onboarding/source/controller.c
```diff
@@ -70,15 +82,19 @@ static void controllerTask(void * pvParameters) {
     if (lightServiceStatus == 0) {
         /* USER CODE BEGIN */
         // Deal with error when initializing light service task and/or queue
-
+        printf("Error - could not create light service");
```
dgobalak: Use sciPrintText instead

onboarding/source/controller.c
```diff
@@ -70,15 +82,19 @@ static void controllerTask(void * pvParameters) {
...
     } else { 
         /* Light service task and queue created successfully */
         BaseType_t xReturned = xTimerStart(ledTimerHandle, 0);
         
         /* USER CODE BEGIN */
         // Start light timer
+        xReturned = xTimerStart(lightTimerHandle, 0); 
 
+        if(xReturned == pdFAIL) {
+            printf("Error - could not create light timer handle");
```
dgobalak: Use sciPrintText instead

dgobalak (changes requested):
Looks good! Just a few things to fix.

onboarding/source/amb_light_service.c
```diff
@@ -18,22 +26,76 @@ static void lightServiceTask(void * pvParameters);
...
     /* USER CODE END */
+
+    
     return 1;
 }
 
 static void lightServiceTask(void * pvParameters) {
     /* USER CODE BEGIN */
     // Wait for MEASURE_LIGHT event in the queue and then print the ambient light value to the serial port.
+    adcData_t adc_data;
```
dgobalak: Variable names should be `camelCase`

dgobalak (changes requested):

onboarding/source/amb_light_service.c
```diff
@@ -57,34 +56,30 @@ uint8_t initLightService(void) {
 static void lightServiceTask(void * pvParameters) {
     /* USER CODE BEGIN */
     // Wait for MEASURE_LIGHT event in the queue and then print the ambient light value to the serial port.
-    adcData_t adc_data;
-    adcData_t *adc_data_ptr = &adc_data;
+    adcData_t adcData;
+    light_event_t *currentEvent = NULL;
```
dgobalak: `xQueueReceive` will copy the event to the address pointed to by `eventReceived`. So `eventReceived` cannot be NULL. It should point to a valid block of memory that you've allocated. I'd suggest just declaring a `light_event_t` variable like `light_event_t event` and passing `&event` into `xQueueReceive`.

onboarding/source/amb_light_service.c
```diff
@@ -57,34 +56,30 @@ uint8_t initLightService(void) {
...
 
     // Wait for MEASURE_LIGHT event in the queue and then print the ambient light value to the serial port.
     while (1) {
-        if (xQueueReceive(xQueue, NULL, (TickType_t) 11) == pdPASS) {
-            adcStartConversion(adcREG1, 
-                                adcGROUP1);
+        if (xQueueReceive(xQueue, currentEvent, portMAX_DELAY) == pdPASS) {
+            
+            if(currentEvent == MEASURE_LIGHT) {
```
dgobalak, line 66: `currentEvent` is a pointer, so you can't compare it to `MEASURE_LIGHT`. You should dereference the pointer.

However, you shouldn't have a separate pointer variable anyways because of what I mentioned in my other comment.

dgobalak (changes requested):

dgobalak:
Also, as a general note, having a macro named `TEXT_SIZE` isn't very descriptive. But don't worry about it for this challenge.

dgobalak:
Good job! Your submission looks good. You can close this PR now.

PR #35, opened 2023-02-07
Challenge/lvl2

chrisj-28:
# Purpose
Challenge lvl2, Chris Jiang

# New Changes
- Explain new changes

# Testing

# Outstanding Changes


drivers/source/serial_io.c
```diff
@@ -25,11 +25,16 @@ void sciMutexInit(void) {
...
+    if (sciLinMutex == NULL) {
+        sciLinMutex = xSemaphoreCreateMutex();
+    }
+
     /* USER CODE END */
 }
 
 uint8_t sciPrintText(sciBASE_t *sci, unsigned char *text, uint32_t length) {
     /* initSciMutex must be called before printing is allowed */
+    sciMutexInit();
```
dgobalak: No need to call this here. It's called for you in our `main` function.

onboarding/source/amb_light_service.c
```diff
@@ -18,22 +25,57 @@ static void lightServiceTask(void * pvParameters);
...
+    }
+    
+    if (xReturned == pdFAIL) {
+            printf("\nlightServiceTask not created...\n");
+    }
+
+    // Create queue
+    if (lightServiceQueue == NULL) {
+        lightServiceQueue = xQueueCreate(LIGHT_QUEUE_LENGTH, LIGHT_QUEUE_LENGTH);
+    }
```
dgobalak, line 51: Check if the queue was created successfully. It can fail if we've run out of memory.

onboarding/source/amb_light_service.c
```diff
@@ -18,22 +25,57 @@ static void lightServiceTask(void * pvParameters);
...
+        xReturned = xTaskCreate(lightServiceTask,
+                                LIGHT_SERVICE_NAME,
+                                LIGHT_SERVICE_STACK_SIZE,
+                                NULL,
+                                LIGHT_SERVICE_PRIORITY,
+                                &lightServiceHandle);
+    }
+    
+    if (xReturned == pdFAIL) {
+            printf("\nlightServiceTask not created...\n");
```
dgobalak, line 41: Use `sciPrintText` instead of `printf`. `printf` isn't configured to print via the serial port whereas `sciPrintText` is.

onboarding/source/amb_light_service.c
```diff
@@ -18,22 +25,57 @@ static void lightServiceTask(void * pvParameters);
...
+    }
 
     /* USER CODE END */
     return 1;
 }
 
 static void lightServiceTask(void * pvParameters) {
     /* USER CODE BEGIN */
+    adcData_t adc_data;
+    adcData_t *adc_data_ptr = &adc_data;
```
dgobalak: No need to create a separate pointer variable. You can just use `&adc_data` directly.

onboarding/source/amb_light_service.c
```diff
@@ -18,22 +25,57 @@ static void lightServiceTask(void * pvParameters);
...
+        lightServiceQueue = xQueueCreate(LIGHT_QUEUE_LENGTH, LIGHT_QUEUE_LENGTH);
+    }
 
     /* USER CODE END */
     return 1;
 }
 
 static void lightServiceTask(void * pvParameters) {
     /* USER CODE BEGIN */
+    adcData_t adc_data;
```
dgobalak: We use `camelCase` for variable names

onboarding/source/amb_light_service.c
```diff
@@ -18,22 +25,57 @@ static void lightServiceTask(void * pvParameters);
...
+    adcData_t *adc_data_ptr = &adc_data;
+
     // Wait for MEASURE_LIGHT event in the queue and then print the ambient light value to the serial port.
+    while (1) {
+        if (xQueueReceive(lightServiceQueue, NULL, (TickType_t) 11) == pdPASS) {
```
dgobalak: Try not to use magic numbers. Instead, define a macro for this block time or use `portMAX_DELAY`

dgobalak: U need to store the received event somewhere. You should be passing a pointer to a `light_event_t` variable as the 2nd argument

onboarding/source/amb_light_service.c
```diff
@@ -18,22 +25,57 @@ static void lightServiceTask(void * pvParameters);
...
+            adcStartConversion(adcREG1, adcGROUP1);
```
dgobalak: Check that you received a `MEASURE_LIGHT` event before doing this conversion

onboarding/source/amb_light_service.c
```diff
@@ -18,22 +25,57 @@ static void lightServiceTask(void * pvParameters);
...
+        }
+    }
     /* USER CODE END */
 }
 
 uint8_t sendToLightServiceQueue(light_event_t *event) {
     /* USER CODE BEGIN */
     // Send the event to the queue.
-    
+    xQueueSend(lightServiceQueue, (light_event_t*) event, (TickType_t) 10);
```
dgobalak: No need to cast `event` to a `light_event_t *`. Since the type is already `light_event_t *` in the function argument, this is redundant.

onboarding/source/amb_light_service.c
```diff
@@ -18,22 +25,57 @@ static void lightServiceTask(void * pvParameters);
...
+
     // Wait for MEASURE_LIGHT event in the queue and then print the ambient light value to the serial port.
+    while (1) {
+        if (xQueueReceive(lightServiceQueue, NULL, (TickType_t) 11) == pdPASS) {
+            adcStartConversion(adcREG1, adcGROUP1);
+            while (!adcIsConversionComplete(adcREG1, adcGROUP1));
+            adcGetData(adcREG1, adcGROUP1, adc_data_ptr);
+
+            unsigned char txt[2];
+            txt[0] = adc_data.value && 0xFF;
```
dgobalak: adc_data_ptr->value returns the ADC value we're interested in. It's a uint16_t. It looks like you're separating the upper and lower bytes of the value, implicitly casting them to chars when you store them in `txt`, and printing `txt`. For example, imagine adc_data_ptr->value is 10. Then, txt[0] = `0b1010` = '\n' and text[1] = `0` = '\0' as those numbers correspond to these characters in the ASCII table. You'll be printing `\n\0` instead of `10`.

dgobalak: Look into using `snprintf` instead

onboarding/source/amb_light_service.c
dgobalak: Define a macro/constant for the block time or use portMAX_DELAY

onboarding/source/controller.c
```diff
@@ -43,6 +44,10 @@ uint8_t initController(void) {
                                 NULL,                       /* Parameter passed into the task. */
                                 CONTROLLER_PRIORITY,        /* Priority at which the task is created. */
                                 &controllerTaskHandle);     /* Used to pass out the created task's handle. */
+
+        if (xReturned == pdFAIL) {
+            printf("\ncontrollerTask not created...\n");
```
dgobalak, line 49: Use sciPrintText instead

onboarding/source/controller.c
```diff
@@ -56,7 +61,14 @@ uint8_t initController(void) {
...
     /* USER CODE BEGIN */
     // Create light timer and check if task/timers were created successfully
-
+    if (lightTimerHandle == NULL) {
+        // Create light timer
+        lightTimerHandle = xTimerCreate(LIGHT_TIMER_NAME,
+                                    LIGHT_TIMER_PERIOD,
+                                    LIGHT_TIMER_AUTORELOAD,
+                                    (void *) 0,
+                                    lightTimerCallback);
```
dgobalak, line 70: Check if the timer was created successfully.

onboarding/source/controller.c
```diff
@@ -70,14 +82,19 @@ static void controllerTask(void * pvParameters) {
     if (lightServiceStatus == 0) {
         /* USER CODE BEGIN */
         // Deal with error when initializing light service task and/or queue
-
+        printf("Failed to init light service");
```
dgobalak: Use sciPrintText instead

onboarding/source/controller.c
```diff
@@ -70,14 +82,19 @@ static void controllerTask(void * pvParameters) {
...
     } else { 
         /* Light service task and queue created successfully */
         BaseType_t xReturned = xTimerStart(ledTimerHandle, 0);
         
         /* USER CODE BEGIN */
         // Start light timer
+        xReturned = xTimerStart(lightTimerHandle, 0);
+        
+        if (xReturned == pdFAIL) {
+            printf("\nlightTimer not started...\n");
```
dgobalak: Use sciPrintText instead

dgobalak (changes requested):
Great job! A few issues to address, but looks good overall.

onboarding/source/amb_light_service.c
```diff
@@ -52,20 +56,22 @@ uint8_t initLightService(void) {
 
 static void lightServiceTask(void * pvParameters) {
     /* USER CODE BEGIN */
-    adcData_t adc_data;
-    adcData_t *adc_data_ptr = &adc_data;
+    adcData_t adcData;
+    light_event_t *eventReceived = NULL;
```
dgobalak: `xQueueReceive` will copy the event to the address pointed to by `eventReceived`. So `eventReceived` cannot be NULL. It should point to a valid block of memory that you've allocated. I'd suggest just declaring a `light_event_t` variable like `light_event_t event` and passing `&event` into `xQueueReceive`.

dgobalak (changes requested):
One final thing to fix

dgobalak:
Also, as a general note, having a macro named TEXT_SIZE isn't very descriptive. But don't worry about it for this challenge.

dgobalak:
Looks good. You can close this PR now.

PR #41, opened 2023-02-27
onboarding challenge pull request

junxin777:
# Purpose
Complete onboarding challenge


drivers/source/serial_io.c
```diff
@@ -64,16 +72,29 @@ obc_error_code_t sciPrintText(unsigned char *text, uint32_t length) {
 
     /* USER CODE BEGIN */
     // Print text to the serial port using sciSendBytes. Use the mutex to protect the SCI module.
-
+    if (UART_PRINT_REG == scilinREG)
```
dgobalak: We want to be able to print to either serial port, so remove this check.

drivers/source/serial_io.c
```diff
@@ -64,16 +72,29 @@ obc_error_code_t sciPrintText(unsigned char *text, uint32_t length) {
...
+    {
+        if (sciLinMutex != NULL)
```
dgobalak: No need for this check. It's already done near the top of this function.

drivers/source/serial_io.c
```diff
@@ -64,16 +72,29 @@ obc_error_code_t sciPrintText(unsigned char *text, uint32_t length) {
...
+        {
+            if (xSemaphoreTake(sciLinMutex, UART_MUTEX_BLOCK_TIME) == pdTRUE)
```
dgobalak: When printing to the other SCI interface, we should be taking `sciMutex`. Instead of passing in `sciMutex` or `scilinMutex` here, use `mutex`. It's declared near the top of this function. `mutex` is equal to either `sciMutex` or `scilinMutex` depending on what `UART_PRINT_REG` is defined as.

drivers/source/serial_io.c
```diff
@@ -64,16 +72,29 @@ obc_error_code_t sciPrintText(unsigned char *text, uint32_t length) {
...
+            {
+                sciSendBytes(text, length);
+                xSemaphoreGive(sciLinMutex);
+            }
+            return OBC_ERR_CODE_SUCCESS;
```
dgobalak: Only return success if we sent the data. This should be moved to right after `xSemaphoreGive`

drivers/source/serial_io.c
```diff
@@ -64,16 +72,29 @@ obc_error_code_t sciPrintText(unsigned char *text, uint32_t length) {
...
+        }
+    }
+    return 0;
```
dgobalak: Return the relevant error code

drivers/source/serial_io.c
```diff
@@ -19,42 +19,50 @@
...
     /* USER CODE END */
 
     // Static allocation of mutexes can only fail if the buffer is NULL
     configASSERT(sciMutex != NULL);
     configASSERT(sciLinMutex != NULL);
 }
 
-obc_error_code_t sciPrintText(unsigned char *text, uint32_t length) {
+obc_error_code_t sciPrintText(unsigned char *text, uint32_t length)
+{
```
dgobalak, line 65: We usually put the opening curly brace (`{`) on the same line as the function declaration. No need to change this for the onboarding tho.

onboarding/include/amb_light_service.h
```diff
@@ -15,6 +16,15 @@ typedef enum {
...
 obc_error_code_t initLightService(void);
 
+/**
+ * @brief get data from ADC
+ *
+ * @param void
+ * @note use in lightServiceTask(void *pvParameters)
+ * @return uint32, return data value from sensor
+ */
+uint32_t getLightSensorData(void);
```
dgobalak: Nice job making a helper function for this!

onboarding/source/controller.c
```diff
@@ -86,43 +92,71 @@ obc_error_code_t initController(void) {
...
+        sciPrintText((unsigned char *)message, sizeof(message));
         /* USER CODE END */
-    } else { 
+    }
+    else
+    {
         /* Light service task and queue created successfully */
-        BaseType_t xReturned; 
+        BaseType_t xReturned;
         xReturned = xTimerStart(ledTimerHandle, 0);
```
dgobalak, line 130: Check if this succeeded

onboarding/source/amb_light_service.c
```diff
@@ -17,37 +21,126 @@
 
 /* USER CODE BEGIN */
 // Define light service queue config here
+#define QUEUE_LENGTH 10
+#define ITEM_SIZE sizeof(uint64_t)
```
dgobalak: The items sent to the queue have a type of `light_event_t`. You should use `sizeof(light_event_t)` instead

onboarding/source/amb_light_service.c
```diff
@@ -17,37 +21,126 @@
...
 
-static void lightServiceTask(void * pvParameters) {
+uint32_t getLightSensorData(void)
+{
+    adcData_t adc_data;
+
+    /** - Start Group1 ADC Conversion
+     *     Select Channel 6 - Light Sensor for Conversion
+     */
+    adcInit();
```
dgobalak: This is called in our `main` function. So you can remove it from here.

onboarding/source/amb_light_service.c
```diff
@@ -17,37 +21,126 @@
...
+    }
+
+    return OBC_ERR_CODE_SUCCESS;
     /* USER CODE END */
 }
 
-static void lightServiceTask(void * pvParameters) {
+uint32_t getLightSensorData(void)
+{
+    adcData_t adc_data;
```
dgobalak, line 85: Variables should be `camelCase`

onboarding/source/amb_light_service.c
```diff
@@ -17,37 +21,126 @@
...
+    {
+        return OBC_ERR_CODE_QUEUE_CREATION_FAILED;
+    }
+
+    return OBC_ERR_CODE_SUCCESS;
     /* USER CODE END */
 }
 
-static void lightServiceTask(void * pvParameters) {
+uint32_t getLightSensorData(void)
```
dgobalak: `adc_data.value` is a `uint16_t`, so that should be the return type.

onboarding/source/amb_light_service.c
```diff
@@ -17,37 +21,126 @@
...
+    /** - Wait for ADC Group1 conversion to complete */
+
+    adcStartConversion(adcREG1, adcGROUP1);
+    /** - Read the conversion result
+     *     The data contains the Light sensor data
+     */
+    while (!adcIsConversionComplete(adcREG1, adcGROUP1))
+        ;
+
+    /** - Transmit the Conversion data to PC using SCI
```
dgobalak: It'd make more sense to move this comment to line 122 or just remove it altogether.

onboarding/source/amb_light_service.c
```diff
@@ -17,37 +21,126 @@
...
+    {
+        light_event_t lightEvent;
+        while (1)
+        {
+            // queueBuffer has different type with event, will it convert automatically?
+            if (xQueueReceive(queueHandle, &lightEvent, (TickType_t)0) == pdTRUE)
+            {
+                if (lightEvent == MEASURE_LIGHT)
+                {
+                    uint32_t lightData = getLightSensorData();
```
dgobalak: This should be `uint16_t`

onboarding/source/amb_light_service.c
```diff
@@ -17,37 +21,126 @@
...
+                    sciPrintf("%u\n", lightData);
```
dgobalak, line 115: For future reference, if `lightData` is a `uint32_t`, you should use the `%lu` format specifier.

onboarding/source/amb_light_service.c
```diff
@@ -17,37 +21,126 @@
...
+obc_error_code_t sendToLightServiceQueue(light_event_t *event)
+{
     /* USER CODE BEGIN */
     // Send the event to the queue. Return error code if event was not sent successfully.
-    
+    if (queueHandle != NULL)
+    {
+        if (xQueueSend(queueHandle, event, (TickType_t)0) == pdTRUE)
+        {
+            // Are there any other issue that may cause queuesend fail?
```
dgobalak: In this specific program, the queue will probably never get full. However, you should always check if you sent the event successfully (like you've done here). If the queue is full, then `xQueueSend` would return false. It's good that you returned `OBC_ERR_CODE_QUEUE_SEND_FAILED` in that case.

dgobalak (changes requested):
This is really well done! Just had a few comments.

onboarding/include/amb_light_service.h
junxin777: Thanks!!

drivers/source/serial_io.c
```diff
@@ -65,15 +73,24 @@ obc_error_code_t sciPrintText(unsigned char *text, uint32_t length) {
...
     // Print text to the serial port using sciSendBytes. Use the mutex to protect the SCI module.
 
-    /* USER CODE END */
+    if (xSemaphoreTake(mutex, UART_MUTEX_BLOCK_TIME) == pdTRUE)
+    {
+        sciSendBytes(text, length);
+        xSemaphoreGive(sciLinMutex);
+        return OBC_ERR_CODE_SUCCESS;
+    }
 }
```
dgobalak, line 88: This is where the build error comes from. If we fail to take the mutex, you should return an error code.

drivers/source/serial_io.c
```diff
@@ -65,15 +73,24 @@ obc_error_code_t sciPrintText(unsigned char *text, uint32_t length) {
...
 
-obc_error_code_t sciPrintf(const char *s, ...){
+return OBC_ERR_CODE_SCI_SEND_FAILED;
```
dgobalak: Is this supposed to be the error code I talked about in my other comment? Instead of returning this, it'd be better to return `OBC_ERR_CODE_MUTEX_TIMEOUT` since, if `xSemaphoreTake` fails, it means we timed out when trying to acquire the mutex.

dgobalak (changes requested):

common/include/obc_errors.h
```diff
@@ -14,12 +15,16 @@ typedef enum {
     OBC_ERR_CODE_TASK_CREATION_FAILED = 102,
     OBC_ERR_CODE_TIMER_CREATION_FAILED = 103,
     OBC_ERR_CODE_MUTEX_CREATION_FAILED = 104,
-
+    OBC_ERR_CODE_QUEUE_CREATION_FAILED = 105,
+    OBC_ERR_CODE_TIMER_START_FAILED = 106,
+    OBC_ERR_CODE_QUEUE_SEND_FAILED = 107,
+    OBC_ERR_CODE_MUTEX_TAKE_FAILED = 108,
```
dgobalak: No need for this one. We already have an error code for it called `OBC_ERR_CODE_MUTEX_TIMEOUT`

common/include/obc_errors.h
```diff
@@ -14,12 +15,16 @@ typedef enum {
     OBC_ERR_CODE_TASK_CREATION_FAILED = 102,
     OBC_ERR_CODE_TIMER_CREATION_FAILED = 103,
     OBC_ERR_CODE_MUTEX_CREATION_FAILED = 104,
-
+    OBC_ERR_CODE_QUEUE_CREATION_FAILED = 105,
+    OBC_ERR_CODE_TIMER_START_FAILED = 106,
+    OBC_ERR_CODE_QUEUE_SEND_FAILED = 107,
```
dgobalak: No need for this error code. We already have one for it called `OBC_ERR_CODE_QUEUE_FULL`

drivers/source/serial_io.c
```diff
@@ -65,15 +73,29 @@ obc_error_code_t sciPrintText(unsigned char *text, uint32_t length) {
...
+        return OBC_ERR_CODE_MUTEX_TAKE_FAILED;
+    }
+    else
+    {
+        sciSendBytes(text, length);
+        xSemaphoreGive(sciLinMutex);
+        return OBC_ERR_CODE_SUCCESS;
+    }
+
+    return OBC_ERR_CODE_MUTEX_TIMEOUT;
```
dgobalak, line 87: This line is never reached because of the return in your else statement. In general, if you're returning inside your `if` statement, there's no need for an `else`.

onboarding/source/amb_light_service.c
```diff
@@ -17,37 +21,119 @@
...
+            // should I check if the queue is full?
+            return OBC_ERR_CODE_SUCCESS;
+        }
+    }
+    return OBC_ERR_CODE_QUEUE_SEND_FAILED;
```
dgobalak: Remember to change this error code. I forgot we had an error code for when the queue is full.

onboarding/source/amb_light_service.c
```diff
@@ -17,37 +21,119 @@
...
     /* USER CODE END */
 }
 
-obc_error_code_t sendToLightServiceQueue(light_event_t *event) {
+obc_error_code_t sendToLightServiceQueue(light_event_t *event)
+{
     /* USER CODE BEGIN */
     // Send the event to the queue. Return error code if event was not sent successfully.
-    
+    if (queueHandle != NULL)
```
dgobalak: If `queueHandle` is `NULL`, we should return a more appropriate error code than just `OBC_ERR_CODE_QUEUE_SEND_FAILED` or `OBC_ERR_CODE_QUEUE_FULL`. Return something like `OBC_ERR_CODE_INVALID_STATE` instead.

dgobalak: Also, instead of using a nested if statement here, it's more clear to write something like:
```
if (queueHandle == NULL)
    return OBC_ERR_CODE_INVALID_STATE;
    
if (xQueueSend(queueHandle, event, (TickType_t)0) == pdTRUE)
    return OBC_ERR_CODE_SUCCESS;

return OBC_ERR_CODE_QUEUE_FULL;
```

dgobalak (changes requested):
Looks good! Just made a few more comments about the error codes.

dgobalak (changes requested):
Looks good! I'll close this PR now

PR #42, opened 2023-03-03
Jerry firmware onboarding

Jitao-Hu:
# Purpose
CDH firmware onboarding: https://uworbital.notion.site/Firmware-Onboarding-Challenge-b208ef6d62cd436a8e738dc99f6d2c4d#c1b5ac11629547669eb9de789797066c

# New Changes
- Followed the firmware onboarding guide above and completed the challenge

# Testing
- Hope to flash the code on the board

# Outstanding Changes


onboarding/source/amb_light_service.c
```diff
@@ -34,20 +44,81 @@ static void lightServiceTask(void * pvParameters);
...
+    {
+        light_event_t event;
+        while (1)
+        {
+            if (xQueueReceive (lightServiceQueueHandle, 
+                                &event,
+                                ( TickType_t) 0) == pdTRUE)
+            {
+                adcStartConversion(adcREG1, adcGROUP1);
+                while(1)
```
Gjjjiang: No need for the second loop. We only want to print light value when we receive a light event

Gjjjiang (commented):
Just one small change to the light task, otherwise, the rest looks good to me.

onboarding/source/amb_light_service.c
```diff
@@ -34,20 +44,78 @@ static void lightServiceTask(void * pvParameters);
...
+    if (lightServiceQueueHandle != NULL)
+    {
+        light_event_t event;
+        while (1)
+        {
+            if (xQueueReceive (lightServiceQueueHandle, 
+                                &event,
+                                ( TickType_t) 0) == pdTRUE)
+            {
+                adcStartConversion(adcREG1, adcGROUP1);
```
dgobalak: Check if the received event is `MEASURED_LIGHT`

dgobalak (changes requested):
Looks good! Just 1 comment.

onboarding/source/amb_light_service.c
Jitao-Hu: Thank you for the feedback. Just updated it.

onboarding/source/amb_light_service.c
```diff
@@ -34,20 +44,78 @@ static void lightServiceTask(void * pvParameters);
...
     /* USER CODE BEGIN */
     // Wait for MEASURE_LIGHT event in the queue and then print the ambient light value to the serial port.
+    if (lightServiceQueueHandle != NULL)
+    {
+        light_event_t event;
+        while (1)
+        {
+            if (xQueueReceive (lightServiceQueueHandle, 
+                                &event,
+                                ( TickType_t) 0) == pdTRUE && event == MEASURE_LIGHT)
```
dgobalak: This is a little sketchy since `event` gets updated in the first part of the condition. I.e. changing this to `if (event == MEASURE_LIGHT && QueueReceive (lightServiceQueueHandle, &event, ( TickType_t) 0))` would yield different results from what you have. Also, this wouldn't scale well if we added new events, not just `MEASURE_LIGHT`. 

Instead, add a separate if statement to check if `event == MEASURE_LIGHT`

onboarding/source/amb_light_service.c
```diff
@@ -34,20 +44,78 @@ static void lightServiceTask(void * pvParameters);
...
+            {
+                adcStartConversion(adcREG1, adcGROUP1);
+                if (adcIsConversionComplete(adcREG1, adcGROUP1) == 8)
```
dgobalak: For the challenge, this is fine. For future reference, there's an enum defined in `adc.h` you can use here instead of 8. It has a value called `ADC_CONVERSION_IS_FINISHED` which equals 8

onboarding/source/amb_light_service.c
```diff
@@ -34,20 +44,78 @@ static void lightServiceTask(void * pvParameters);
...
+    if (lightServiceQueueHandle != NULL)
+    {
+        light_event_t event;
+        while (1)
+        {
+            if (xQueueReceive (lightServiceQueueHandle, 
+                                &event,
+                                ( TickType_t) 0) == pdTRUE && event == MEASURE_LIGHT)
+            {
+                adcStartConversion(adcREG1, adcGROUP1);
```
dgobalak: For future reference, it'd be better to move all the ADC-related stuff into a helper function. This would make it easier to test and debug your code.

onboarding/source/amb_light_service.c
```diff
@@ -34,20 +44,78 @@ static void lightServiceTask(void * pvParameters);
...
+        }
+        
+    }
     /* USER CODE END */
 }
 
 obc_error_code_t sendToLightServiceQueue(light_event_t *event) {
     /* USER CODE BEGIN */
     // Send the event to the queue. Return error code if event was not sent successfully.
+    if (lightServiceQueueHandle != NULL)
```
dgobalak, line 124: Check if `event` is NULL

onboarding/source/amb_light_service.c
```diff
@@ -34,20 +44,78 @@ static void lightServiceTask(void * pvParameters);
...
+    {
+        if (xQueueSend( lightServiceQueueHandle,
+                        event,
+                        ( TickType_t ) 0 ) != pdTRUE)
+        {
+            return OBC_ERR_CODE_QUEUE_SEND_FAILED;
+        } 
+        return OBC_ERR_CODE_SUCCESS;
+    }
+    return OBC_ERR_CODE_INVALID_ARG;
```
dgobalak: If `lightServiceQueueHandle == NULL`, it'd make more sense to return an error code like `OBC_ERR_CODE_INVALID_STATE` since it'd mean `initLightService` wasn't called successfully. `OBC_ERR_CODE_INVALID_ARG` is used when a function argument is invalid.

dgobalak (changes requested):
Just a few more comments

dgobalak (commented):
Looks good to me! I'll close this PR now

PR #44, opened 2023-03-09
Kaitlyn's onboarding challenge

kaitlynwesting:
# Purpose
Completed [firmware onboarding challenge](https://uworbital.notion.site/Firmware-Onboarding-Challenge-b208ef6d62cd436a8e738dc99f6d2c4d).

# New Changes
N/A

# Testing
N/A

# Outstanding Changes
N/A


onboarding/source/amb_light_service.c
```diff
@@ -35,19 +48,86 @@ obc_error_code_t initLightService(void) {
...
+
+    if (lightServiceQueueHandle == NULL)
+        return OBC_ERR_CODE_QUEUE_CREATION_FAILED;
+
+    return OBC_ERR_CODE_SUCCESS;
+
     /* USER CODE END */
 }
 
+uint32_t getLightSensorData(void) {
```
dgobalak: Great job making a helper function for this!

It'd make sense to declare this a `static` function. That would make it accessible only to functions in the same compilation unit. A compilation unit is _basically_ just a source file with the content of its included header files copied into it. You can read more about them [here](https://www.cs.auckland.ac.nz/references/unix/digital/AQTLTBTE/DOCU_015.HTM)). For example, a static function in this `.c` source file would only be accessible by other functions in this file.

Functions, by default (i.e. without the `static` keyword), are accessible in other compilation units. However, if this is your intention, you'd typically put its function declaration in a header file (so `amb_light_service.h` in this case).

onboarding/source/amb_light_service.c
```diff
@@ -35,19 +48,86 @@ obc_error_code_t initLightService(void) {
...
+ 	while(!adcIsConversionComplete(adcREG1, adcGROUP1)); 
+
+	/** - Read the conversion result
+	*     The data contains the Light sensor data
+    */
+	adcGetData(adcREG1, adcGROUP1, &lightData);
+	
+	/** - Transmit the Conversion data to PC using SCI
+    */
+	return (lightData.value);
```
dgobalak, line 102: `lightData.value` is a `uint16_t` type (You can find the `adcData_t` struct definition in the `adc.h` file). So you should change this function's return type accordingly.

onboarding/source/amb_light_service.c
```diff
@@ -35,19 +48,86 @@ obc_error_code_t initLightService(void) {
...
 static void lightServiceTask(void * pvParameters) {
     /* USER CODE BEGIN */
     // Wait for MEASURE_LIGHT event in the queue and then print the ambient light value to the serial port.
 
+    light_event_t lightEvent; // Buffer into which the event will be copied
+    
+    if (lightServiceQueueHandle != NULL) {
+        if (xQueueReceive(lightServiceQueueHandle, &lightEvent, (TickType_t) 10) == pdPASS) {
+            if (lightEvent == MEASURE_LIGHT) {
+                    uint32_t lightSensorData = getLightSensorData();
```
dgobalak: Remember to change the type of `lightSensorData` once u change the return type of `getLightSensorData`

onboarding/source/amb_light_service.c
```diff
@@ -35,19 +48,86 @@ obc_error_code_t initLightService(void) {
...
+
+                    sciPrintf("Ambient light value measured: %u", lightSensorData);
```
dgobalak: For future reference, the format specifier for `uint32_t` values is `%lu`, not `%u`.

dgobalak: You should print a new line after this so it's easier to read once it gets printed out.

onboarding/source/amb_light_service.c
```diff
@@ -35,19 +48,86 @@ obc_error_code_t initLightService(void) {
...
+            }
+        }
+    }
     /* USER CODE END */
 }
 
 obc_error_code_t sendToLightServiceQueue(light_event_t *event) {
     /* USER CODE BEGIN */
     // Send the event to the queue. Return error code if event was not sent successfully.
```
dgobalak: Check if `event` is `NULL`

onboarding/source/amb_light_service.c
```diff
@@ -35,19 +48,86 @@ obc_error_code_t initLightService(void) {
...
     
+    if (xQueueSend(lightServiceQueueHandle, event, (TickType_t) 10) == pdPASS) {
+        return OBC_ERR_CODE_SUCCESS;
+    }
+
+    return OBC_ERR_CODE_SCI_SEND_FAILED;
```
dgobalak: `OBC_ERR_CODE_SCI_SEND_FAILED` is used if we failed to print via the serial interface. In this case, if `xQueueSend` fails, it'd mean the queue was full. So returning `OBC_ERR_CODE_QUEUE_FULL` instead would make sense.

onboarding/source/controller.c
```diff
@@ -99,6 +117,8 @@ static void controllerTask(void * pvParameters) {
         /* USER CODE BEGIN */
         // Deal with error when initializing light service task and/or queue
 
+        sciPrintf("ERROR: Failed to initialize light service task and/or queue");
```
dgobalak: For future reference, you should print a `\n` after these error messages so they're easier to view. Not a big deal tho, so dw about it for this challenge

dgobalak (changes requested):
Great job! The code looks good; just had a few comments.

onboarding/source/amb_light_service.c
```diff
@@ -35,19 +48,87 @@ obc_error_code_t initLightService(void) {
...
+        }
+    }
     /* USER CODE END */
 }
 
 obc_error_code_t sendToLightServiceQueue(light_event_t *event) {
     /* USER CODE BEGIN */
     // Send the event to the queue. Return error code if event was not sent successfully.
-    
+    if (event != NULL) {
```
dgobalak: If `event == NULL`, you should return `OBC_ERR_CODE_INVALID_ARG`.

dgobalak (changes requested):
1 small change. Other than that, looks good! After making the change, I'll close this PR

dgobalak (commented):
Perfect, great job! I'll close this PR now

PR #45, opened 2023-05-08
Asmi - Onboarding Challenge

asmi-g:
# Purpose
CDH Onboarding Challenge

# New Changes
Initial code for onboarding

# Testing
--N/A


drivers/source/serial_io.c
```diff
@@ -64,7 +67,15 @@ obc_error_code_t sciPrintText(unsigned char *text, uint32_t length) {
 
     /* USER CODE BEGIN */
     // Print text to the serial port using sciSendBytes. Use the mutex to protect the SCI module.
-
+    
+    if (xSemaphoreTake(mutex, (TickType_t)UART_MUTEX_BLOCK_TIME) == pdTRUE){
```
dgobalak: `UART_MUTEX_BLOCK_TIME` is `portMAX_DELAY` which already has type `TickType_t` so this cast isn't needed.

drivers/source/serial_io.c
```diff
@@ -64,7 +67,15 @@ obc_error_code_t sciPrintText(unsigned char *text, uint32_t length) {
...
+        sciSendBytes(text, length);
+        xSemaphoreGive(mutex);
+        return OBC_ERR_CODE_SUCCESS;
+    }
+    else{
+        return OBC_ERR_CODE_SCI_RECEIVE_FAILED;
```
dgobalak: If `xSemaphoreTake` fails, that means the mutex timed out. Returning `OBC_ERR_CODE_MUTEX_TIMEOUT` would be better.

onboarding/source/amb_light_service.c
```diff
@@ -34,20 +47,81 @@ static void lightServiceTask(void * pvParameters);
...
+        lightServiceQueueHandle = xQueueCreateStatic (LS_QUEUE_LENGTH,
+                                                      LS_ITEM_SIZE,
+                                                      lightServiceQueueStorage,
+                                                      &lightServiceQueue  );
+    }
+    
+    if(lightServiceQueueHandle == NULL)
+        return OBC_ERR_CODE_QUEUE_CREATION_FAILED;
 
+    return OBC_ERR_CODE_TASK_CREATION_FAILED;
```
dgobalak: If we get to this point, the function succeeded, so you should return `OBC_ERR_CODE_SUCCESS`

onboarding/source/amb_light_service.c
```diff
@@ -34,20 +47,81 @@ static void lightServiceTask(void * pvParameters);
...
     /* USER CODE END */
 }
 
+
+uint16_t Get_Light_Sensor_data(void)
```
dgobalak: Functions names should be `camelCase`

onboarding/source/amb_light_service.c
```diff
@@ -34,20 +47,81 @@ static void lightServiceTask(void * pvParameters);
...
+{
+    //Function taken from https://git.ti.com/cgit/hercules_examples/hercules_examples/tree/Launchpad/RM/RM46L8/Project_1/demoapp/source/adc_demos.c
+	adcData_t adc_data;
+	adcData_t *adc_data_ptr = &adc_data;
```
dgobalak: Variable names should be `camelCase`

dgobalak: Creating a separate variable for this pointer isn't needed. You can just pass `&adcData` into `adcGetData`

onboarding/source/amb_light_service.c
```diff
@@ -34,20 +47,81 @@ static void lightServiceTask(void * pvParameters);
...
+	/** - Transmit the Conversion data to PC using SCI
+    */
+	return (adc_data_ptr->value);
+}
+
 static void lightServiceTask(void * pvParameters) {
     /* USER CODE BEGIN */
     // Wait for MEASURE_LIGHT event in the queue and then print the ambient light value to the serial port.
-
+    light_event_t LightEvent;
```
dgobalak: Variable name should be `camelCase`

onboarding/source/amb_light_service.c
```diff
@@ -34,20 +47,81 @@ static void lightServiceTask(void * pvParameters);
...
+    uint16_t lightdata = Get_Light_Sensor_data();
+    if (xQueueReceive(lightServiceQueueHandle, &LightEvent, 10) == pdPASS && LightEvent == MEASURE_LIGHT){
+        sciPrintf("Ambient light value: ", lightdata);
+    }
```
dgobalak: This would technically work, but try not to have a condition that's dependent on another condition in the same `if` statement. For example, the result of `LightEvent == MEASURE_LIGHT` depends on `xQueueReceive(lightServiceQueueHandle, &LightEvent, 10) == pdPASS`. To improve readability (and also write this in a more scalable way - like if we support more events), create a separate `if` or `switch` statement to check which event was received.

onboarding/source/amb_light_service.c
```diff
@@ -34,20 +47,81 @@ static void lightServiceTask(void * pvParameters);
...
     /* USER CODE BEGIN */
     // Send the event to the queue. Return error code if event was not sent successfully.
-    
+    if (xQueueSend(lightServiceQueueHandle, event, 10)==pdPASS){
+        return OBC_ERR_CODE_SUCCESS;
+    }
+
+    if (event == NULL){
+        return OBC_ERR_CODE_INVALID_ARG;
+    }
```
dgobalak, line 120: This check should be done before `xQueueSend`

dgobalak (changes requested):
Looks good, just a few comments

onboarding/source/amb_light_service.c
```diff
@@ -34,20 +47,82 @@ static void lightServiceTask(void * pvParameters);
...
+ 	while(!adcIsConversionComplete(adcREG1, adcGROUP1)); 
+
+	/** - Read the conversion result
+	*     The data contains the Light sensor data
+    */
+	adcGetData(adcREG1, adcGROUP1, &adcData);
+	
+	/** - Transmit the Conversion data to PC using SCI
+    */
+	return ((&adcData)->value);
```
dgobalak: `adcData` is a struct, so you can just return `adcData.value`. The `->` operator just dereferences a pointer and then applies the `.` operator. 

I.e. `(&adcData)->value` = `(*(&adcData)).value` = `adcData.value`

dgobalak (changes requested):

onboarding/source/amb_light_service.c
dgobalak: Also, you can resolve the PR comment once you've fixed the issue.

PR #47, opened 2023-05-09
Fayzulloh lvl1

fayzull0h:
# Purpose
Verify solution to mini c programming challenge.

### DIFF
--- a/challenge.c
+++ b/challenge.c
@@ -2,7 +2,7 @@
 #include <stdbool.h>
 #include <stdlib.h>
 #include <stdio.h>
-#include <inttypes.h>
+//#include <inttypes.h>
 
 //-------------------------------------------------------------------------
 // Question 0
--- a/challenge.c
+++ b/challenge.c
@@ -2,12 +2,13 @@
 #include <stdbool.h>
 #include <stdlib.h>
 #include <stdio.h>
+#include <inttypes.h>
 
 //-------------------------------------------------------------------------
 // Question 0
 // Include the challenge.h header file
 //-------------------------------------------------------------------------
-
+#include "challenge.h"
 
 //-------------------------------------------------------------------------
 // Question 1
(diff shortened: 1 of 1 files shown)


challenge.c
```diff
@@ -119,7 +157,13 @@ typedef struct {
...
 
 error_t q10(q10_t *q10) {
-
+    error_t result;
+    if (q9(&(q10->a), &(q10->b)) == 0) {
+        result = SUCCESS;
+    } else {
+        result = FAIL;
+    }
+    return result;
```
dgobalak: You _could_ cut this down to a single ternary statement (not required tho)

challenge.c
```diff
@@ -50,10 +59,18 @@ uint16_t q3(uint8_t x, uint8_t y) {
...
+        return -1;
     }
+
+    int32_t sum = 0;
+    for (uint8_t i = 0, overflow = 0; (uint32_t)i + overflow * 255 < arrayLength; i++) {
+        sum += array[i + overflow * 255];
+        if (i == 255) {
+            overflow++;
+        }
+    }
```
dgobalak: It'd be more readable to just make `i` have type `uint32_t`

challenge.c
```diff
@@ -2,28 +2,31 @@
 #include <stdbool.h>
 #include <stdlib.h>
 #include <stdio.h>
+#include <inttypes.h>
```
dgobalak: Is this header used somewhere in this file? Including `stdint.h` should be sufficient

dgobalak (changes requested):

challenge.c
fayzull0h: No, it's not. I think I wanted to use a function from there at some point, but I ended up not requiring it and forgot to remove it , sorry.

fayzull0h:
Should I make the changes and resubmit?

dgobalak:
> Should I make the changes and resubmit?

Yep, and then u can get started on the level2 onboarding

dgobalak:
lgtm

PR #48, opened 2023-05-11
Fayzulloh - Onboarding Challenge

fayzull0h:
# Purpose
Onboarding challenge
(I don't have a Windows machine, so I used the hal file from the team drive)


drivers/source/serial_io.c
```diff
@@ -64,6 +67,16 @@ obc_error_code_t sciPrintText(unsigned char *text, uint32_t length) {
...
+    if (xSemaphoreTake(mutex, UART_MUTEX_BLOCK_TIME) == pdTRUE) {
+
+        obc_error_code_t result = sciSendBytes(text, length);
+        if (xSemaphoreGive(mutex) == pdTRUE)
+            return OBC_ERR_CODE_QUEUE_FULL;
+
+        return result;
+    } else {
+        return OBC_ERR_CODE_MUTEX_TIMEOUT;
+    }
```
dgobalak: No need for the `else` branch since the `if ` branch returns. You can just do:
```
if (...) {
  ...
  return result;
}

return OBC_ERR_CODE_MUTEX_TIMEOUT;
```

onboarding/source/controller.c
```diff
@@ -122,6 +143,13 @@ static void ledTimerCallback(TimerHandle_t xTimer) {
 static void lightTimerCallback(TimerHandle_t xTimer) {
     /* USER CODE BEGIN */
     // Send light event to light service queue
+    ASSERT(xTimer != NULL);
+    light_event_t * measureLightPtr = MEASURE_LIGHT;
+    obc_error_code_t lightTimerReturn = sendToLightServiceQueue(measureLightPtr);
```
dgobalak: `light_event_t * measureLightPtr = MEASURE_LIGHT;` would create a pointer to address `0x0` (since MEASURE_LIGHT = 0). This isn't exactly what we want. We want to create a variable of type `light_event_t` with a  value of `MEASURE_LIGHT`. `sendToLightServiceQueue` expects a pointer to this variable.

onboarding/source/amb_light_service.c
```diff
@@ -34,20 +46,71 @@ static void lightServiceTask(void * pvParameters);
...
+        adcStartConversion(adcREG1, adcGROUP1);
+
+        // Wait for conversion to complete
+        while (!adcIsConversionComplete(adcREG1, adcGROUP1));
+
+        // Read conversion result data into adcData
+        adcGetData(adcREG1, adcGROUP1, adcDataPtr);
+
+        sciPrintf("%d\n", adcDataPtr->value);
+    }
```
dgobalak: This task should continue running this repeatedly. I.e. all of this should be wrapped in a `while(1)` loop.

onboarding/source/amb_light_service.c
```diff
@@ -34,20 +46,71 @@ static void lightServiceTask(void * pvParameters);
...
+    light_event_t lightEventBuffer;
+
+    BaseType_t xCheckEventReturned = xQueueReceive( eventQueueHandle, 
+                                                    &lightEventBuffer, 
+                                                    0);
+
+    if (xCheckEventReturned == pdTRUE) {
+        // Measure the light and print the result
+        adcData_t adcData;
+        adcData_t *adcDataPtr = &adcData;
```
dgobalak: No need to declare a separate variable for the pointer. You can just pass `&adcData` into `adcGetData`

onboarding/source/amb_light_service.c
```diff
@@ -34,20 +46,71 @@ static void lightServiceTask(void * pvParameters);
...
+
+        // Initialize ADC drivers
+        adcInit();
```
dgobalak: This is called for you in the `main` function

onboarding/source/amb_light_service.c
```diff
@@ -34,20 +46,71 @@ static void lightServiceTask(void * pvParameters);
...
+
+        sciPrintf("%d\n", adcDataPtr->value);
+    }
 
     /* USER CODE END */
 }
 
 obc_error_code_t sendToLightServiceQueue(light_event_t *event) {
     /* USER CODE BEGIN */
     // Send the event to the queue. Return error code if event was not sent successfully.
```
dgobalak, line 109: Check if `event` is NULL

dgobalak (changes requested):
Looks good. Just a few more changes to make

dgobalak (changes requested):
Looks good. Just a few more changes to make

dgobalak:
Lgtm. You can close this PR now

PR #50, opened 2023-06-30
Aryan - Onboarding Challenge

AesthicEthics:
# Purpose
Onboarding Challenge Review PR


drivers/source/serial_io.c
```diff
@@ -64,7 +66,20 @@ obc_error_code_t sciPrintText(unsigned char *text, uint32_t length) {
 
     /* USER CODE BEGIN */
     // Print text to the serial port using sciSendBytes. Use the mutex to protect the SCI module.
+    // take mutex
+    if (sciLinMutex != NULL){
```
dgobalak: This check is done on line 64 so u don't need to do it again.

drivers/source/serial_io.c
```diff
@@ -64,7 +66,20 @@ obc_error_code_t sciPrintText(unsigned char *text, uint32_t length) {
...
+        if (xSemaphoreTake( sciLinMutex, ( TickType_t ) UART_MUTEX_BLOCK_TIME == pdTRUE)){
```
dgobalak: Use `mutex` so we can support printing using either the SCI1 or SCI2/sciLin ports (depending on what UART_PRINT_REG  is defined as).

dgobalak (changes requested):

drivers/source/serial_io.c
Navtajh04: this should be `if (xSemaphoreTake( sciLinMutex, ( TickType_t ) UART_MUTEX_BLOCK_TIME) == pdTRUE)` ie you're trying to compare the return value of xSemaphoreTake to see if it's successful and not UART_MUTEX_BLOCK_TIME

dgobalak (commented):
lgtm. I'll close this PR now

PR #51, opened 2023-07-10
Added Mock functionality for recv and send

Gjjjiang:
# Purpose
Mock i2c send and receive

# New Changes
- Added Mock functionality for recv and send
- Added global variable I2CSentData for send
- Added i2c timeout error

# Testing
- Built

# Outstanding Changes
- Handle incorrect numBytes in send and receive


sys/i2c/i2c_io.c
```diff
@@ -48,7 +49,15 @@ error_code_t i2cSendTo(uint8_t sAddr, uint8_t *buf, uint16_t numBytes) {
   }
 
   /* Mock the transmit */
-
+  if(sAddr == LM75BD_I2C_ADDR){
+    memcpy(&I2CSentData, buf, sizeof(I2CSentData));
+    if(numBytes>sizeof(I2CSentData)){
+      // To Do
+    }
```
dgobalak: Maybe add a check to make sure the reg address (first byte in the buffer) is either the temp reg or config reg?

sys/i2c/i2c_io.c
```diff
@@ -63,7 +72,20 @@ error_code_t i2cReceiveFrom(uint8_t sAddr, uint8_t *buf, uint16_t numBytes) {
...
+  if(sAddr == LM75BD_I2C_ADDR){
+    uint16_t recv_buff = getLm75bdNextTempRegVal();
+    memset(buf, 0, numBytes);
+    if(numBytes <= sizeof(recv_buff)){
+      memcpy(buf, &recv_buff, numBytes);
+    } 
+    else {
+      memcpy(buf, &recv_buff, sizeof(recv_buff));
+      // exit or throw error instead?
+    }
```
dgobalak: The received data depends on the reg address that was sent in the last `i2cSendTo`. So maybe do a check to make sure it was the temp reg?

sys/i2c/i2c_io.c
```diff
@@ -12,12 +12,13 @@
 
 // The mutex timeout for the I2C bus
 #define I2C_MUTEX_TIMEOUT pdMS_TO_TICKS(500)
-
+#define LM75BD_I2C_ADDR 0x4FU
```
dgobalak: This is defined in the header already as `LM75BD_OBC_I2C_ADDR`

dgobalak (changes requested):

sys/i2c/i2c_io.c
Gjjjiang: Is there anything we do if it matches with it or if it doesn't match. Do we actually throw an incorrect reg error or should we write to the I2CSentData regardless?

dgobalak: Either return an error code (and maybe print an error message during `printConsole`). Or just exit the program with a non-zero error code

sys/i2c/i2c_io.c
```diff
@@ -63,7 +91,33 @@ error_code_t i2cReceiveFrom(uint8_t sAddr, uint8_t *buf, uint16_t numBytes) {
...
+    case config_reg:
+      // do something
+      break;
+    default:
+      recv_buff = 0U;
+    }
+
+    memset(buf, 0, numBytes);
+    if(numBytes <= sizeof(recv_buff)){
+      memcpy(buf, &recv_buff, numBytes);
```
dgobalak, line 112: The MSB is received first and should be placed at index 0.

dgobalak (changes requested):

dgobalak (approved):
Lgtm

PR #56, opened 2023-07-14
[BUGFIX] Buffer goes out of scope before unit test can check it

dgobalak:
# Purpose
Fixed bug where the buffer passed into `i2cSendTo` goes out of scope before we can check it in the unit test. This resulted in tests being inconsistent.

# New Changes
- Created a custom mock to copy the buffer contents to a `static` buffer.

# Testing
- Tested against my own solution to the onboarding.

# Outstanding Changes
N/A


Navtajh04 (approved):
LGTM

PR #63, opened 2023-07-14
Update workflow for level1 challenge

dgobalak:
# Purpose
Workflow should run on pushes and PRs. The workflow should fail if an assertion fails.


Navtajh04 (approved):
LGTM

PR #69, opened 2023-07-26
PR for onboarding 

PratyushMakkar:
# Purpose

# New Changes
- Explain new changes

# Testing

# Outstanding Changes

### DIFF
--- a/lm75bd/lm75bd.c
+++ b/lm75bd/lm75bd.c
@@ -33,9 +33,14 @@ error_code_t readTempLM75BD(uint8_t devAddr, float *temp) {
   if (temp == NULL) return ERR_CODE_INVALID_ARG;
 
   uint8_t tmpPointerRegister = LM75BD_TEMP_REG;
-  i2cSendTo(devAddr, &tmpPointerRegister, 1);
+  error_code_t errorCode = {0};
+
+  errorCode =  i2cSendTo(devAddr, &tmpPointerRegister, 1);
+  if (errorCode != ERR_CODE_SUCCESS) return errorCode;
   uint8_t buffer[2] = {0};
-  i2cReceiveFrom(devAddr, buffer, 2);
+
+  errorCode = i2cReceiveFrom(devAddr, buffer, 2);
+  if (errorCode != ERR_CODE_SUCCESS) return errorCode;
 
   uint16_t tmpData =  (((uint16_t) buffer[0]) << 3)  | (((uint16_t) buffer[1])  >> 5);
   float realTmp = 0;
--- a/lm75bd/lm75bd.c
+++ b/lm75bd/lm75bd.c
@@ -6,8 +6,11 @@
 #include <string.h>
 #include <math.h>
 
+#include <stdio.h>
 /* LM75BD Registers (p.8) */
 #define LM75BD_REG_CONF 0x01U  /* Configuration Register (R/W) */
+#define LM75BD_TEMP_REG 0x00
+#define D10_MASK_BYTE 0x80
 
 error_code_t lm75bdInit(lm75bd_config_t *config) {
   error_code_t errCode;
@@ -27,8 +30,19 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
 }
 
 error_code_t readTempLM75BD(uint8_t devAddr, float *temp) {
-  /* Implement this driver function */
-  
+  uint8_t tmp_reg_pointer = LM75BD_TEMP_REG;
+  i2cSendTo(devAddr, &tmp_reg_pointer, 1);
+  uint8_t buffer[] = {0x00, 0x00};
+  i2cReceiveFrom(devAddr, &buffer[0], 2);
+
+  uint16_t tmp_data =  (((uint16_t) buffer[0]) << 3)  | (((uint16_t) buffer[1])  >> 5);
+  float real_tmp = 0;
+  if (buffer[0] & D10_MASK_BYTE) {
+    real_tmp = - ((tmp_data ^ 0x07FF) + 0x0001) * 0.125;
+  } else {
+    real_tmp = tmp_data * 0.125;
+  }
+  *temp = real_tmp; 
   return ERR_CODE_SUCCESS;
 }
 
--- a/lm75bd/lm75bd.c
+++ b/lm75bd/lm75bd.c
@@ -30,19 +30,21 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
 }
 
 error_code_t readTempLM75BD(uint8_t devAddr, float *temp) {
-  uint8_t tmp_reg_pointer = LM75BD_TEMP_REG;
-  i2cSendTo(devAddr, &tmp_reg_pointer, 1);
-  uint8_t buffer[] = {0x00, 0x00};
-  i2cReceiveFrom(devAddr, &buffer[0], 2);
+  if (temp == NULL) return ERR_CODE_INVALID_ARG;
 
-  uint16_t tmp_data =  (((uint16_t) buffer[0]) << 3)  | (((uint16_t) buffer[1])  >> 5);
-  float real_tmp = 0;
-  if (buffer[0] & D10_MASK_BYTE) {
-    real_tmp = - ((tmp_data ^ 0x07FF) + 0x0001) * 0.125;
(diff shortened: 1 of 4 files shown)


lm75bd/lm75bd.c
```diff
@@ -27,8 +30,19 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
 }
 
 error_code_t readTempLM75BD(uint8_t devAddr, float *temp) {
-  /* Implement this driver function */
-  
+  uint8_t tmp_reg_pointer = LM75BD_TEMP_REG;
```
Navtajh04: variable names should be camelCase. This applies everywhere else for your other variables as well

lm75bd/lm75bd.c
```diff
@@ -27,8 +30,19 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
...
+  i2cSendTo(devAddr, &tmp_reg_pointer, 1);
+  uint8_t buffer[] = {0x00, 0x00};
+  i2cReceiveFrom(devAddr, &buffer[0], 2);
+
+  uint16_t tmp_data =  (((uint16_t) buffer[0]) << 3)  | (((uint16_t) buffer[1])  >> 5);
+  float real_tmp = 0;
+  if (buffer[0] & D10_MASK_BYTE) {
+    real_tmp = - ((tmp_data ^ 0x07FF) + 0x0001) * 0.125;
```
Navtajh04: Use a macro or add a comment so that it's clear where these numbers are coming from

services/thermal_mgr/thermal_mgr.c
```diff
@@ -42,19 +42,34 @@ void initThermalSystemManager(lm75bd_config_t *config) {
 }
 
 error_code_t thermalMgrSendEvent(thermal_mgr_event_t *event) {
-  /* Send an event to the thermal manager queue */
-
+  if (xQueueSend(thermalMgrQueueHandle, event, portMAX_DELAY) != pdPASS) {
```
Navtajh04: This function is called in an ISR and we should never be blocking in an ISR, so we shouldn't be using `portMAX_DELAY` since this means we will block indefinitely until there is space in the queue

services/thermal_mgr/thermal_mgr.c
```diff
@@ -42,19 +42,34 @@ void initThermalSystemManager(lm75bd_config_t *config) {
 }
 
 error_code_t thermalMgrSendEvent(thermal_mgr_event_t *event) {
-  /* Send an event to the thermal manager queue */
```
Navtajh04, line 45: check if event is not null and that the queue has been created

lm75bd/lm75bd.c
```diff
@@ -27,8 +30,19 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
 }
 
 error_code_t readTempLM75BD(uint8_t devAddr, float *temp) {
-  /* Implement this driver function */
-
```
Navtajh04, line 31: Check if temp is null

Navtajh04 (changes requested):
Looks good, just a few things that should be changed

lm75bd/lm75bd.c
```diff
@@ -27,8 +30,19 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
 }
 
 error_code_t readTempLM75BD(uint8_t devAddr, float *temp) {
-  /* Implement this driver function */
-  
+  uint8_t tmp_reg_pointer = LM75BD_TEMP_REG;
+  i2cSendTo(devAddr, &tmp_reg_pointer, 1);
+  uint8_t buffer[] = {0x00, 0x00};
```
dgobalak: If you want to declare and initialize a buffer (of size 2) to all zeros, you can also do `uint8_t buffer[2] = {0};`. If you have a brace-enclosed list and there are fewer initialized values than the number of total elements in the list/structure, the uninitialized values will become zero (this is guaranteed by the C standard). This is particularly useful if you want to initialize a large buffer.

lm75bd/lm75bd.c
```diff
@@ -27,8 +30,19 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
...
+  i2cReceiveFrom(devAddr, &buffer[0], 2);
```
dgobalak: This is equivalent to `i2cReceiveFrom(devAddr, buffer, 2)`. We tend to do it this way since it's a bit cleaner.

services/thermal_mgr/thermal_mgr.c
```diff
@@ -42,19 +42,34 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
-
+  if (xQueueSend(thermalMgrQueueHandle, event, portMAX_DELAY) != pdPASS) {
+    return ERR_CODE_INVALID_QUEUE_MSG;    //IDK WHAT TO SEND HERE
```
dgobalak: `xQueueSend` will either return `pdPASS` (which is equivalent to `pdTRUE`) or `errQUEUE_FULL` (look at https://www.freertos.org/a00117.html). So a more appropriate error code to return here would be `ERR_CODE_QUEUE_FULL`

services/thermal_mgr/thermal_mgr.c
```diff
@@ -42,19 +42,34 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+  }
   return ERR_CODE_SUCCESS;
 }
 
 void osHandlerLM75BD(void) {
-  /* Implement this function */
+  thermal_mgr_event_t eventTriggered;
```
dgobalak: It's usually a good idea to initialize structs to zero. https://stackoverflow.com/questions/11152160/initializing-a-struct-to-0

services/thermal_mgr/thermal_mgr.c
```diff
@@ -42,19 +42,34 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+  thermalMgrSendEvent(&eventTriggered);
 }
 
 static void thermalMgr(void *pvParameters) {
+  thermal_mgr_event_t event;
   /* Implement this task */
   while (1) {
-    
+    float tmp;
+    lm75bd_config_t config  = * (lm75bd_config_t *) pvParameters;
```
dgobalak: Try and limit variable scope as much as possible. In this case, `event` isn't used outside the `while` loop. And `tmp` isn't used outside the ` if (xQueueReceive(thermalMgrQueueHandle, &event, portMAX_DELAY) == pdPASS)` block.

dgobalak (changes requested):

services/thermal_mgr/thermal_mgr.c
```diff
@@ -42,19 +42,34 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+    if (xQueueReceive(thermalMgrQueueHandle, &event, portMAX_DELAY) == pdPASS) {
+      if (event.type == THERMAL_MGR_EVENT_MEASURE_TEMP_CMD) {
+        readTempLM75BD(config.devAddr, &tmp);
+        addTemperatureTelemetry(tmp);
+      } else if (event.type == THERMAL_EVENT_DETECTED) {
+        readTempLM75BD(config.devAddr, &tmp);
```
dgobalak: Also, check the return value of `readTempLM75BD` before using `tmp`. If you failed to read the temperature, then `tmp` could be anything (since you didn't initialize `tmp` to anything).

dgobalak (changes requested):

lm75bd/lm75bd.c
```diff
@@ -27,8 +30,21 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
...
 
 error_code_t readTempLM75BD(uint8_t devAddr, float *temp) {
-  /* Implement this driver function */
-  
+  if (temp == NULL) return ERR_CODE_INVALID_ARG;
+
+  uint8_t tmpPointerRegister = LM75BD_TEMP_REG;
+  i2cSendTo(devAddr, &tmpPointerRegister, 1);
+  uint8_t buffer[2] = {0};
+  i2cReceiveFrom(devAddr, buffer, 2);
```
dgobalak: Check the return value of these i2c functions. If a non-success error code is returned, then return that error code.

dgobalak (changes requested):

Navtajh04 (approved):
LGTM

PR #70, opened 2023-07-29
Firmware Onboarding Challenge Submission

Allan-Liu05:
# Purpose
Finished every portion detailed in the notion document, submitting to see what mistakes were made using the automated tests.
# New Changes
- Explain new changes
I'm fairly certain I addressed the issues brought up in the previous pull request. I changed up the flow in the osHandler() function by removing the call to readTemperatureLM75BD(). Added an event type in the enum thermal_mgr_event_type_t so osHandler() just sends an interrupt event now. 
# Testing

# Outstanding Changes


services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,19 +43,44 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
   /* Implement this function */
+  
+  float *temp;
+  readTempLM75BD(LM75BD_OBC_I2C_ADDR, temp);
+  if (*temp > 75.0) {
+    overTemperatureDetected();
+  }
+  else {
+    safeOperatingConditions();
+  }
```
dgobalak: `readTempLM75BD` is a blocking function call. Internally, it'll try to take a mutex that protects the I2C bus. So you shouldn't call it from `osHandlerLM75BD` since this function is meant to execute from within an ISR.

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,19 +43,44 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+  else {
+    safeOperatingConditions();
+  } 
 }
 
 static void thermalMgr(void *pvParameters) {
   /* Implement this task */
+  
+  float *temp;
```
dgobalak: You're not actually allocating any memory to store the temperature value. You're just allocating memory to store a pointer and passing the pointer into `readLM75BD`. This could result in the program accessing a random part of memory (and potentially causing a segmentation fault) when it dereferences that pointer.

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,19 +43,44 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
   while (1) {
-    
+    thermal_mgr_event_t check; 
+    if (xQueueReceive(thermalMgrQueueHandle, pvParameters , 0) == pdTRUE) {
+      if (check.type == THERMAL_MGR_EVENT_MEASURE_TEMP_CMD) {
+        readTempLM75BD(LM75BD_OBC_I2C_ADDR, temp);
+        addTemperatureTelemetry(*temp); 
+      }
```
dgobalak: `check` is never updated here

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,19 +43,44 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
 static void thermalMgr(void *pvParameters) {
   /* Implement this task */
+  
+  float *temp;
   while (1) {
-    
+    thermal_mgr_event_t check; 
+    if (xQueueReceive(thermalMgrQueueHandle, pvParameters , 0) == pdTRUE) {
+      if (check.type == THERMAL_MGR_EVENT_MEASURE_TEMP_CMD) {
+        readTempLM75BD(LM75BD_OBC_I2C_ADDR, temp);
```
dgobalak: Check the returned value of `readTempLM75BD` to verify that the function succeeded. Otherwise, you won't know if the temperature value is valid

lm75bd/lm75bd.c
```diff
@@ -29,6 +29,36 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
...
   /* Implement this driver function */
   
+  if (temp == NULL) {
+    return ERR_CODE_INVALID_ARG;
+  } 
+  uint8_t tempWriteBit = 0x00;
+ 
+  i2cSendTo(devAddr, &tempWriteBit,(uint16_t) 1);
+  uint8_t tempData[2];
+  i2cReceiveFrom(devAddr, tempData, (uint16_t) 2);
```
dgobalak: No need to cast the literals to `uint16_t`

lm75bd/lm75bd.c
```diff
@@ -29,6 +29,36 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
 error_code_t readTempLM75BD(uint8_t devAddr, float *temp) {
   /* Implement this driver function */
   
+  if (temp == NULL) {
+    return ERR_CODE_INVALID_ARG;
+  } 
+  uint8_t tempWriteBit = 0x00;
+ 
+  i2cSendTo(devAddr, &tempWriteBit,(uint16_t) 1);
+  uint8_t tempData[2];
```
dgobalak: Initialize the buffer to all zeros

lm75bd/lm75bd.c
```diff
@@ -29,6 +29,36 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
 error_code_t readTempLM75BD(uint8_t devAddr, float *temp) {
   /* Implement this driver function */
   
+  if (temp == NULL) {
+    return ERR_CODE_INVALID_ARG;
+  } 
+  uint8_t tempWriteBit = 0x00;
+ 
+  i2cSendTo(devAddr, &tempWriteBit,(uint16_t) 1);
```
dgobalak: Check the return value to ensure it succeeded

lm75bd/lm75bd.c
dgobalak: Check the return value to ensure it succeeded

lm75bd/lm75bd.c
```diff
@@ -29,6 +29,36 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
...
+  /* Bitshift right 5 because the 5 LSB are useless so delete */
+  rawData = rawData >> 5;
+
+  if (posOrNeg == 0) /* Positive Temperature */{ 
+    *temp = (float)(rawData) * 0.125;
+  }
+
+  else /* Negative Temperature*/ {
+    /* Convert to two's complement */
+    uint16_t flipLeadingBits = 0xF800;
```
dgobalak: Since this is a constant, declare it with `const`

dgobalak (changes requested):
Looks pretty good. Just a few issues. If you take a look at the "Simulated Test Run" workflow (https://github.com/UWOrbital/OBC-Firmware-Onboarding/actions/runs/5698878766/job/15447214415), you'll notice that the "Test Run" step timed out. Once the issues are resolved, you should see the actual test run output.

lm75bd/lm75bd.c
Navtajh04: this can also just be made into a macro

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,19 +43,44 @@ void initThermalSystemManager(lm75bd_config_t *config) {
 
 error_code_t thermalMgrSendEvent(thermal_mgr_event_t *event) {
   /* Send an event to the thermal manager queue */
-
```
Navtajh04, line 46: check to see if the queue has been successfully created

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,19 +43,44 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+  
+  if (event == NULL) {
+    return ERR_CODE_INVALID_ARG;
+  }
+  xQueueSend(thermalMgrQueueHandle, event, 0);
```
Navtajh04: check the return value since this call could return `pdFAIL` if the queue is full

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,19 +43,44 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
 }
 
 static void thermalMgr(void *pvParameters) {
   /* Implement this task */
+  
+  float *temp;
   while (1) {
-    
+    thermal_mgr_event_t check; 
+    if (xQueueReceive(thermalMgrQueueHandle, pvParameters , 0) == pdTRUE) {
```
Navtajh04: There's no need to be polling to see if anything has entered the queue by setting the block time to 0. It would be better to have this block for longer and wait for something to be added to the queue  before continuing

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,19 +43,44 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+      if (check.type == THERMAL_MGR_EVENT_MEASURE_TEMP_CMD) {
```
Navtajh04: check was never passed into the call to `xQueueReceive` so it will never be updated after your initialization in line 74

services/thermal_mgr/thermal_mgr.c
Navtajh04: `pvParameters` isn't what you want to be passing in for the second argument of this function. The second parameter is where the received even from the queue will be stored, while `pvParameters` is a parameter that can optionally be passed into the task when it is created and then is never updated again

Navtajh04 (changes requested):

lm75bd/lm75bd.c
```diff
@@ -28,7 +28,43 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
 
 error_code_t readTempLM75BD(uint8_t devAddr, float *temp) {
   /* Implement this driver function */
-  
+  error_code_t errCodeSend;
+  error_code_t errCodeReceive;
+  if (temp == NULL) {
+    return ERR_CODE_INVALID_ARG;
+  } 
+  uint8_t tempWriteBit = 0x00;
```
Navtajh04: This should be renamed since you are sending a while byte not just a single bit

Navtajh04: Also add a comment or use a macro to explain the `0x00`

lm75bd/lm75bd.c
```diff
@@ -28,7 +28,43 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
...
+ 
+  errCodeSend = i2cSendTo(devAddr, &tempWriteBit, 1);
+  if (errCodeSend == ERR_CODE_INVALID_ARG) {
+    return ERR_CODE_INVALID_ARG;
```
Navtajh04: Just check if `errCodeSend != ERR_CODE_SUCCESS` and return the errCode since if the function ever changes and returns some other error code on this call then you won't catch the error here

lm75bd/lm75bd.c
```diff
@@ -28,7 +28,43 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
...
+  }
+  uint8_t tempData[2] = {0};
+  errCodeReceive = i2cReceiveFrom(devAddr, tempData, 2);
+  if (errCodeReceive == ERR_CODE_INVALID_ARG) {
+    return ERR_CODE_INVALID_ARG;
```
Navtajh04: Just check if `errCodeSend != ERR_CODE_SUCCESS` and return the errCode since if the function ever changes and returns some other error code on this call then you won't catch the error here

services/thermal_mgr/thermal_mgr.c
Navtajh04: I don't think this was addressed

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,21 +46,56 @@ void initThermalSystemManager(lm75bd_config_t *config) {
 
 error_code_t thermalMgrSendEvent(thermal_mgr_event_t *event) {
   /* Send an event to the thermal manager queue */
-
+  BaseType_t checkQueueSend;
+  if (event == NULL) {
+    return ERR_CODE_INVALID_ARG;
+  }
+  checkQueueSend = xQueueSend(thermalMgrQueueHandle, (void *) event, (TickType_t) 10);
+  if (checkQueueSend == pdFAIL) {
```
Navtajh04: this can be combined to just but `if(xQueueSend(thermalMgrQueueHandle, (void *) event, (TickType_t) 10) == pdFAIL)`

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,21 +46,56 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+    return ERR_CODE_INVALID_ARG;
```
Navtajh04: this error code should only be used if an invalid argument is passed into a function, but xQueueSend will fail when the queue is full

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,21 +46,56 @@ void initThermalSystemManager(lm75bd_config_t *config) {
 
 error_code_t thermalMgrSendEvent(thermal_mgr_event_t *event) {
   /* Send an event to the thermal manager queue */
-
+  BaseType_t checkQueueSend;
+  if (event == NULL) {
+    return ERR_CODE_INVALID_ARG;
+  }
+  checkQueueSend = xQueueSend(thermalMgrQueueHandle, (void *) event, (TickType_t) 10);
```
Navtajh04: when you pass in (TickType_t) 10, this means that this will block for 10 ticks if the queue is full, but you are calling this function from an ISR in osHandler so you should not be blocking

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,21 +46,56 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
 
 static void thermalMgr(void *pvParameters) {
   /* Implement this task */
+  float tempData;
+  error_code_t readTempCheck;
+  thermal_mgr_event_t eventTypeCheck; 
+
   while (1) {
-    
+    if (xQueueReceive(thermalMgrQueueHandle, (void *) &eventTypeCheck , (TickType_t) 10) == pdTRUE) {
```
Navtajh04: block time should be longer

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,21 +46,56 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+  interrupt.type = THERMAL_MGR_EVENT_TYPE_INTERRUPT;
+  thermalMgrSendEvent(&interrupt); 
+  
 }
 
 static void thermalMgr(void *pvParameters) {
   /* Implement this task */
+  float tempData;
+  error_code_t readTempCheck;
+  thermal_mgr_event_t eventTypeCheck;
```
Navtajh04: In general try to limit the scope of your variables to be as small as possible, so since these are not used outside the while loop they should not be declared outside of it

Navtajh04 (changes requested):
Looks good, just a few more changes

lm75bd/lm75bd.c
```diff
@@ -28,7 +28,43 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
...
+  error_code_t errCodeSend;
+  error_code_t errCodeReceive;
+  if (temp == NULL) {
+    return ERR_CODE_INVALID_ARG;
+  } 
+  uint8_t tempWriteByte = 0x00; // 0x00 because that's what the notion document specified for the i2cSendTo() function
+ 
+  errCodeSend = i2cSendTo(devAddr, &tempWriteByte, 1);
+  if (errCodeSend != ERR_CODE_SUCCESS) {
+    return ERR_CODE_INVALID_ARG;
```
Navtajh04: return `errCodeSend` since it could be anything

lm75bd/lm75bd.c
```diff
@@ -28,7 +28,43 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
 
 error_code_t readTempLM75BD(uint8_t devAddr, float *temp) {
   /* Implement this driver function */
-  
+  error_code_t errCodeSend;
+  error_code_t errCodeReceive;
```
Navtajh04: you can just use one `errCode` instead of using a separate one for each function

lm75bd/lm75bd.c
```diff
@@ -28,7 +28,43 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
...
+  uint8_t tempWriteByte = 0x00; // 0x00 because that's what the notion document specified for the i2cSendTo() function
+ 
+  errCodeSend = i2cSendTo(devAddr, &tempWriteByte, 1);
+  if (errCodeSend != ERR_CODE_SUCCESS) {
+    return ERR_CODE_INVALID_ARG;
+  }
+  uint8_t tempData[2] = {0};
+  errCodeReceive = i2cReceiveFrom(devAddr, tempData, 2);
+  if (errCodeReceive != ERR_CODE_SUCCESS) {
+    return ERR_CODE_INVALID_ARG;
```
Navtajh04: return the error code variable instead of hard coding the return since the error code could've been anything

Navtajh04 (changes requested):
Just 3 quick changes

Navtajh04 (commented):
LGTM

PR #73, opened 2023-08-03
FM - Onboarding Challenge Level 3

farzanmirshekari:
# Purpose
Complete the onboarding challenge for the FW team.

# New Changes
Drivers for the sensor, OS handler.

# Testing
Unit and integration tests pass locally.

# Outstanding Changes
N/a

### DIFF
--- a/services/thermal_mgr/thermal_mgr.c
+++ b/services/thermal_mgr/thermal_mgr.c
@@ -43,7 +43,9 @@ void initThermalSystemManager(lm75bd_config_t *config) {
 
 error_code_t thermalMgrSendEvent(thermal_mgr_event_t *event) {
   /* Send an event to the thermal manager queue */
-
+  if (xQueueSend(thermalMgrQueueHandle, event, 0) != pdTRUE) {
+    return ERR_CODE_QUEUE_FULL;
+  }
   return ERR_CODE_SUCCESS;
 }
 
@@ -53,8 +55,24 @@ void osHandlerLM75BD(void) {
 
 static void thermalMgr(void *pvParameters) {
   /* Implement this task */
+  thermal_mgr_event_t event;
   while (1) {
-    
+    if (xQueueReceive(thermalMgrQueueHandle, &event, 0) == pdTRUE) {
+      error_code_t errCode;
+      float tempC;
+      switch (event.type) {
+        case THERMAL_MGR_EVENT_MEASURE_TEMP_CMD:
+          errCode = readTempLM75BD(0x48, &tempC);
+          if (errCode == ERR_CODE_SUCCESS) {
+            addTemperatureTelemetry(tempC);
+          } else {
+            printConsole("Error reading temperature: %d\n", errCode);
+          }
+          break;
+        default:
+          break;
+      }
+    }
   }
 }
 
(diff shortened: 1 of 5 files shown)


services/thermal_mgr/thermal_mgr.c
```diff
@@ -44,17 +44,48 @@ void initThermalSystemManager(lm75bd_config_t *config) {
 error_code_t thermalMgrSendEvent(thermal_mgr_event_t *event) {
   /* Send an event to the thermal manager queue */
 
+  if (event == NULL) return ERR_CODE_INVALID_QUEUE_MSG;
+  else if (thermalMgrQueueHandle == NULL) return ERR_CODE_COULD_NOT_CREATE_QUEUE;
```
Navtajh04: the error message name `ERR_CODE_COULD_NOT_CREATE_QUEUE` is a bit misleading since this function doesn't actually create the queue and it could just be that the function that creates the queue just hasn't been called yet

services/thermal_mgr/thermal_mgr.c
```diff
@@ -44,17 +44,48 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
   /* Implement this function */
+  thermal_mgr_event_t interrupt;
+  interrupt.type = THERMAL_MGR_EVENT_INTERRUPT;
+  thermalMgrSendEvent(&interrupt);
 }
 
 static void thermalMgr(void *pvParameters) {
   /* Implement this task */
+  thermal_mgr_event_t event;
+  float tempC;
```
Navtajh04: In general try to limit variables to the smallest possible scope, so declare these in the while loop

services/thermal_mgr/thermal_mgr.c
```diff
@@ -44,17 +44,48 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
   while (1) {
     
+    if (xQueueReceive(thermalMgrQueueHandle, &event, 0) == pdTRUE) {
```
Navtajh04: we don't need to be polling the queue here so a block time of 5000ms or `portMAX_DELAY` would be more appropriate than block time of 0 to avoid wasting CPU time

services/thermal_mgr/thermal_mgr.c
```diff
@@ -44,17 +44,48 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+            addTemperatureTelemetry(tempC);
+          }
+          break;
+        case THERMAL_MGR_EVENT_INTERRUPT:
+          if (readTempLM75BD(LM75BD_OBC_I2C_ADDR, &tempC) == ERR_CODE_SUCCESS) {
+            if (tempC > LM75BD_DEFAULT_OT_THRESH) {
+              overTemperatureDetected();
+            } else if (tempC < LM75BD_DEFAULT_HYST_THRESH) {
+              safeOperatingConditions();
+            }
```
Navtajh04: there will naturally be latency between the interrupt being triggered and us reading the temperature, so if the temperature were to fluctuate and drop back below 82 degrees to 81 for example both of these if statements would evaluate to false and we wouldn't end up dealing with the overtemperature event so we should just be checking if its > 75 for over temp event and else its safe

services/thermal_mgr/thermal_mgr.c
Navtajh04: Invalid state would be more appropriate

Navtajh04 (changes requested):
Looks really good, just a few comments

Navtajh04 (approved):
LGTM send your email in the discord or DM me it and I'll add you to the notion

PR #74, opened 2023-08-04
Fayzulloh - Onboarding Challenge Level 3

fayzull0h:
Onboarding

### DIFF
--- a/lm75bd/lm75bd.c
+++ b/lm75bd/lm75bd.c
@@ -28,16 +28,18 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
 
 error_code_t readTempLM75BD(uint8_t devAddr, float *temp) {
   /* Implement this driver function */
+  if (temp == NULL) return ERR_CODE_INVALID_ARG;
+  error_code_t err;
 
   /* Select the temperature register */
   uint8_t tempRegister[1] = {0x00};
-  error_code_t selectTemp = i2cSendTo(devAddr, tempRegister, 1);
-  if (selectTemp != ERR_CODE_SUCCESS) return selectTemp;
+  err = i2cSendTo(devAddr, tempRegister, 1);
+  if (err != ERR_CODE_SUCCESS) return err;
   
   /* Receive data from temperature reader */
   uint8_t recTempBuff[2];
-  error_code_t recTemp = i2cReceiveFrom(devAddr, recTempBuff, 2);
-  if (recTemp != ERR_CODE_SUCCESS) return recTemp;
+  err = i2cReceiveFrom(devAddr, recTempBuff, 2);
+  if (err != ERR_CODE_SUCCESS) return err;
 
   /* Convert data to Celsius */
   uint16_t tempData16 = ((recTempBuff[0] << 8) | recTempBuff[1]) >> 5;
--- a/lm75bd/lm75bd.c
+++ b/lm75bd/lm75bd.c
@@ -28,7 +28,23 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
 
 error_code_t readTempLM75BD(uint8_t devAddr, float *temp) {
   /* Implement this driver function */
+
+  /* Select the temperature register */
+  uint8_t tempRegister[1] = {0x00};
+  error_code_t selectTemp = i2cSendTo(devAddr, tempRegister, 1);
+  if (selectTemp != ERR_CODE_SUCCESS) return selectTemp;
   
+  /* Receive data from temperature reader */
+  uint8_t recTempBuff[2];
+  error_code_t recTemp = i2cReceiveFrom(devAddr, recTempBuff, 2);
+  if (recTemp != ERR_CODE_SUCCESS) return recTemp;
+
+  /* Convert data to Celsius */
+  uint16_t tempData16 = ((recTempBuff[0] << 8) | recTempBuff[1]) >> 5;
+  uint8_t tempSign = recTempBuff[0] >> 7; // tempSign == true means negative
+
+  *temp = tempSign ? -((uint16_t)~(tempData16 | 0xF800) + 1) * 0.125 : tempData16 * 0.125;
+
   return ERR_CODE_SUCCESS;
 }
 
--- a/services/thermal_mgr/thermal_mgr.c
+++ b/services/thermal_mgr/thermal_mgr.c
@@ -44,7 +44,7 @@ void initThermalSystemManager(lm75bd_config_t *config) {
 error_code_t thermalMgrSendEvent(thermal_mgr_event_t *event) {
   /* Send an event to the thermal manager queue */
   if (event == NULL) return ERR_CODE_INVALID_ARG;
-  if (!thermalMgrQueueHandle) return ERR_CODE_UNKNOWN;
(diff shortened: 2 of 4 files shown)


lm75bd/lm75bd.c
```diff
@@ -28,7 +28,23 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
...
   /* Implement this driver function */
+
+  /* Select the temperature register */
+  uint8_t tempRegister[1] = {0x00};
+  error_code_t selectTemp = i2cSendTo(devAddr, tempRegister, 1);
+  if (selectTemp != ERR_CODE_SUCCESS) return selectTemp;
   
+  /* Receive data from temperature reader */
+  uint8_t recTempBuff[2];
+  error_code_t recTemp = i2cReceiveFrom(devAddr, recTempBuff, 2);
```
Navtajh04: you can just initialize one error code variable at the top of the function and use that for every function call instead of having a separate one for every function call

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +43,41 @@ void initThermalSystemManager(lm75bd_config_t *config) {
 
 error_code_t thermalMgrSendEvent(thermal_mgr_event_t *event) {
   /* Send an event to the thermal manager queue */
+  const void * itemToQueue = (const void *)event;
```
Navtajh04: this varaible isn't needed, just pass in event to the queue and type cast it to `const void *`

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +43,41 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+  BaseType_t sendEventResult = xQueueSend(thermalMgrQueueHandle, itemToQueue, 0);
+  if (sendEventResult != pdTRUE) return ERR_CODE_QUEUE_FULL;
```
Navtajh04: This variable isn't need you can just do `if(xQueueSend(...) != pdTRUE)`

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +43,41 @@ void initThermalSystemManager(lm75bd_config_t *config) {
 
 error_code_t thermalMgrSendEvent(thermal_mgr_event_t *event) {
   /* Send an event to the thermal manager queue */
```
Navtajh04, line 45: make sure event is not null

Navtajh04: make sure the queue has been created

lm75bd/lm75bd.c
```diff
@@ -28,7 +28,23 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
 
 error_code_t readTempLM75BD(uint8_t devAddr, float *temp) {
   /* Implement this driver function */
+
```
Navtajh04, line 33: check to make sure temp is not null

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +43,41 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+
+  /* Add interrupt event to the queue */
+  thermalMgrSendEvent(&checkOverUnder);
 }
 
 static void thermalMgr(void *pvParameters) {
   /* Implement this task */
+  lm75bd_config_t config = *(lm75bd_config_t *)pvParameters;
+  float currTemp;
+  thermal_mgr_event_t nextEvent;
```
Navtajh04: in general try to limit the scope of variables as small as possible so these should be declared in the while loop

lm75bd/lm75bd.c
```diff
@@ -28,7 +28,23 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
```
Navtajh04, line 28: your simulated test run results dont look right

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +43,41 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+
   while (1) {
-    
+    /* Receive next event and read temperature */
+    BaseType_t recResult = xQueueReceive(thermalMgrQueueHandle, &nextEvent, 0);
```
Navtajh04: this variable isnt need you can just check the xQueueReceive return value directly in an if statement

Navtajh04: Also the return value should be checked since right now you are continuing and doing the rest of the code even if `xQueueReceive` returns pdFALSE, and since your block time is 0, you are essentially performing the code in an infinite loop instead of waiting for an event

Navtajh04: Also the block time should be set higher since we know that we'll only be reading temperature every 5 seconds anyways and this task has nothing else to do if we don't receive anything, there's no reason to spend CPU time polling the xQueueReceive even though it's empty. When an item enters the queue this task will become unblocked anyways

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +43,41 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+    error_code_t tempReadResult = readTempLM75BD(config.devAddr, &currTemp);
```
Navtajh04: we shouldnt be reading the temperature no matter what the queue message is since it could also be an invalid queue message

Navtajh04 (changes requested):

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +43,42 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+    float currTemp;
+    thermal_mgr_event_t nextEvent;
+
+    /* Receive next event */
+    if (xQueueReceive(thermalMgrQueueHandle, &nextEvent, (TickType_t) portTICK_PERIOD_MS * 5000) != pdTRUE) 
+      continue;
+
+    /* Check for interrupt condition else check for telemetry request */
+    if (nextEvent.type == THERMAL_MGR_EVENT_OVER_UNDER) {
+      if (currTemp > config.hysteresisThresholdCelsius) overTemperatureDetected();
```
Navtajh04, line 77: you should be reading the temperature to update this value once you get the interrupt to decide if its  overTemp or safe since if you don't this value will be an old value

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +43,42 @@ void initThermalSystemManager(lm75bd_config_t *config) {
 
 error_code_t thermalMgrSendEvent(thermal_mgr_event_t *event) {
   /* Send an event to the thermal manager queue */
+  if (event == NULL) return ERR_CODE_INVALID_ARG;
+  if (!thermalMgrQueueHandle) return ERR_CODE_UNKNOWN;
```
Navtajh04: `ERR_CODE_INVALID_STATE` would be a better err code to return here

Navtajh04 (changes requested):
Looks good, just 2 more things

Navtajh04 (commented):
LGTM

PR #76, opened 2023-08-15
passed tests

RYZEXY:
# Purpose

# New Changes
- Explain new changes

# Testing

# Outstanding Changes

### DIFF
--- a/lm75bd/lm75bd.c
+++ b/lm75bd/lm75bd.c
@@ -1,6 +1,6 @@
 #include "lm75bd.h"
 #include "i2c_io.h"
-#include "../sys/errors.h"
+#include "errors.h"
 
 #include <stdint.h>
 #include <string.h>
@@ -43,12 +43,11 @@ error_code_t readTempLM75BD(uint8_t devAddr, float *temp)
   {
     return errCode;
   }
-  error_code_t errCode2;
-  errCode2 = i2cReceiveFrom(devAddr, buff, 2);
+  errCode = i2cReceiveFrom(devAddr, buff, 2);
 
-  if (errCode2 != ERR_CODE_SUCCESS)
+  if (errCode != ERR_CODE_SUCCESS)
   {
-    return errCode2;
+    return errCode;
   }
 
   uint16_t tempData = ((uint16_t)buff[0] << 3) | ((uint16_t)buff[1] >> 5);
--- a/lm75bd/lm75bd.c
+++ b/lm75bd/lm75bd.c
@@ -1,6 +1,6 @@
 #include "lm75bd.h"
 #include "i2c_io.h"
-#include "errors.h"
+#include "../sys/errors.h"
 
 #include <stdint.h>
 #include <string.h>
@@ -26,9 +26,43 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
   return ERR_CODE_SUCCESS;
 }
 
-error_code_t readTempLM75BD(uint8_t devAddr, float *temp) {
+error_code_t readTempLM75BD(uint8_t devAddr, float *temp)
+{
   /* Implement this driver function */
-  
+  if (!temp)
+  {
+    return ERR_CODE_INVALID_ARG;
+  }
+
+  uint8_t tempPointerRegister = 0x00;
+  uint8_t buff[2] = {0};
+
+  error_code_t errCode;
+  errCode = i2cSendTo(devAddr, &tempPointerRegister, 1);
+  if (errCode != ERR_CODE_SUCCESS)
+  {
+    return errCode;
+  }
+  error_code_t errCode2;
+  errCode2 = i2cReceiveFrom(devAddr, buff, 2);
+
+  if (errCode2 != ERR_CODE_SUCCESS)
+  {
+    return errCode2;
+  }
+
+  uint16_t tempData = ((uint16_t)buff[0] << 3) | ((uint16_t)buff[1] >> 5);
+
+  if (buff[0] >> 7)
+  {
+    tempData = (~tempData + 1) ^ (0xF800);
+    *temp = -tempData * 0.125;
+  }
+  else
+  {
+    *temp = tempData * 0.125;
+  }
+
   return ERR_CODE_SUCCESS;
 }
 
--- a/lm75bd/lm75bd.c
+++ b/lm75bd/lm75bd.c
@@ -43,7 +43,7 @@ error_code_t readTempLM75BD(uint8_t devAddr, float *temp)
   {
     return errCode;
   }
-  errCode = i2cReceiveFrom(devAddr, buff, 2);
+  errCode = i2cReceiveFrom(devAddr, &buff, 2);
 
   if (errCode != ERR_CODE_SUCCESS)
   {
--- a/lm75bd/lm75bd.h
+++ b/lm75bd/lm75bd.h
@@ -1,11 +1,11 @@
 #pragma once
 
-#include "errors.h"
+#include "../sys/errors.h"
 
 #include <stdint.h>
 
 /* LM75BD I2C Device Address */
-#define LM75BD_OBC_I2C_ADDR /* Define the address here */
+#define LM75BD_OBC_I2C_ADDR 0x4F/* Define the address here */
 
 /* LM75BD Configuration Values */
 #define LM75BD_DEV_OP_MODE_NORMAL 0x00U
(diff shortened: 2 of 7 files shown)


lm75bd/lm75bd.c
```diff
@@ -1,6 +1,6 @@
 #include "lm75bd.h"
 #include "i2c_io.h"
-#include "errors.h"
+#include "../sys/errors.h"
```
Navtajh04: you shouldn't need to do a relative include here since the cmake is already set up with the proper include directories

lm75bd/lm75bd.c
```diff
@@ -26,9 +26,43 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
...
+{
   /* Implement this driver function */
-  
+  if (!temp)
+  {
+    return ERR_CODE_INVALID_ARG;
+  }
+
+  uint8_t tempPointerRegister = 0x00;
+  uint8_t buff[2] = {0};
```
Navtajh04, line 38: I would declare this variable after i2cSendTo since you don't need it right now so it's a bit confusing to be declaring it here, especially since if i2cSendTo fails there is no need for this variable

lm75bd/lm75bd.c
```diff
@@ -26,9 +26,43 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
...
+
+  error_code_t errCode;
+  errCode = i2cSendTo(devAddr, &tempPointerRegister, 1);
+  if (errCode != ERR_CODE_SUCCESS)
+  {
+    return errCode;
+  }
+  error_code_t errCode2;
```
Navtajh04: you can just use the same errCode and reassign to it

lm75bd/lm75bd.h
```diff
@@ -1,11 +1,11 @@
 #pragma once
 
-#include "errors.h"
+#include "../sys/errors.h"
```
Navtajh04: same comment as eariler

services/controller/controller.c
```diff
@@ -2,7 +2,7 @@
 #include "console.h"
 #include "thermal_mgr.h"
 #include "lm75bd.h"
-#include "errors.h"
+#include "../../sys/errors.h"
```
Navtajh04: same comment as earlier

services/thermal_mgr/thermal_mgr.c
```diff
@@ -1,5 +1,5 @@
 #include "thermal_mgr.h"
-#include "errors.h"
+#include "../../sys/errors.h"
```
Navtajh04: same comment as earlier

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +43,49 @@ void initThermalSystemManager(lm75bd_config_t *config) {
 
 error_code_t thermalMgrSendEvent(thermal_mgr_event_t *event) {
   /* Send an event to the thermal manager queue */
+  if(!event){
+    return ERR_CODE_INVALID_ARG;
+  }
+
```
Navtajh04, line 49: check to make sure that the queue has already been created

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +43,49 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
 
+  if(xQueueSend(thermalMgrQueueHandle, (const void *) event, 0) != pdTRUE){
+    return ERR_CODE_INVALID_QUEUE_MSG;
```
Navtajh04: this would fail in the case of the queue being full or the queue having not been created only, not for an invalid queue message since the queue itself has no way to check whether or not it is a valid message

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +43,49 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
 static void thermalMgr(void *pvParameters) {
   /* Implement this task */
+  thermal_mgr_event_t event;
+  float temp;
+
   while (1) {
-    
+    if(xQueueReceive(thermalMgrQueueHandle, &event, portMAX_DELAY) == pdTRUE){
+      if(event.type == THERMAL_MGR_EVENT_MEASURE_TEMP_CMD ){
+        readTempLM75BD(LM75BD_OBC_I2C_ADDR, &temp);
```
Navtajh04: check return value

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +43,49 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+        addTemperatureTelemetry(temp);
+      }
+      else if(event.type == THERMAL_MGR_EVENT_INTERRUPT){
+        if(temp<80){
```
Navtajh04, line 90: you need to read the temperature before making a decision about whether it is safe or overtemperature since ur previous value could potentially be 5 seconds old

sys/i2c/i2c_io.h
```diff
@@ -1,6 +1,6 @@
 #pragma once
 
-#include "errors.h"
+#include "../errors.h"
```
Navtajh04: same as earlier, this applies to everywhere this was changed

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +43,49 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+          safeOperatingConditions();
+        }
+        else{
+          overTemperatureDetected();
+        }
+      }
+    }  
+
+    else{
+      printConsole("Invalid Thermal Event Type");
```
Navtajh04: only time we would enter this block is if the `xQueueReceive` times out so the error message should reflect that. Although we technically will never enter here since the block time is `portMAX_DELAY`

Navtajh04 (changes requested):
Looks good, just a few things

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +43,62 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+        if(errCode != ERR_CODE_SUCCESS){
+          break;
+        }
+        addTemperatureTelemetry(temp);
+      }
+
+      else if(event.type == THERMAL_MGR_EVENT_INTERRUPT){
+        errCode = readTempLM75BD(LM75BD_OBC_I2C_ADDR, &temp);
+        if(errCode != ERR_CODE_SUCCESS){
+          break;
```
Navtajh04: just continue if there is an error, you should not be breaking out of the while(1) loop

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +43,62 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+  float temp;
+  error_code_t errCode;
   while (1) {
-    
+    if(xQueueReceive(thermalMgrQueueHandle, &event, portMAX_DELAY) == pdTRUE){
+
+      if(event.type == THERMAL_MGR_EVENT_MEASURE_TEMP_CMD ){
+        errCode = readTempLM75BD(LM75BD_OBC_I2C_ADDR, &temp);
+        if(errCode != ERR_CODE_SUCCESS){
+          break;
```
Navtajh04: just continue if there is an error, you should not be breaking out of the while(1) loop

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +43,62 @@ void initThermalSystemManager(lm75bd_config_t *config) {
 
 error_code_t thermalMgrSendEvent(thermal_mgr_event_t *event) {
   /* Send an event to the thermal manager queue */
+  if(!event){
+    return ERR_CODE_INVALID_ARG;
+  }
+
+
+  if (!thermalMgrQueueHandle) {
+    return ERR_CODE_INVALID_QUEUE_MSG;
```
Navtajh04: this error is if we have not created the queue yet so it should be `ERR_CODE_INVALID_STATE` not invalid queue msg

lm75bd/lm75bd.c
Navtajh04: I don't think this was addressed

Navtajh04 (changes requested):

Navtajh04 (commented):
LGTM

PR #77, opened 2023-08-17
Finished Onboarding

Yarik-Popov:
# Purpose
- Ignore the *.sh in the .gitignore I used for testing scripts, didn't want to copy and paste commands into the CLI

# New Changes
- Explain new changes

# Testing

# Outstanding Changes


lm75bd/lm75bd.c
```diff
@@ -28,6 +28,31 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
 
 error_code_t readTempLM75BD(uint8_t devAddr, float *temp) {
   /* Implement this driver function */
+  // Initialize local variables
+  uint8_t buffer[READ_BYTES];
```
dgobalak: Initialize the buffer to 0s

lm75bd/lm75bd.c
```diff
@@ -28,6 +28,31 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
 
 error_code_t readTempLM75BD(uint8_t devAddr, float *temp) {
   /* Implement this driver function */
+  // Initialize local variables
```
dgobalak: Check if temp is NULL

lm75bd/lm75bd.h
```diff
@@ -5,7 +5,11 @@
 #include <stdint.h>
 
 /* LM75BD I2C Device Address */
-#define LM75BD_OBC_I2C_ADDR /* Define the address here */
+#define LM75BD_OBC_I2C_ADDR 0b1001111
+#define TEMPERATURE_ADDR 0x00U
+#define WRITE_BYTES 1U
+#define READ_BYTES 2U
+#define TEMPERATURE_CONVERSION_FACTOR 0.125f
```
dgobalak: These should be defined in the `.c` file to minimize the "scope" of the macro. In general, only include things in the header if they should be accessible to the user of the module.

lm75bd/lm75bd.c
```diff
@@ -28,6 +28,31 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
...
+  uint8_t buffer[READ_BYTES];
+  uint8_t tempReg = TEMPERATURE_ADDR;
+
+  // Set temperature registry
+  if (i2cSendTo(devAddr, &tempReg, WRITE_BYTES) != ERR_CODE_SUCCESS) return ERR_CODE_INVALID_ARG;
+
+  // read from temperature registry
+  if (i2cReceiveFrom(devAddr, buffer, READ_BYTES) != ERR_CODE_SUCCESS) return ERR_CODE_INVALID_ARG;
```
dgobalak: Instead of returning `ERR_CODE_INVALID_ARG`, you should return whatever `i2cSendTo` and `i2cReceiveFrom` returned cuz they may return a different error code.

lm75bd/lm75bd.c
```diff
@@ -28,6 +28,31 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
...
+
+  // Calculate temperature
+  if (msb == 0) {
+    *temp = buffer[0] + (buffer[1] * TEMPERATURE_CONVERSION_FACTOR);
+  } else {
+    buffer[1] = (~buffer[1]) + 1;
+    buffer[1] = (buffer[1] == 8) ? 0 : buffer[1];
+    buffer[0] = (~buffer[0]) + (buffer[1] == 0 ? 1 : 0);
+    *temp = -buffer[0] - buffer[1] * TEMPERATURE_CONVERSION_FACTOR;
+  }
```
dgobalak: This logic is more complicated than it needs to be. It could be as simple as:
```
```

services/thermal_mgr/thermal_mgr.h
```diff
@@ -3,9 +3,11 @@
 #include "lm75bd.h"
 #include "errors.h"
 
+#define TEMPERATURE_WAIT_TIME pdMS_TO_TICKS(10)
```
dgobalak: Define this in the `.c` file since it's not needed by the user

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +43,47 @@ void initThermalSystemManager(lm75bd_config_t *config) {
 
 error_code_t thermalMgrSendEvent(thermal_mgr_event_t *event) {
   /* Send an event to the thermal manager queue */
+  if(event == NULL) return ERR_CODE_INVALID_ARG;
+
+  if(xQueueSend(thermalMgrQueueHandle, (void *) event, TEMPERATURE_WAIT_TIME) != pdTRUE) {
+    return ERR_CODE_QUEUE_FULL;
+  }
```
dgobalak: Check if the queue has already been created (i.e. check if `thermalMgrQueueHandle` is NULL). If it hasn't been created, then return an invalid state error code.

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +43,47 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
 }
 
 void osHandlerLM75BD(void) {
-  /* Implement this function */
+  thermal_mgr_event_t event = {.type = THERMAL_OS_HANDLE};
+  BaseType_t xHigherPriorityTaskWoken = pdFALSE;
+  while(1) {
+   if(xQueueSendFromISR(thermalMgrQueueHandle, (void *) &event, &xHigherPriorityTaskWoken) == pdTRUE)
+    break;
+  }
```
dgobalak: We definitely don't want a possible infinite loop here since it runs in an ISR. The ISR's execution time must be kept as low as possible. So remove the infinite loop. For now, it's fine if `xQueueSendFromISR` fails.

Also, the queue will remain full until the thermal manager reads from it. However, this won't happen while the ISR is running. So I'd expect this to be a guaranteed infinite loop if it was actually executing from an ISR.

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +43,47 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+    return ERR_CODE_QUEUE_FULL;
+  }
 
   return ERR_CODE_SUCCESS;
 }
 
 void osHandlerLM75BD(void) {
-  /* Implement this function */
+  thermal_mgr_event_t event = {.type = THERMAL_OS_HANDLE};
+  BaseType_t xHigherPriorityTaskWoken = pdFALSE;
```
dgobalak: You're currently not making use of this variable. If a higher priority task is unblocked, you need to call `taskYIELD_FROM_ISR ` at the end of the ISR. Take a look at https://www.freertos.org/a00119.html

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +43,47 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+   if(xQueueSendFromISR(thermalMgrQueueHandle, (void *) &event, &xHigherPriorityTaskWoken) == pdTRUE)
+    break;
+  }
 }
 
 static void thermalMgr(void *pvParameters) {
-  /* Implement this task */
+  // Set variables and get config for lm75bd
+  thermal_mgr_event_t event;
+  float temp;
```
dgobalak: Limit the scope of this variable. It's not used outside the `if (xQueueReceive(thermalMgrQueueHandle, (void *) &event, TEMPERATURE_WAIT_TIME) == pdTRUE)` block.

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +43,47 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+
+      // Part 3 of challenge 
+      } else if(event.type == THERMAL_OS_HANDLE) {
+        if(temp > config->hysteresisThresholdCelsius) {
+          overTemperatureDetected();
+        } else {
+          safeOperatingConditions();
+        }
+      }
+    }
```
dgobalak: Handle the case where an invalid event is received (i.e. `.type` is an incorrect value)

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +43,47 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
-  /* Implement this task */
+  // Set variables and get config for lm75bd
+  thermal_mgr_event_t event;
+  float temp;
+  lm75bd_config_t *config = (lm75bd_config_t *) pvParameters;
+  
   while (1) {
-    
+    if (xQueueReceive(thermalMgrQueueHandle, (void *) &event, TEMPERATURE_WAIT_TIME) == pdTRUE) {
+      readTempLM75BD(config->devAddr, &temp);
```
dgobalak: Check return value so you know that the temperature was read successfully. This is important to do before you use the `temp` variable

services/thermal_mgr/thermal_mgr.h
```diff
@@ -3,9 +3,11 @@
...
+
 typedef enum {
-  THERMAL_MGR_EVENT_MEASURE_TEMP_CMD,
-  
+  THERMAL_MGR_EVENT_MEASURE_TEMP_CMD = 0,
+  THERMAL_OS_HANDLE = 1,
```
dgobalak: Add the `THERMAL_MGR_EVENT_` prefix. It's important to give anything defined in a header file a name that prevents naming conflicts with other modules (i.e. try to make the name unique). In this case, we do this using the prefix.

dgobalak (changes requested):

services/thermal_mgr/thermal_mgr.c
Navtajh04: you should not be reading the temperature without checking the received message from the queue first since if it was invalid it does not make sense to read the temperature anyways

services/thermal_mgr/thermal_mgr.h
Navtajh04: This also does not need to be this short of a block time since you receive events to read temp every 5 minutes

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +43,47 @@ void initThermalSystemManager(lm75bd_config_t *config) {
 
 error_code_t thermalMgrSendEvent(thermal_mgr_event_t *event) {
   /* Send an event to the thermal manager queue */
+  if(event == NULL) return ERR_CODE_INVALID_ARG;
+
+  if(xQueueSend(thermalMgrQueueHandle, (void *) event, TEMPERATURE_WAIT_TIME) != pdTRUE) {
```
Navtajh04: I would use separate macros for block times for queue send and receive

Navtajh04 (changes requested):

lm75bd/lm75bd.c
Yarik-Popov: I wrote this because I got confused by the datasheet saying to take the two's complement of the buffer. Yeah this is a lot better

Yarik-Popov: I wrote this because I got confused by the datasheet saying to take the two's complement of the buffer. Yeah this is a lot better

dgobalak (commented):
lgtm

lm75bd/lm75bd.c
```diff
@@ -1,74 +1,97 @@
...
+
+  uint8_t buffer[READ_BYTES] = {0};
+  uint8_t tempReg = LM75BD_REG_TEMP;
+
+  // Set temperature registry
+  error_code_t error1 = i2cSendTo(devAddr, &tempReg, WRITE_BYTES);
+  if (error1 != ERR_CODE_SUCCESS) return error1;
+
+  // read from temperature registry
+  error_code_t error2 = i2cReceiveFrom(devAddr, buffer, READ_BYTES);
```
Navtajh04: you do not need to declare a new error code variable here, you can just reassign to the old one

Navtajh04 (changes requested):
just one small thing

Navtajh04 (commented):
LGTM

PR #87, opened 2023-09-09
Onboarding Level 3

nbrodbec:


services/thermal_mgr/thermal_mgr.c
```diff
@@ -20,52 +20,80 @@ static StackType_t thermalMgrTaskStack[THERMAL_MGR_STACK_SIZE];
...
+    interruptEvent.type = THERMAL_MGR_EVENT_INTERRUPT_CMD;
+    (void)thermalMgrSendEvent(&interruptEvent);
 }
 
 static void thermalMgr(void *pvParameters) {
-  /* Implement this task */
-  while (1) {
-    
-  }
+    if (pvParameters == NULL) return;
```
dgobalak: In FreeRTOS, tasks shouldn't return (see https://www.freertos.org/implementing-a-FreeRTOS-task.html). For now, you can assume that `pvParameters` will be valid. Or if you want, you can call `vTaskDelete` to properly exit the function.

services/thermal_mgr/thermal_mgr.c
```diff
@@ -20,52 +20,80 @@ static StackType_t thermalMgrTaskStack[THERMAL_MGR_STACK_SIZE];
...
+
+    lm75bd_config_t config = *(lm75bd_config_t *)pvParameters;
+    thermal_mgr_event_t taskEvent;
+    error_code_t taskStatus;
+    float temp = 0;
```
dgobalak: You can limit the scope of these variables since it looks like they'd only be needed inside the `if (xQueueReceive(thermalMgrQueueHandle, &taskEvent, portMAX_DELAY) == pdTRUE) {` block.

lm75bd/lm75bd.c
```diff
@@ -1,74 +1,90 @@
...
-  
-  return ERR_CODE_SUCCESS;
+    if (temp == NULL) return ERR_CODE_INVALID_ARG;
+
+    uint8_t regTemp = LM75BD_REG_TEMP;
+    error_code_t writeErrCode = i2cSendTo(devAddr, &regTemp, 1);
+    if (writeErrCode != ERR_CODE_SUCCESS) return writeErrCode;
+
+    uint8_t readBuf[2] = {0};
+    error_code_t readErrCode = i2cReceiveFrom(devAddr, readBuf, 2);
```
dgobalak: You don't need to define 2 separate error code variables. We typically define a general error code variable like `error_code_t errCode` at the top of a function and use it throughout the function

dgobalak (changes requested):
Looks really good. Just a few comments

dgobalak (approved):
lgtm

PR #88, opened 2023-09-09
Onboarding Challenge Level 3 Completed

sardormirk:


lm75bd/lm75bd.h
```diff
@@ -5,7 +5,9 @@
 #include <stdint.h>
 
 /* LM75BD I2C Device Address */
-#define LM75BD_OBC_I2C_ADDR /* Define the address here */
+#define LM75BD_OBC_I2C_ADDR 0b01001111
+#define TEMP_ADD 0b00000000
+#define READ_BUF_SIZE 2
```
dgobalak: We only define things in a header file if they're required by the user of the module. Since these macros are only used internally by the driver, define them in the `.c` file

lm75bd/lm75bd.h
```diff
@@ -5,7 +5,9 @@
 #include <stdint.h>
 
 /* LM75BD I2C Device Address */
-#define LM75BD_OBC_I2C_ADDR /* Define the address here */
+#define LM75BD_OBC_I2C_ADDR 0b01001111
+#define TEMP_ADD 0b00000000
```
dgobalak: In the `.c` file, you'll see the macro defined for the config register. Try to follow the same naming convention. Something like `LM75BD_REG_TEMP` should be good.

lm75bd/lm75bd.h
dgobalak: Define a more appropriate name for `READ_BUF_SIZE`. Think about why the buffer size is 2. Something like `TEMP_REG_SIZE_B` (`_B` to say the size is in bytes) would be better

lm75bd/lm75bd.c
```diff
@@ -26,15 +29,47 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
...
+  uint16_t tempBytes = ((buffer[0] << 8) | buffer[1]) >> 5;
+
+  if (!(buffer[0] & 0b10000000))
+  {
+    *temp = tempBytes * 0.125f;
+  }
+  else
+  {
+    *temp = -(((~tempBytes) + 1) & 0x07FF) * 0.125f;
+  }
```
dgobalak: Note that this is a little more complicated than it needs to be. It could instead be done with:
```
```

services/thermal_mgr/thermal_mgr.c
```diff
@@ -21,51 +21,100 @@ static StackType_t thermalMgrTaskStack[THERMAL_MGR_STACK_SIZE];
...
+      thermalMgrQueueStorageArea, &thermalMgrQueueBuffer);
 }
 
-error_code_t thermalMgrSendEvent(thermal_mgr_event_t *event) {
+error_code_t thermalMgrSendEvent(thermal_mgr_event_t *event)
+{
   /* Send an event to the thermal manager queue */
 
+  if (thermalMgrQueueHandle == NULL)
+    return ERR_CODE_UNKNOWN;
```
dgobalak: A more appropriate error code would be `ERR_CODE_INVALID_STATE`

services/thermal_mgr/thermal_mgr.c
```diff
@@ -21,51 +21,100 @@ static StackType_t thermalMgrTaskStack[THERMAL_MGR_STACK_SIZE];
...
   return ERR_CODE_SUCCESS;
 }
 
-void osHandlerLM75BD(void) {
-  /* Implement this function */
+void osHandlerLM75BD(void)
+{
+
+  float temp;
+  readTempLM75BD(LM75BD_OBC_I2C_ADDR, &temp);
```
dgobalak: In an ISR, you shouldn't call any function that could take too long or call an RTOS function with a non-zero block time. Since `osHandlerLM75BD` is assumed to execute from within an ISR, there are 2 problems here:
1. I2C transfers are pretty slow. If an ISR's execution time is too long, it's more likely that you'll delay other interrupts.
2. `readTempLM75BD` calls `i2cSendTo` and `i2cReceiveFrom`. Both attempt to take a mutex with a non-zero block time. 

So you shouldn't call `readTempLM75BD` from `osHandlerLM75BD`

services/thermal_mgr/thermal_mgr.c
```diff
@@ -21,51 +21,100 @@ static StackType_t thermalMgrTaskStack[THERMAL_MGR_STACK_SIZE];
...
+
+  else if (temp <= LM75BD_DEFAULT_HYST_THRESH)
+    safeOperatingConditions();
 }
 
-static void thermalMgr(void *pvParameters) {
+static void thermalMgr(void *pvParameters)
+{
+
+  float temp = 0;
```
dgobalak: In general, limit the scope of variables as much as possible. You can limit the scope of `temp`.

services/thermal_mgr/thermal_mgr.c
```diff
@@ -21,51 +21,100 @@ static StackType_t thermalMgrTaskStack[THERMAL_MGR_STACK_SIZE];
...
   /* Implement this task */
-  while (1) {
-    
+
+  printConsole("Hello\n");
+
+  while (1)
+  {
+
+    if (xQueueReceive(thermalMgrQueueHandle, &eventItem, (TickType_t)10) == pdTRUE)
```
dgobalak: Since we don't expect items to appear in the queue that often, you can increase the block time to a few seconds (instead of 10ms)

dgobalak (changes requested):

lm75bd/lm75bd.c
```diff
@@ -26,15 +31,40 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
...
+  if (sendErr != ERR_CODE_SUCCESS)
+  {
+    return sendErr;
+  }
+
+  error_code_t readErr = i2cReceiveFrom(devAddr, &buffer, 2);
+  if (readErr != ERR_CODE_SUCCESS)
+  {
+    return readErr;
+  }
```
Navtajh04: you can just use the same error code for both of these instead of creating a new error code variable each time

Navtajh04 (changes requested):
Looks good, just one small change

Navtajh04 (commented):
LGTM

PR #89, opened 2023-09-10
Firmware Onboarding

MostafaH04:


lm75bd/lm75bd.c
```diff
@@ -28,7 +28,31 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
...
+    return sendErr;
+  }
+
+  // read:
+  uint8_t tempBuf[2];
+  error_code_t recivErr = i2cReceiveFrom(devAddr, &tempBuf, 2);
+  if (recivErr != ERR_CODE_SUCCESS)
+  {
+    return recivErr;
+  }
```
Navtajh04: just use the same error code variable instead of creating a new one each time

lm75bd/lm75bd.h
```diff
@@ -21,6 +21,19 @@
...
+  TEMP_REG  = 0x00U,
+  CONF_REG  = 0x01U,
+  THYST_REG = 0x02U,
+  TOS_REG   = 0x03U
+}pointer_value_type_t;
+
+// ratio between temp value and temp
+#define TEMP_CONVERSION 0.125f
+
+// Personal Defines End --------------
```
Navtajh04: move these into the .c file since these macros are only used by the driver and don't need to be accessed by other files

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +43,68 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
   /* Send an event to the thermal manager queue */
+  if (thermalMgrQueueHandle == NULL)
+    return ERR_CODE_INVALID_STATE;
 
+  if (event == NULL)
+    return ERR_CODE_INVALID_ARG;
+
+  if( xQueueSend(thermalMgrQueueHandle,
+                ( void * )event,
+                ( TickType_t ) 10) == errQUEUE_FULL)
```
Navtajh04: since this function is used in an ISR the block time should be 0

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +43,68 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+  }
   return ERR_CODE_SUCCESS;
 }
 
 void osHandlerLM75BD(void) {
   /* Implement this function */
+  float tempC;
+  thermal_mgr_event_t eventCall;
+
+  error_code_t errTemp = readTempLM75BD(LM75BD_OBC_I2C_ADDR, &tempC);
```
Navtajh04: you shouldn't be reading the temperature in an ISR
1. the temperature read uses I2C which is relatively slow and we want the ISR to be as quick as possible
2. this function call will eventually call one of the I2C helper functions, which will try to take the I2C bus mutex and block

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +43,68 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+  if (errEvent != ERR_CODE_SUCCESS)
+    printConsole("Error sending event in OS Handler");
+    //logging for good measure
 }
 
 static void thermalMgr(void *pvParameters) {
   /* Implement this task */
+  float tempC = 0;
+  thermal_mgr_event_t eventIn;
+  error_code_t errIn;
```
Navtajh04: in general try to limit the scope of your variables as much as possible

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +43,68 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+
   while (1) {
-    
+    if(xQueueReceive(thermalMgrQueueHandle,
+                     &eventIn,
+                     (TickType_t) 10) == pdTRUE)
```
Navtajh04: you could make this block time longer since your only expecting to receive an event every couple of seconds

Navtajh04 (changes requested):

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +43,67 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
 }
 
 void osHandlerLM75BD(void) {
   /* Implement this function */
+  thermal_mgr_event_t eventCall;
+  eventCall.type = OS_INTERRUPT_EVENT;
+  
+  error_code_t errEvent = thermalMgrSendEvent(&eventCall);
+  if (errEvent != ERR_CODE_SUCCESS)
+    printConsole("Error sending event in OS Handler");
```
Navtajh04: you should never print directly from an ISR since this function call tries to take a mutex and printing in general isn't reentrant

Navtajh04: If an error happens here, just ignore it

services/thermal_mgr/thermal_mgr.h
```diff
@@ -5,7 +5,7 @@
 
 typedef enum {
   THERMAL_MGR_EVENT_MEASURE_TEMP_CMD,
-  
+  OS_INTERRUPT_EVENT
```
Navtajh04: its good style to keep ur prefixs for an enum consistent like naming this `THERMAL_MGR_EVENT_OS_INTERRUPT` or something

Navtajh04 (changes requested):
Just 2 more small changes

Navtajh04 (commented):
LGTM

PR #93, opened 2023-09-17
Completed Bootcamp

orekk:
# Purpose
- Implemented Temperature Sensor Driver Functions.
- Created Thermal Management Task.
- Added OS Handler for temperature sensor interrupts.

# Changes
- Defined LM75BD sensor's I2C device address.
- Implemented readTempLM75BD function.
- Created Thermal Management Task to collect temperature data.
- Added OS Handler for LM75BD temperature sensor interrupts.

# Testing
- Unit tests passed for implemented functions.

# Outstanding Changes
None.


lm75bd/lm75bd.c
```diff
@@ -26,9 +26,69 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
...
+error_code_t readTempLM75BD(uint8_t devAddr, float *temp) { // Question 1. what do I use temp for, if theregister I am using is 0x00
   /* Implement this driver function */
-  
+    error_code_t errCode;
+
+    if (temp == NULL) return ERR_CODE_INVALID_ARG;
+
+    // to implement this we need to select the sensor's internaltemeprature register (7.4.1)
+	  uint8_t pointer_register_value = 0x00; // Point to the Temperature Register
+		devAddr = LM75BD_OBC_I2C_ADDR;
```
Navtajh04: nit: fix the indenting

lm75bd/lm75bd.c
```diff
@@ -26,9 +26,69 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
...
+    //     *temp = (temp_data >> 5) * 0.125; 
+    // }else{
+    //     // temperature is negative 
+    //     temp_data = ~temp_data + 1; // Convert to positive representation
+    //     int16_t temp_int = (int16_t)(temp_data >> 5);
+    //     *temp = -((temp_int) >> 5) * 0.125; 
+    // }
+
+    const uint16_t sign_bit_mask = 0x0400;
+    const uint16_t temperature_bits_mask = 0x03FF;
```
Navtajh04: use macros for these instead of const variables

lm75bd/lm75bd.c
```diff
@@ -26,9 +26,69 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
...
+		// once selected we can read temperature from the sensor 
+		// errCode = i2cSendTo(LM75BD_OBC_I2C_ADDR, pointer_register_value, 2); 
+
+    // error_code_t i2cSendTo(uint8_t sAddr, uint8_t *buf, uint16_t numBytes);
+
+    // error_code_t i2cReceiveFrom(uint8_t sAddr, uint8_t *buf, uint16_t numBytes);
+
+    
+    // uint16_t numBytes = 2; 
+    errCode = i2cSendTo(devAddr,&pointer_register_value, 1U); // read the temperature from the sensor
```
Navtajh04: not need to explicitly typecast the constant

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +46,46 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+  // Question: a, I using the correct devAddr 
+    error_code_t errCode = readTempLM75BD(LM75BD_OBC_I2C_ADDR, temperature); // Use your readTempLM75BD function
+
+    if (temperature > HYS_THRESHOLD) {
+        // Temperature is above or equal to the overtemperature threshold
+        overTemperatureDetected();
+    } else if (temperature <= HYS_THRESHOLD) {
+        // Temperature is below or equal to the hysteresis threshold
+        safeOperatingConditions(); 
+    }
```
Navtajh04: we should not be calling readTemp in an ISR for 2 reasons
1. readTemp will call the i2c helper functions which will block on the I2C bus mutex and we should never be trying to block in an ISR
2. I2C is relatively slow and we want ISRs to always be as short and quick as possible to keep our system responsive

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +46,46 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+  if (send != pdPASS) return ERR_CODE_INVALID_QUEUE_MSG/*failed case*/;
   return ERR_CODE_SUCCESS;
 }
 
-void osHandlerLM75BD(void) {
+void osHandlerLM75BD(void) { 
   /* Implement this function */
+  float* temperature;
+  // Question: a, I using the correct devAddr 
+    error_code_t errCode = readTempLM75BD(LM75BD_OBC_I2C_ADDR, temperature); // Use your readTempLM75BD function
```
Navtajh04: since you are declaring temperature to be a pointer to a float but not setting it to anything, when you pass it into this function it is a NULL pointer so readTemp will return an error

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +46,46 @@ void initThermalSystemManager(lm75bd_config_t *config) {
 
 error_code_t thermalMgrSendEvent(thermal_mgr_event_t *event) {
   /* Send an event to the thermal manager queue */
-
```
Navtajh04, line 46: check if event is null

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +46,46 @@ void initThermalSystemManager(lm75bd_config_t *config) {
 
 error_code_t thermalMgrSendEvent(thermal_mgr_event_t *event) {
   /* Send an event to the thermal manager queue */
```
Navtajh04, line 49: check to make sure queue has been created

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +46,46 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
-
+  BaseType_t send = xQueueSend(thermalMgrQueueHandle, event, 0); // 0 ticks to wait 
+  if (send != pdPASS) return ERR_CODE_INVALID_QUEUE_MSG/*failed case*/;
```
Navtajh04: you can just do `if(xQueueSend(thermalMgrQueueHandle, event, 0) != pdPASS)` instead of declaring a new variable

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +46,46 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+			 // Check if it's the temperature measure event
+       // Kemi: Obtain type how
+			 if (receivedEvent.type == THERMAL_MGR_EVENT_MEASURE_TEMP_CMD) {
+			     // collect temperature data using drive fcn 
+			     float* temperature; 
+           // Kemi: Obtain devAddr Question, is this the correct devAddr
+					 error_code_t errCode = readTempLM75BD(LM75BD_OBC_I2C_ADDR, temperature);
+					 if (errCode != ERR_CODE_SUCCESS) /*do something*/;
+			     // Send it as telemetry
+			     addTemperatureTelemetry(*temperature);
```
Navtajh04: you shuld only be adding the telemetry temperature if the read was successful

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +46,46 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+    // Question , what do I declare this as? 
+    thermal_mgr_event_t receivedEvent;
+
+    // Task should only perform an action if it receives an event through the thermal manager queue
+		if (xQueueReceive(thermalMgrQueueHandle, &receivedEvent, portMAX_DELAY) == pdPASS) { // set the addresses for this 
+			 // Check if it's the temperature measure event
+       // Kemi: Obtain type how
+			 if (receivedEvent.type == THERMAL_MGR_EVENT_MEASURE_TEMP_CMD) {
+			     // collect temperature data using drive fcn 
+			     float* temperature;
```
Navtajh04: same comment as in the os handler

Navtajh04 (changes requested):
Looks good, just a few things, but I would try to fix the indentation to make it easier to read and review

lm75bd/lm75bd.c
```diff
@@ -26,9 +29,68 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
...
+error_code_t readTempLM75BD(uint8_t devAddr, float *temp) { // Question 1. what do I use temp for, if theregister I am using is 0x00
   /* Implement this driver function */
-  
+    error_code_t errCode;
+
+    if (temp == NULL) return ERR_CODE_INVALID_ARG;
+
+    // to implement this we need to select the sensor's internaltemeprature register (7.4.1)
+	  uint8_t pointer_register_value = 0x00; // Point to the Temperature Register
+		devAddr = LM75BD_OBC_I2C_ADDR;
```
Navtajh04: device address is passed into the function, so you don't need to set it again. You can just use the address that was passed in, this is because if we have multiple lm75bd temperature sensors each would have a different address, so we would want the address to be a parameter to choose which sensor we want to read from

services/thermal_mgr/thermal_mgr.c
```diff
@@ -41,19 +44,61 @@ void initThermalSystemManager(lm75bd_config_t *config) {
 
 }
 
+
 error_code_t thermalMgrSendEvent(thermal_mgr_event_t *event) {
   /* Send an event to the thermal manager queue */
-
+  if(event == NULL) return ERR_CODE_INVALID_ARG; 
+  if(thermalMgrQueueHandle == NULL) return ERR_CODE_INVALID_QUEUE_MSG;
+  if (xQueueSend(thermalMgrQueueHandle, event, 0) != pdPASS) return ERR_CODE_INVALID_QUEUE_MSG/*failed case*/;
```
Navtajh04: this function would only fail if the queue is full, not due to an invalid queue msg

services/thermal_mgr/thermal_mgr.c
```diff
@@ -41,19 +44,61 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+			     // Send it as telemetry
+           // only add telemetry temperature if the reading was a success
+			     if(errCode == ERR_CODE_SUCCESS) addTemperatureTelemetry(temperature);
+			  }
+        if (receivedEvent.type == THERMAL_MGR_EVENT_OS_INTERRUPT) {
+			     // Kemi: Note that, if you decide to do this, you’ll need to define a new event type to send to the thermal manager queue.
+          // Team Lead -> u dont necessarily need to read the temperature in the OShandler, u just need a way to communicate that an 
+          // interrupt has happened so that u can read the temperature and deal with it from outside to OS handler
+          float temperature; 
+          error_code_t errCode = readTempLM75BD(LM75BD_OBC_I2C_ADDR, &temperature); // Use your readTempLM75BD function
```
Navtajh04, line 92: check whether the readTempLM75BD was successful or not

Navtajh04 (changes requested):
Just a couple more small changes

Navtajh04 (commented):
LGTM

PR #98, opened 2023-10-11
Onboarding Challenge Attempt

ziedmasmoudi:
# Purpose
Firmware onboarding complete

# New Changes
- Implemented Temperature Sensor Driver Functions
- Implemented Thermal Management Task
- Implemented OS handler

# Testing
- Unit tests ran successfully
- Integration behavior is as expected

# Outstanding Changes
None

### DIFF
--- a/lm75bd/lm75bd.c
+++ b/lm75bd/lm75bd.c
@@ -29,20 +29,19 @@ error_code_t readTempLM75BD(uint8_t devAddr, float *temp) {
   /* Implement this driver function */
   error_code_t errCode;
   
-  uint8_t reg_addr = 0x0U;
-  errCode = i2cSendTo(LM75BD_OBC_I2C_ADDR, &reg_addr, 1U);
-  if (errCode != ERR_CODE_SUCCESS) return errCode;
+  uint8_t regAddr = 0x0U;
+  RETURN_IF_ERROR_CODE(i2cSendTo(LM75BD_OBC_I2C_ADDR, &regAddr, 1U));
 
-  uint8_t buff[2] = {0};
-  errCode = i2cReceiveFrom(LM75BD_OBC_I2C_ADDR, buff, 2U);
-  if (errCode != ERR_CODE_SUCCESS) return errCode;
+  uint8_t tempRegData[2] = {0};
+  RETURN_IF_ERROR_CODE(i2cReceiveFrom(LM75BD_OBC_I2C_ADDR, tempRegData, 2U));
 
-  uint16_t temp_data = (buff[0] << 3) | (buff[1] >> 5);
+  uint16_t tempData = (tempRegData[0] << 3) | (tempRegData[1] >> 5);
 
-  if (buff[0] >> 7){
-    *temp = -((-temp_data) & 2047) * 0.125;
+  if (tempRegData[0] >> 7){
+    //take the negative of temp data and keep 11 LSB bits to reverse 2's complement
+    *temp = -((-tempData) & 0b11111111111) * 0.125;
   } else {
-    *temp = temp_data * 0.125;
+    *temp = tempData * 0.125;
   }
 
   return ERR_CODE_SUCCESS;
--- a/lm75bd/lm75bd.c
+++ b/lm75bd/lm75bd.c
@@ -27,7 +27,24 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
 
 error_code_t readTempLM75BD(uint8_t devAddr, float *temp) {
   /* Implement this driver function */
+  error_code_t errCode;
   
+  uint8_t reg_addr = 0x0U;
+  errCode = i2cSendTo(LM75BD_OBC_I2C_ADDR, &reg_addr, 1U);
+  if (errCode != ERR_CODE_SUCCESS) return errCode;
+
+  uint8_t buff[2] = {0};
+  errCode = i2cReceiveFrom(LM75BD_OBC_I2C_ADDR, buff, 2U);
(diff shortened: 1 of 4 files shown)


lm75bd/lm75bd.c
```diff
@@ -27,7 +27,24 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
 
 error_code_t readTempLM75BD(uint8_t devAddr, float *temp) {
   /* Implement this driver function */
+  error_code_t errCode;
   
+  uint8_t reg_addr = 0x0U;
```
Navtajh04: we use `camelCase` foe variable names

lm75bd/lm75bd.c
```diff
@@ -27,7 +27,24 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
...
+  errCode = i2cSendTo(LM75BD_OBC_I2C_ADDR, &reg_addr, 1U);
+  if (errCode != ERR_CODE_SUCCESS) return errCode;
+
+  uint8_t buff[2] = {0};
+  errCode = i2cReceiveFrom(LM75BD_OBC_I2C_ADDR, buff, 2U);
+  if (errCode != ERR_CODE_SUCCESS) return errCode;
```
Navtajh04: use `RETURN_IF_ERROR_CODE(function)` to ensure this error gets logged as well as to maintain consistency with our actual codebase

lm75bd/lm75bd.c
```diff
@@ -27,7 +27,24 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
 
 error_code_t readTempLM75BD(uint8_t devAddr, float *temp) {
   /* Implement this driver function */
+  error_code_t errCode;
   
+  uint8_t reg_addr = 0x0U;
+  errCode = i2cSendTo(LM75BD_OBC_I2C_ADDR, &reg_addr, 1U);
+  if (errCode != ERR_CODE_SUCCESS) return errCode;
+
+  uint8_t buff[2] = {0};
```
Navtajh04: it would be better to make this variable name more descriptive

lm75bd/lm75bd.c
```diff
@@ -27,7 +27,24 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
...
+  errCode = i2cReceiveFrom(LM75BD_OBC_I2C_ADDR, buff, 2U);
+  if (errCode != ERR_CODE_SUCCESS) return errCode;
+
+  uint16_t temp_data = (buff[0] << 3) | (buff[1] >> 5);
+
+  if (buff[0] >> 7){
+    *temp = -((-temp_data) & 2047) * 0.125;
```
Navtajh04: make the 2047 a macro with a descriptive name or add a comment explaining this line to avoid magic numbers

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +43,39 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
 error_code_t thermalMgrSendEvent(thermal_mgr_event_t *event) {
   /* Send an event to the thermal manager queue */
-
-  return ERR_CODE_SUCCESS;
+  if (event == NULL)
+    return ERR_CODE_INVALID_QUEUE_MSG;
+  if (xQueueSend(thermalMgrQueueHandle, event, (TickType_t )0) == pdTRUE)
+    return ERR_CODE_SUCCESS;
+  else
+    return ERR_CODE_UNKNOWN;
```
Navtajh04: if `xQueueSend` fails its because the queue is full so that error code should be used instead

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +43,39 @@ void initThermalSystemManager(lm75bd_config_t *config) {
 
 error_code_t thermalMgrSendEvent(thermal_mgr_event_t *event) {
   /* Send an event to the thermal manager queue */
-
-  return ERR_CODE_SUCCESS;
```
Navtajh04, line 47: check to make sure that the queue has already been created

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +43,39 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
 static void thermalMgr(void *pvParameters) {
   /* Implement this task */
+  lm75bd_config_t params = *(lm75bd_config_t *)pvParameters;
+  thermal_mgr_event_t event;
+
   while (1) {
-    
+    if (xQueueReceive(thermalMgrQueueHandle, &event, (TickType_t )0) == pdTRUE){
+      float temp = 0;
+      readTempLM75BD(params.devAddr, &temp);
```
Navtajh04: you should be checking the event before reading temperature, since if the event was something other than a read temp or interrupt event like if it was an invalid event for example you wouldn't need to read the temperature

Navtajh04: Also check the return value

Navtajh04 (changes requested):

services/thermal_mgr/thermal_mgr.c
ziedmasmoudi: just by checking that the pointer isn't null? like this?
  if (thermalMgrQueueHandle == NULL)
    return ERR_CODE_UNKNOWN;

Navtajh04: Yup

Navtajh04 (commented):
LGTM

PR #99, opened 2023-10-13
Onboarding

MikeKatsnelson:
# Purpose

# New Changes
- Explain new changes

# Testing

# Outstanding Changes

### DIFF
--- a/lm75bd/lm75bd.c
+++ b/lm75bd/lm75bd.c
@@ -6,6 +6,9 @@
 #include <string.h>
 #include <math.h>
 
+//temp
+#include <stdio.h>
+
 /* LM75BD Registers (p.8) */
 #define LM75BD_REG_CONF 0x01U  /* Configuration Register (R/W) */
 
@@ -29,6 +32,37 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
 error_code_t readTempLM75BD(uint8_t devAddr, float *temp) {
   /* Implement this driver function */
   
+  // Store value of pointer register that points to temp. register
+  uint8_t writeBuffer = 0x00;
+
+  // Set pointer register to point to temp. register
+  i2cSendTo(devAddr, &writeBuffer, 1); 
+
+  // Store temp. value retrieved from sensor
+  uint8_t readBuffer[2];
+
+  // Read data from temp. register
+  i2cReceiveFrom(devAddr, &readBuffer[0], 2);
+
+  // Convert temp. data to Celsius
+  uint16_t tempC;
+  float multiplier = 0.125;
+
+  //1) Make into abs. value
+  if ( readBuffer[0] > 128 ) {
+    multiplier = -0.125;
+    readBuffer[1] *= -1; 
+    readBuffer[0] *= -1;
+  }
+
+  // 2) Concatenate MSB with LSB
+  tempC = ((readBuffer[0] ) << 8) | (readBuffer[1]);
+
+  // 3) Bitshift by 5 ignored bits
+  tempC = tempC >> 5;
+
+  *temp = tempC * multiplier;
+
   return ERR_CODE_SUCCESS;
 }
 
--- a/lm75bd/lm75bd.c
+++ b/lm75bd/lm75bd.c
@@ -7,11 +7,11 @@
 #include <string.h>
 #include <math.h>
 
-//temp
-#include <stdio.h>
-
 /* LM75BD Registers (p.8) */
 #define LM75BD_REG_CONF 0x01U  /* Configuration Register (R/W) */
+
+// Multiplier to convert temp. register value to Celsius
+#define LM75BD_TEMP_C_MULTIPLIER 0.125
 
 error_code_t lm75bdInit(lm75bd_config_t *config) {
   error_code_t errCode;
--- a/lm75bd/lm75bd.c
+++ b/lm75bd/lm75bd.c
@@ -30,37 +30,31 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
 
 error_code_t readTempLM75BD(uint8_t devAddr, float *temp) {
   /* Implement this driver function */
-  
+  error_code_t errCode;
+
   // Store value of pointer register that points to temp. register
   uint8_t writeBuffer = 0x00;
 
   // Set pointer register to point to temp. register
-  i2cSendTo(devAddr, &writeBuffer, 1); 
+  RETURN_IF_ERROR_CODE(i2cSendTo(devAddr, &writeBuffer, 1));
 
   // Store temp. value retrieved from sensor
-  uint8_t readBuffer[2];
+  uint8_t readBuffer[2] = {0};
 
   // Read data from temp. register
-  i2cReceiveFrom(devAddr, &readBuffer[0], 2);
+  RETURN_IF_ERROR_CODE(i2cReceiveFrom(devAddr, &readBuffer[0], 2));
 
   // Convert temp. data to Celsius
-  uint16_t tempC;
-  float multiplier = 0.125;
+  int16_t tempC;
 
-  //1) Make into abs. value
-  if ( readBuffer[0] > 128 ) {
-    multiplier = -0.125;
-    readBuffer[1] *= -1; 
-    readBuffer[0] *= -1;
-  }
-
-  // 2) Concatenate MSB with LSB
+  // 1) Concatenate MSB with LSB
   tempC = ((readBuffer[0] ) << 8) | (readBuffer[1]);
 
-  // 3) Bitshift by 5 ignored bits
+  // 2) Bitshift by 5 ignored bits
   tempC = tempC >> 5;
 
-  *temp = tempC * multiplier;
+  // 3) Multiply by multiplier
+  *temp = tempC * LM75BD_TEMP_C_MULTIPLIER;
 
   return ERR_CODE_SUCCESS;
 }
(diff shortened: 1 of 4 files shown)


lm75bd/lm75bd.c
```diff
@@ -29,6 +32,37 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
...
+  uint8_t writeBuffer = 0x00;
+
+  // Set pointer register to point to temp. register
+  i2cSendTo(devAddr, &writeBuffer, 1); 
+
+  // Store temp. value retrieved from sensor
+  uint8_t readBuffer[2];
+
+  // Read data from temp. register
+  i2cReceiveFrom(devAddr, &readBuffer[0], 2);
```
Navtajh04: you should be using `RETURN_IF_ERROR_CODE` to catch any errors that may happen when calling the i2c functions

lm75bd/lm75bd.c
```diff
@@ -29,6 +32,37 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
...
   /* Implement this driver function */
   
+  // Store value of pointer register that points to temp. register
+  uint8_t writeBuffer = 0x00;
+
+  // Set pointer register to point to temp. register
+  i2cSendTo(devAddr, &writeBuffer, 1); 
+
+  // Store temp. value retrieved from sensor
+  uint8_t readBuffer[2];
```
Navtajh04: in general its best to initialize your arrays to 0 like `uint8_t readBuffer[2] = {0};`

lm75bd/lm75bd.c
```diff
@@ -29,6 +32,37 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
...
+
+  // Read data from temp. register
+  i2cReceiveFrom(devAddr, &readBuffer[0], 2);
+
+  // Convert temp. data to Celsius
+  uint16_t tempC;
+  float multiplier = 0.125;
```
Navtajh04: make this a macro instead since its a constant

lm75bd/lm75bd.c
```diff
@@ -6,6 +6,9 @@
 #include <string.h>
```
Navtajh04: your simulated test run looks wrong

lm75bd/lm75bd.c
```diff
@@ -29,6 +32,37 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
...
+
+  // Read data from temp. register
+  i2cReceiveFrom(devAddr, &readBuffer[0], 2);
+
+  // Convert temp. data to Celsius
+  uint16_t tempC;
+  float multiplier = 0.125;
+
+  //1) Make into abs. value
+  if ( readBuffer[0] > 128 ) {
```
Navtajh04: better for readability to explicitly use a bit mask when your trying to check any bit ie (readBuffer[0] & YOUR_BIT_MASK_MACRO_NAME)

lm75bd/lm75bd.c
```diff
@@ -29,6 +32,37 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
...
+    multiplier = -0.125;
+    readBuffer[1] *= -1; 
+    readBuffer[0] *= -1;
+  }
```
Navtajh04: I dont think these lines are needed since the temperature sensor outputs a 2s compliment number anyways

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +43,39 @@ void initThermalSystemManager(lm75bd_config_t *config) {
 
 error_code_t thermalMgrSendEvent(thermal_mgr_event_t *event) {
   /* Send an event to the thermal manager queue */
-
```
Navtajh04, line 46: check if event is null

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +43,39 @@ void initThermalSystemManager(lm75bd_config_t *config) {
 
 error_code_t thermalMgrSendEvent(thermal_mgr_event_t *event) {
   /* Send an event to the thermal manager queue */
```
Navtajh04, line 44: check to make sure queue has already been created

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +43,39 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
-
+  xQueueSend(thermalMgrQueueHandle, event, 0);
```
Navtajh04: check the return value to see if the function call failed or succeeded

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +43,39 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
   return ERR_CODE_SUCCESS;
 }
 
 void osHandlerLM75BD(void) {
   /* Implement this function */
+  thermal_mgr_event_t newEvent;
+  newEvent.type = THERMAL_MGR_EVENT_CHECK_TEMP_CMD;
```
Navtajh04: this event name sounds like the same as the existing one so it should be renamed to mention interrupt

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +43,39 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
   while (1) {
-    
+    if ( xQueueReceive(thermalMgrQueueHandle, &eventBuffer, 10) == pdPASS) {
+      float temp;
+      if (readTempLM75BD(LM75BD_OBC_I2C_ADDR, &temp) == ERR_CODE_SUCCESS) {
+        if (eventBuffer.type == THERMAL_MGR_EVENT_MEASURE_TEMP_CMD) {
+          addTemperatureTelemetry(temp);
+        }
+      }
+      if (eventBuffer.type == THERMAL_MGR_EVENT_MEASURE_TEMP_CMD) {
```
Navtajh04: this is checking the same as the previous if statement

Navtajh04: You want to check if the event was sent from the OSHandler after an interrupt

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +43,39 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+        if (temp >= 75) {
```
Navtajh04: you should read the temperature first instead of relying on an old value since it may be outdated and if an interrupt has triggered then for sure the temperature is now different than it was previously

Navtajh04 (changes requested):

services/thermal_mgr/thermal_mgr.c
MikeKatsnelson: The queue is being created in initThermalSystemManager(). Should I be editing this function? And if I wanted to make sure the queue has been created, how would I do so?

services/thermal_mgr/thermal_mgr.c
MikeKatsnelson: In my new code, I checked if it the event points to a null pointer. I made the function return ERR_CODE_UNKNOWN if the event points to a null pointer or xQeueueSend() fails. Is this an okay approach?

services/thermal_mgr/thermal_mgr.c
Navtajh04: you can just check in this function if `thermalMgrQueueHandle == NULL` since if it's not created it will be initialized to NULL since it's a static variable

services/thermal_mgr/thermal_mgr.c
Navtajh04: it's best to deal with each error case separately so that your error codes can be as specific as possible, so for example if event is NULL you would want to return `ERR_CODE_INVALID_ARG` or if the queue send fails you would want to return `ERR_CODE_QUEUE_FULL`

services/thermal_mgr/thermal_mgr.c
```diff
@@ -38,23 +38,47 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
 }
 
 static void thermalMgr(void *pvParameters) {
   /* Implement this task */
+  thermal_mgr_event_t eventBuffer;
   while (1) {
-    
+    if (xQueueReceive(thermalMgrQueueHandle, &eventBuffer, 10) == pdPASS) {
+      float temp;
+      if (readTempLM75BD(LM75BD_OBC_I2C_ADDR, &temp) == ERR_CODE_SUCCESS) {
```
Navtajh04: check the received command before reading the temperature since if the command is something other than measure temp or interrupt like if it's an invalid cmd for example it doesn't make sure to read the temperature anyways

lm75bd/lm75bd.h
```diff
@@ -21,6 +21,9 @@
 #define LM75BD_DEFAULT_OT_THRESH 80.0f
 #define LM75BD_DEFAULT_HYST_THRESH 75.0f
 
+// Multiplier to convert temp. register value to Celsius
+#define LM75BD_TEMP_C_MULTIPLIER 0.125
```
Navtajh04: this should go in the .c file since it's only used internally within the driver and doesn't need to be accessible to other modules that try to include the driver through this .h file

Navtajh04 (changes requested):
Looks good, just a couple more changes

Navtajh04 (commented):
LGTM

PR #101, opened 2023-11-02

Shrey-Varma:
# Purpose

# New Changes
Finished implementing readTempLM75BD() , thermal_mgr task, and osHandlerLM75BD()

# Testing
Passed all unit tests

### DIFF
--- a/lm75bd/lm75bd.c
+++ b/lm75bd/lm75bd.c
@@ -29,17 +29,20 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
 }
 
 error_code_t readTempLM75BD(uint8_t devAddr, float *temp) {
-  /* Implement this driver function */
+  // check if temp is allocated
+  if (!temp)
+    return ERR_CODE_NULL_ARG;
 
   // set pointer register to temperature
-  uint8_t point_reg = 0x00;
-  uint8_t buf[1] = {point_reg};
-  i2cSendTo(devAddr, buf, 1);
-
+  uint8_t buf[1] = {LM75BD_OBC_TEMP_REG};
+  error_code_t errCode;
+  RETURN_IF_ERROR_CODE(i2cSendTo(devAddr, buf, 1));
   // read data from sensor
-  uint8_t temp_bytes[2] = {};
-  i2cReceiveFrom(devAddr, temp_bytes, 2);
-  int16_t combined = ((temp_bytes[0] << 8) | temp_bytes[1]);
+  uint8_t tempBytes[2] = {0};
+  RETURN_IF_ERROR_CODE(i2cReceiveFrom(devAddr, tempBytes, 2));
+
+  // convert bytes to temperature
+  int16_t combined = ((tempBytes[0] << 8) | tempBytes[1]);
   combined = combined >> 5;
   *temp = combined * 0.125;
   return ERR_CODE_SUCCESS;
--- a/lm75bd/lm75bd.c
+++ b/lm75bd/lm75bd.c
@@ -1,22 +1,25 @@
 #include "lm75bd.h"
-#include "i2c_io.h"
+#include "console.h"
 #include "errors.h"
+#include "i2c_io.h"
 #include "logging.h"
 
+#include <math.h>
 #include <stdint.h>
 #include <string.h>
-#include <math.h>
 
 /* LM75BD Registers (p.8) */
-#define LM75BD_REG_CONF 0x01U  /* Configuration Register (R/W) */
+#define LM75BD_REG_CONF 0x01U /* Configuration Register (R/W) */
 
 error_code_t lm75bdInit(lm75bd_config_t *config) {
   error_code_t errCode;
 
-  if (config == NULL) return ERR_CODE_INVALID_ARG;
+  if (config == NULL)
+    return ERR_CODE_INVALID_ARG;
 
-  RETURN_IF_ERROR_CODE(writeConfigLM75BD(config->devAddr, config->osFaultQueueSize, config->osPolarity,
-                                         config->osOperationMode, config->devOperationMode));
+  RETURN_IF_ERROR_CODE(writeConfigLM75BD(
+      config->devAddr, config->osFaultQueueSize, config->osPolarity,
+      config->osOperationMode, config->devOperationMode));
 
   // Assume that the overtemperature and hysteresis thresholds are already set
   // Hysteresis: 75 degrees Celsius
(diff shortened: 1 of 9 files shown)


lm75bd/lm75bd.c
```diff
@@ -27,13 +30,25 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
 
 error_code_t readTempLM75BD(uint8_t devAddr, float *temp) {
   /* Implement this driver function */
-  
+
```
Navtajh04, line 35: check if temp is NULL

lm75bd/lm75bd.c
```diff
@@ -27,13 +30,25 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
...
+  // set pointer register to temperature
+  uint8_t point_reg = 0x00;
```
Navtajh04: variable names should be camelCase

Navtajh04: Also this can just be a macro instead of a var

lm75bd/lm75bd.c
```diff
@@ -27,13 +30,25 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
...
+  uint8_t buf[1] = {point_reg};
+  i2cSendTo(devAddr, buf, 1);
```
Navtajh04: use `RETURN_IF_ERROR_CODE` to make sure this error gets dealt with if `i2cSendTo` fails

lm75bd/lm75bd.c
```diff
@@ -27,13 +30,25 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
...
+
+  // read data from sensor
+  uint8_t temp_bytes[2] = {};
```
Navtajh04: in general its a good idea to explicitly initialize buffers to 0 using `= {0}`

Navtajh04: Also variable names should be camelCase

services/thermal_mgr/thermal_mgr.c
```diff
@@ -20,41 +23,58 @@ static StackType_t thermalMgrTaskStack[THERMAL_MGR_STACK_SIZE];
...
-    thermalMgrQueueStorageArea, &thermalMgrQueueBuffer);
-
+  thermalMgrQueueHandle =
+      xQueueCreateStatic(THERMAL_MGR_QUEUE_LENGTH, THERMAL_MGR_QUEUE_ITEM_SIZE,
+                         thermalMgrQueueStorageArea, &thermalMgrQueueBuffer);
 }
 
 error_code_t thermalMgrSendEvent(thermal_mgr_event_t *event) {
   /* Send an event to the thermal manager queue */
-
```
Navtajh04, line 46: check if event is NULL

services/thermal_mgr/thermal_mgr.c
```diff
@@ -20,41 +23,58 @@ static StackType_t thermalMgrTaskStack[THERMAL_MGR_STACK_SIZE];
...
-  return ERR_CODE_SUCCESS;
```
Navtajh04, line 47: check to make sure queue has already been created

services/thermal_mgr/thermal_mgr.c
```diff
@@ -20,41 +23,58 @@ static StackType_t thermalMgrTaskStack[THERMAL_MGR_STACK_SIZE];
...
+  if (xQueueSend(thermalMgrQueueHandle, event, (TickType_t)10) == pdTRUE)
```
Navtajh04: since this function is used in an ISR, you should never be blocking, so block time should be 0

services/thermal_mgr/thermal_mgr.c
```diff
@@ -20,41 +23,58 @@ static StackType_t thermalMgrTaskStack[THERMAL_MGR_STACK_SIZE];
...
+    if (xQueueReceive(thermalMgrQueueHandle, &event, (TickType_t)10) ==
+        pdTRUE) {
+      if (event.type == THERMAL_MGR_EVENT_MEASURE_TEMP_CMD) {
+        float val;
+        float *temp = &val;
+        readTempLM75BD(details.devAddr, temp);
+        if (*temp > details.hysteresisThresholdCelsius)
+          overTemperatureDetected();
+        else
+          safeOperatingConditions();
```
Navtajh04: You should only be calling these functions when we either had an OT event or returned to a safe operating temp, but right now if we're at a safe operating temp and we read temp normally (like for telemetry data or something or general housekeeping) then we'd be saying we returned to safe operating conditions which isn't true

Navtajh04: Measure temp event should be different from the interrupt event so that we can deal with these 2 cases distinctly

services/thermal_mgr/thermal_mgr.c
```diff
@@ -20,41 +23,58 @@ static StackType_t thermalMgrTaskStack[THERMAL_MGR_STACK_SIZE];
...
+  lm75bd_config_t details = *(lm75bd_config_t *)pvParameters;
   while (1) {
-    
+    thermal_mgr_event_t event;
+    if (xQueueReceive(thermalMgrQueueHandle, &event, (TickType_t)10) ==
+        pdTRUE) {
+      if (event.type == THERMAL_MGR_EVENT_MEASURE_TEMP_CMD) {
+        float val;
+        float *temp = &val;
+        readTempLM75BD(details.devAddr, temp);
```
Navtajh04: log this and continue to the next loop iteration if this failed

Navtajh04 (changes requested):

services/thermal_mgr/thermal_mgr.c
Navtajh04: I dont think this was addressed

services/thermal_mgr/thermal_mgr.c
```diff
@@ -20,41 +24,70 @@ static StackType_t thermalMgrTaskStack[THERMAL_MGR_STACK_SIZE];
...
+      if (event.type == THERMAL_MGR_EVENT_MEASURE_TEMP_CMD ||
+          event.type == THERNAL_MGR_EVENT_CHECK_OS) {
+        float val;
+        error_code_t errCode = readTempLM75BD(details.devAddr, &val);
+        if (errCode != ERR_CODE_SUCCESS) {
+          LOG_ERROR_CODE(errCode);
+          // ensure that if CHECK_OS event was sent, that the temperature gets
+          // if it failed the first time attempt to CHECK_OS again
+          if (event.type == THERNAL_MGR_EVENT_CHECK_OS)
+            thermalMgrSendEvent(&event);
```
Navtajh04: Ignore handling errors for the sake of this onboarding. The problem with this approach is that if there is some kind of permanent issue causing the readTemp function to fail this task would essentially enter an infinite loop where it keeps sending itself an event and then trying to read the temp and failing because of that event then trying again

services/thermal_mgr/thermal_mgr.c
```diff
@@ -20,41 +24,70 @@ static StackType_t thermalMgrTaskStack[THERMAL_MGR_STACK_SIZE];
...
+          continue;
+        }
+        if (event.type == THERNAL_MGR_EVENT_CHECK_OS) {
+          if (val > details.hysteresisThresholdCelsius)
+            overTemperatureDetected();
+          else
+            safeOperatingConditions();
+        }
+        addTemperatureTelemetry(val);
```
Navtajh04: this should only  be added if the event received is THERMAL_MGR_EVENT_MEASURE_TEMP_CMD

Navtajh04 (changes requested):
Just a couple more changes

Navtajh04 (commented):
LGTM

PR #102, opened 2023-11-03
finished onboarding challenge

andrei-paraschiv:
# Purpose

# New Changes
- Explain new changes

# Testing

# Outstanding Changes


lm75bd/lm75bd.c
```diff
@@ -27,6 +27,22 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
 
 error_code_t readTempLM75BD(uint8_t devAddr, float *temp) {
   /* Implement this driver function */
+  // Does general rule 4 of the style guide matter for this test and in what situation would they be used?
```
Navtajh04: check to make sure temp isnt NULL

lm75bd/lm75bd.c
```diff
@@ -27,6 +27,22 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
...
+  error_code_t errCode; 
+
+  uint8_t pointerTempValue = 0;
+  uint8_t pointerRegisterSize = 1;
+  RETURN_IF_ERROR_CODE(i2cSendTo(LM75BD_OBC_I2C_ADDR, &pointerTempValue, pointerRegisterSize));
+
+  uint8_t tempRegisterSize = 2;
+  uint8_t tempBuff[2] = {0};  
+  RETURN_IF_ERROR_CODE(i2cReceiveFrom(LM75BD_OBC_I2C_ADDR, tempBuff, tempRegisterSize));
+  // Does this need to be repeated twice at startup since first read is incorrect or does it not matter for this onboarding test? 7.4.1
```
Navtajh04: it's fine to ignore it for the purpose of the onboarding

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +43,41 @@ void initThermalSystemManager(lm75bd_config_t *config) {
 
 error_code_t thermalMgrSendEvent(thermal_mgr_event_t *event) {
```
Navtajh04, line 45: check if event is NULL

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +43,41 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
   /* Send an event to the thermal manager queue */
```
Navtajh04, line 46: check to make sure the queue has already been created by making sure the handle is not NULL

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +43,41 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+  if (xQueueSend(thermalMgrQueueHandle, event, 1) != pdTRUE) return ERR_CODE_QUEUE_FULL; // timeout?
```
Navtajh04: since we are calling this function in an ISR, the block time should be set to 0

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +43,41 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
 
   return ERR_CODE_SUCCESS;
 }
 
 void osHandlerLM75BD(void) {
   /* Implement this function */
+  thermal_mgr_event_t event;
+  event.type = THERMAL_MGR_EVENT_INTERRUPT;
+  if (thermalMgrSendEvent(&event) != ERR_CODE_SUCCESS) ; // What should be done if error?
```
Navtajh04: we don't have any way to log the error in the onboarding currently from an ISR, but in our actual code base you would use `LOG_ERROR_CODE_FROM_ISR`

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +43,41 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
 
 static void thermalMgr(void *pvParameters) {
   /* Implement this task */
+  lm75bd_config_t config = *(lm75bd_config_t *)pvParameters;
+  thermal_mgr_event_t event;
+  float temp;
+
   while (1) {
-    
+    if (xQueueReceive(thermalMgrQueueHandle, &event, 0) == pdTRUE) { // timeout?
```
Navtajh04: since you aren't doing anything else in this case if the queue receive fails its fine to set it to `portMAX_DELAY`

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +43,41 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+      switch (event.type) {
+        case THERMAL_MGR_EVENT_MEASURE_TEMP_CMD:
+          if (readTempLM75BD(config.devAddr, &temp) != ERR_CODE_SUCCESS) break; // What should be done if error?
```
Navtajh04: Log the error if there was an error

Navtajh04: Also nitpick but I would explicitly use `continue` to make it more clear and also in case something is added after the switch case for some reason in the future (I don't see why this would happen though)

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +43,41 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+          addTemperatureTelemetry(temp);
+          break;
+        case THERMAL_MGR_EVENT_INTERRUPT:
+          if (readTempLM75BD(config.devAddr, &temp) != ERR_CODE_SUCCESS) break;
```
Navtajh04: log if this is an error code and `continue;`

Navtajh04 (changes requested):
Looks good, just a couple of minor changes

services/thermal_mgr/thermal_mgr.c
andrei-paraschiv: I'm not sure I understand, I believe continue doesn't work in a switch case?

services/thermal_mgr/thermal_mgr.c
andrei-paraschiv: Same as above

Navtajh04 (commented):
LGTN

PR #103, opened 2023-11-08
Completed Onboarding Challenge

SeanZhang425:
# Purpose
Onboarding

# New Changes
Implemented temperature sensor driver function, thermal management task, and OS handler.

# Testing
Passed all unit and integration tests


lm75bd/lm75bd.c
```diff
@@ -26,8 +26,24 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
...
 
 error_code_t readTempLM75BD(uint8_t devAddr, float *temp) {
-  /* Implement this driver function */
+  // declare error code variable
+  error_code_t errCode;
   
+  // set pointer register to 'temperature sensor'
+  uint8_t writeBuf[] = {0};
+  errCode = i2cSendTo(devAddr, writeBuf, 1);
+  RETURN_IF_ERROR_CODE(errCode);
```
Navtajh04: these lines can be combined to just be `RETURN_IF_ERROR_CODE(i2cSendTo(devAddr, writeBuf, 1);`

lm75bd/lm75bd.c
```diff
@@ -26,8 +26,24 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
...
+
+  // read temperature from sensor
+  uint8_t readBuf[2] = {0};
+  errCode = i2cReceiveFrom(devAddr, readBuf, 2);
+  RETURN_IF_ERROR_CODE(errCode);
```
Navtajh04: same as above

lm75bd/lm75bd.c
```diff
@@ -26,8 +26,24 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
 }
 
 error_code_t readTempLM75BD(uint8_t devAddr, float *temp) {
-  /* Implement this driver function */
+  // declare error code variable
+  error_code_t errCode;
   
+  // set pointer register to 'temperature sensor'
+  uint8_t writeBuf[] = {0};
```
Navtajh04: since you know this should only be 1 uint8_t long, its best to explicitly initialize it to that size for readability

services/thermal_mgr/thermal_mgr.c
```diff
@@ -42,19 +42,43 @@ void initThermalSystemManager(lm75bd_config_t *config) {
 }
 
 error_code_t thermalMgrSendEvent(thermal_mgr_event_t *event) {
-  /* Send an event to the thermal manager queue */
+  if (event != NULL) {
+    if (xQueueSend(thermalMgrQueueHandle, (void *) event, (TickType_t) 0) == pdPASS)
```
Navtajh04: make sure that the queue has already been created

services/thermal_mgr/thermal_mgr.c
```diff
@@ -42,19 +42,43 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
 
 static void thermalMgr(void *pvParameters) {
-  /* Implement this task */
+  thermal_mgr_event_t eventBuf;
+
   while (1) {
-    
+
+    float temp;
+    error_code_t errCode = readTempLM75BD(LM75BD_OBC_I2C_ADDR, &temp);
```
Navtajh04: receive and check the event first before reading temp, since right now even if you don't receive a measure temp command you'll be reading the temperature

services/thermal_mgr/thermal_mgr.c
```diff
@@ -42,19 +42,43 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
 void osHandlerLM75BD(void) {
-  /* Implement this function */
+  thermal_mgr_event_t event;
+  event.type = THERMAL_MGR_EVENT_INTERRUPT_CMD;
+  thermalMgrSendEvent(&event);
 }
 
 static void thermalMgr(void *pvParameters) {
-  /* Implement this task */
+  thermal_mgr_event_t eventBuf;
```
Navtajh04: in general try to limit the scope of your variables to the smallest scope possible, so move this into the while block

Navtajh04 (changes requested):

services/thermal_mgr/thermal_mgr.c
```diff
@@ -42,19 +42,46 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
   while (1) {
-    
+    thermal_mgr_event_t eventBuf;
+
+    float temp;
+
+    if (xQueueReceive(thermalMgrQueueHandle, &eventBuf, 10) == pdPASS) {
+      if (eventBuf.type == THERMAL_MGR_EVENT_MEASURE_TEMP_CMD) {
+        error_code_t errCode = readTempLM75BD(LM75BD_OBC_I2C_ADDR, &temp);
+        if (errCode == ERR_CODE_SUCCESS)
```
Navtajh04, line 72: make sure to log the error if it was no successful

services/thermal_mgr/thermal_mgr.c
```diff
@@ -42,19 +42,46 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+          addTemperatureTelemetry(temp);
+      }
+      else if (eventBuf.type == THERMAL_MGR_EVENT_INTERRUPT_CMD) {
+        error_code_t errCode = readTempLM75BD(LM75BD_OBC_I2C_ADDR, &temp);
```
Navtajh04, line 78: check if the function call was successful or not

Navtajh04 (changes requested):
Just 2 more quick changes

services/thermal_mgr/thermal_mgr.c
```diff
@@ -42,19 +42,52 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+        return ERR_CODE_SUCCESS;
+      else
+        return ERR_CODE_QUEUE_FULL;
+    }
+    else
+      return ERR_CODE_INVALID_ARG;
+  }
 
-  return ERR_CODE_SUCCESS;
+  return ERR_CODE_INVALID_QUEUE_MSG;
```
Navtajh04: If the thermal mgr queue handle is null, then it makes more sense to return ERR_CODE_INVALID_STATE instead since we are in a state where we have not intitialized the queue yet and it's not related to the queue msg itself

services/thermal_mgr/thermal_mgr.c
```diff
@@ -42,19 +42,52 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
 }
```
Navtajh04, line 55: Also note that when your doing multiple errror checks like this it's a lot cleaner to just use multiple if statements with early returns rather than a bunch of nested if statements. For example something like ```

if(errCodeCondition){ 
  return ERR_CODE
}
if(otherPossibleError){
  return OTHER_ERROR
}
// continue with the rest of the function knowing everything before was success without need an else statement
```

**Review by Navtajh04 — CHANGES_REQUESTED** [review]
Just one more quick change other than that looks good

File: Testing/Temporary/CTestCostData.txt [hunk 11]
```diff
@@ -0,0 +1 @@
+---
```
Navtajh04, line 1: your integration test run looks wrong

services/thermal_mgr/thermal_mgr.c
```diff
@@ -42,19 +42,50 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
 
 error_code_t thermalMgrSendEvent(thermal_mgr_event_t *event) {
-  /* Send an event to the thermal manager queue */
-
+  if (&thermalMgrQueueHandle != NULL)
```
Navtajh04: should be == NULL instead of !=, same for below

Navtajh04 (changes requested):
just a quick change

Navtajh04 (commented):
LGTM, ill add you to the github and then you can send me your email you I can add you to the notion and you can take a look through the FW task board and let me know a couple of tasks that interest you

PR #111, opened 2023-12-22
Completed Onboarding Challenge

Krisphy:
# Purpose
Onboarding complete!

# New Changes
- Thermal Sensor Driver, Events for Thermal Management

# Testing
- Integration and Unit Tests

# Outstanding Changes
-

### DIFF
--- a/lm75bd/lm75bd.c
+++ b/lm75bd/lm75bd.c
@@ -27,15 +27,14 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
 
 error_code_t readTempLM75BD(uint8_t devAddr, float *temp) {
   error_code_t errCode = ERR_CODE_INVALID_ARG;
-  if (temp == NULL) return errCode; //If arguments are problematic return early
-  uint8_t buf[2] = {0x00U, 0x00U}; 
-  uint8_t readBuf = 0x00;
-  RETURN_IF_ERROR_CODE(i2cSendTo(devAddr, &readBuf,1));
-  RETURN_IF_ERROR_CODE(i2cReceiveFrom(devAddr, buf, 2));
-  if (buf[0] & 0x80)
-    *temp = ((((~((((uint16_t)buf[0])<<3)|((uint16_t)(buf[1]>>5)))) & 0x07FF)+1)  & 0x07FF)*-0.125; //If negative temp, converts both numbers into 16 unsigned bits, then adds them, and does twos complement
-  else
-    *temp = ((((uint16_t)buf[0])<<3)|((uint16_t)(buf[1]>>5)))*0.125; //Combines the unsigned bytes and outputs the positive value
+  if (temp == NULL) return errCode; 
+  uint8_t readBuf[2] = {0}; 
+  uint8_t buf = LM75BD_DEV_OP_MODE_NORMAL;
+  RETURN_IF_ERROR_CODE(i2cSendTo(devAddr, &buf,1));
+  RETURN_IF_ERROR_CODE(i2cReceiveFrom(devAddr, readBuf, 2));
+int16_t val = (readBuf[0] << 8) | readBuf[1];
+*temp = (float)(val >> 5) * 0.125;
+
   return ERR_CODE_SUCCESS;
 }
 
--- a/lm75bd/lm75bd.c
+++ b/lm75bd/lm75bd.c
@@ -9,7 +9,7 @@
 
 /* LM75BD Registers (p.8) */
 #define LM75BD_REG_CONF 0x01U  /* Configuration Register (R/W) */
-
+#define LM75BD_REG_TEMP 0x00U
 error_code_t lm75bdInit(lm75bd_config_t *config) {
   error_code_t errCode;
 
@@ -26,15 +26,14 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
 }
 
 error_code_t readTempLM75BD(uint8_t devAddr, float *temp) {
-  error_code_t errCode = ERR_CODE_INVALID_ARG;
-  if (temp == NULL) return errCode; 
+  error_code_t errCode;
+  if (temp == NULL) return ERR_CODE_INVALID_ARG; 
   uint8_t readBuf[2] = {0}; 
-  uint8_t buf = LM75BD_DEV_OP_MODE_NORMAL;
+  uint8_t buf = LM75BD_REG_TEMP;
   RETURN_IF_ERROR_CODE(i2cSendTo(devAddr, &buf,1));
   RETURN_IF_ERROR_CODE(i2cReceiveFrom(devAddr, readBuf, 2));
-int16_t val = (readBuf[0] << 8) | readBuf[1];
-*temp = (float)(val >> 5) * 0.125;
-
+  int16_t val = (readBuf[0] << 8) | readBuf[1];
+  *temp = (float)(val >> 5) * 0.125;
   return ERR_CODE_SUCCESS;
 }
 
--- a/lm75bd/lm75bd.c
+++ b/lm75bd/lm75bd.c
@@ -26,8 +26,16 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
 }
 
 error_code_t readTempLM75BD(uint8_t devAddr, float *temp) {
-  /* Implement this driver function */
-  
+  error_code_t errCode = ERR_CODE_INVALID_ARG;
+  if (temp == NULL) return errCode;
+  uint8_t buf[2] = {0x00U, 0x00U};
+  uint8_t readBuf = 0x00;
+  RETURN_IF_ERROR_CODE(i2cSendTo(devAddr, &readBuf,1));
+  RETURN_IF_ERROR_CODE(i2cReceiveFrom(devAddr, buf, 2));
+  if (buf[0] & 0x80)
+    *temp = ((((~((((uint16_t)buf[0])<<3)|((uint16_t)(buf[1]>>5)))) & 0x07FF)+1)  & 0x07FF)*-0.125;
+  else
+    *temp = ((((uint16_t)buf[0])<<3)|((uint16_t)(buf[1]>>5)))*0.125;
   return ERR_CODE_SUCCESS;
 }
 
--- a/lm75bd/lm75bd.c
+++ b/lm75bd/lm75bd.c
@@ -27,15 +27,15 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
 
 error_code_t readTempLM75BD(uint8_t devAddr, float *temp) {
   error_code_t errCode = ERR_CODE_INVALID_ARG;
-  if (temp == NULL) return errCode;
-  uint8_t buf[2] = {0x00U, 0x00U};
+  if (temp == NULL) return errCode; //If arguments are problematic return early
+  uint8_t buf[2] = {0x00U, 0x00U}; 
   uint8_t readBuf = 0x00;
   RETURN_IF_ERROR_CODE(i2cSendTo(devAddr, &readBuf,1));
   RETURN_IF_ERROR_CODE(i2cReceiveFrom(devAddr, buf, 2));
   if (buf[0] & 0x80)
-    *temp = ((((~((((uint16_t)buf[0])<<3)|((uint16_t)(buf[1]>>5)))) & 0x07FF)+1)  & 0x07FF)*-0.125;
+    *temp = ((((~((((uint16_t)buf[0])<<3)|((uint16_t)(buf[1]>>5)))) & 0x07FF)+1)  & 0x07FF)*-0.125; //If negative temp, converts both numbers into 16 unsigned bits, then adds them, and does twos complement
   else
-    *temp = ((((uint16_t)buf[0])<<3)|((uint16_t)(buf[1]>>5)))*0.125;
+    *temp = ((((uint16_t)buf[0])<<3)|((uint16_t)(buf[1]>>5)))*0.125; //Combines the unsigned bytes and outputs the positive value
   return ERR_CODE_SUCCESS;
 }
 
--- a/services/thermal_mgr/thermal_mgr.c
+++ b/services/thermal_mgr/thermal_mgr.c
@@ -49,12 +49,12 @@ error_code_t thermalMgrSendEvent(thermal_mgr_event_t *event) {
   // If queue handle doesnt describe a queue, error in state
   if (thermalMgrQueueHandle == NULL) return ERR_CODE_INVALID_STATE;
   // If queue is full, cannot add event, error
-  if (xQueueSend(thermalMgrQueueHandle, event, 10) == errQUEUE_FULL) return ERR_CODE_QUEUE_FULL;
+  if (xQueueSend(thermalMgrQueueHandle, event, 0) == errQUEUE_FULL) return ERR_CODE_QUEUE_FULL;
   return ERR_CODE_SUCCESS;
 }
 
 void osHandlerLM75BD(void) {
-  thermal_mgr_event_t interrupt = {THERMAL_MGR_INTERRUPT_EVENT};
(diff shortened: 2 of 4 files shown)


lm75bd/lm75bd.c
```diff
@@ -26,8 +26,16 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
 }
 
 error_code_t readTempLM75BD(uint8_t devAddr, float *temp) {
-  /* Implement this driver function */
-  
+  error_code_t errCode = ERR_CODE_INVALID_ARG;
+  if (temp == NULL) return errCode; //If arguments are problematic return early
```
dgobalak: You can keep these comments for the challenge. For future reference, a lot of your comments are just repeating what a line of code says. For example, it's clear what this line does without the comment.

Comments are good for explaining _why_ you're doing something or how some complicated piece of logic works. Try not to write comments that don't add useful context. This is something you'll get better at with experience though.

lm75bd/lm75bd.c
```diff
@@ -26,8 +26,16 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
...
+  uint8_t buf[2] = {0x00U, 0x00U};
```
dgobalak: nit: initialize buffers to 0 with the shorthand syntax `uint8_t buf[2] = {0};`. In C, if an array is partially initialized, the rest of the elements are initialized to 0.

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +44,43 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
 error_code_t thermalMgrSendEvent(thermal_mgr_event_t *event) {
   /* Send an event to the thermal manager queue */
-
+  // If no event is given arguments are bad
+  if (event == NULL) return ERR_CODE_INVALID_ARG;
+  // If queue handle doesnt describe a queue, error in state
+  if (thermalMgrQueueHandle == NULL) return ERR_CODE_INVALID_STATE;
+  // If queue is full, cannot add event, error
+  if (uxQueueSpacesAvailable(thermalMgrQueueHandle) == 0) return ERR_CODE_QUEUE_FULL;
+  xQueueSend(thermalMgrQueueHandle, event, 10);
```
dgobalak: `xQueueSend` will fail and return an error code if the queue is full. So `uxQueueSpacesAvailable` is unnecessary. Here's the `xQueueSend` documentation: https://www.freertos.org/a00117.html

This also introduces a race condition. Imagine the queue becomes full after uxQueueSpacesAvailable already returns. That would result in xQueueSend failing and then this function mistakingly returning success.

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +44,43 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+  error_code_t errCode;
+  float tempReading = 0.0;
   while (1) {
-    
+    if (xQueueReceive(thermalMgrQueueHandle, &task, 10) == pdTRUE)
+    {
+      if (task.type == THERMAL_MGR_EVENT_MEASURE_TEMP_CMD)
+      {
+        //Imported the logging functions to use this function
+        RETURN_IF_ERROR_CODE(readTempLM75BD(LM75BD_OBC_I2C_ADDR, &tempReading));
```
dgobalak: `RETURN_IF_ERROR_CODE` shouldn't be used in a function with a `void` return type. The compiler caught this and threw a warning. The compiler is pretty good at catching some bugs, so it's important to look through the build output carefully.

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +44,43 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+  xQueueSend(thermalMgrQueueHandle, event, 10);
   return ERR_CODE_SUCCESS;
 }
 
 void osHandlerLM75BD(void) {
   /* Implement this function */
+  error_code_t errCode;
+  float tempReading = 0.0;
+  //Imported the logging functions to use this function
+  RETURN_IF_ERROR_CODE(readTempLM75BD(LM75BD_OBC_I2C_ADDR, &tempReading));
```
dgobalak: `RETURN_IF_ERROR_CODE` shouldn't be used in a function with a `void` return type. The compiler caught this and threw a warning. The compiler is pretty good at catching some bugs, so it's important to look through the build output carefully.

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +44,43 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
 }
 
 static void thermalMgr(void *pvParameters) {
   /* Implement this task */
+  thermal_mgr_event_t task;
+  error_code_t errCode;
+  float tempReading = 0.0;
   while (1) {
-    
+    if (xQueueReceive(thermalMgrQueueHandle, &task, 10) == pdTRUE)
```
dgobalak: If this task doesn't receive anything after 10 ticks, it just goes right back to waiting on the queue. So you could instead set the block time to `portMAX_DELAY`. This would make the task block indefinitely until an item is in the queue.

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +44,43 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+  RETURN_IF_ERROR_CODE(readTempLM75BD(LM75BD_OBC_I2C_ADDR, &tempReading));
+  if (tempReading>=80)
+    overTemperatureDetected();
+  else if (tempReading<=75)
+    safeOperatingConditions();
 }
 
 static void thermalMgr(void *pvParameters) {
   /* Implement this task */
+  thermal_mgr_event_t task;
```
dgobalak: `task` is a confusing variable name for an `event` or queue item.

lm75bd/lm75bd.c
```diff
@@ -26,8 +26,16 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
...
+  error_code_t errCode = ERR_CODE_INVALID_ARG;
+  if (temp == NULL) return errCode; //If arguments are problematic return early
+  uint8_t buf[2] = {0x00U, 0x00U}; 
+  uint8_t readBuf = 0x00;
+  RETURN_IF_ERROR_CODE(i2cSendTo(devAddr, &readBuf,1));
+  RETURN_IF_ERROR_CODE(i2cReceiveFrom(devAddr, buf, 2));
+  if (buf[0] & 0x80)
+    *temp = ((((~((((uint16_t)buf[0])<<3)|((uint16_t)(buf[1]>>5)))) & 0x07FF)+1)  & 0x07FF)*-0.125; //If negative temp, converts both numbers into 16 unsigned bits, then adds them, and does twos complement
+  else
+    *temp = ((((uint16_t)buf[0])<<3)|((uint16_t)(buf[1]>>5)))*0.125; //Combines the unsigned bytes and outputs the positive value
```
dgobalak: This isn't readable. Since the register value is already in 2s complement form, the conversion can be done like this:
```
```

lm75bd/lm75bd.c
```diff
@@ -26,8 +26,16 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
 }
 
 error_code_t readTempLM75BD(uint8_t devAddr, float *temp) {
-  /* Implement this driver function */
-  
+  error_code_t errCode = ERR_CODE_INVALID_ARG;
+  if (temp == NULL) return errCode; //If arguments are problematic return early
+  uint8_t buf[2] = {0x00U, 0x00U}; 
+  uint8_t readBuf = 0x00;
+  RETURN_IF_ERROR_CODE(i2cSendTo(devAddr, &readBuf,1));
```
dgobalak: Sending the contents of `readBuf` is confusing. I'd recommend changing that variable name. It'd make more sense for the buffer passed into `i2cReceiveFrom` to be called `readBuf`

Also, we have a #define for each register address. For example, you can find the configuration register address (LM75BD_REG_CONF) at the top of this file.

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +44,43 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
   /* Implement this function */
+  error_code_t errCode;
+  float tempReading = 0.0;
+  //Imported the logging functions to use this function
+  RETURN_IF_ERROR_CODE(readTempLM75BD(LM75BD_OBC_I2C_ADDR, &tempReading));
+  if (tempReading>=80)
+    overTemperatureDetected();
+  else if (tempReading<=75)
+    safeOperatingConditions();
 }
```
dgobalak: This handler is executed in an ISR so you shouldn't call readTempLM75BD. readTempLM75BD involves I2C read/writes which are slow and ISRs shouldn't take too long to execute.

Another way to do this would be to send an event to the thermal mgr and the mgr will handle it outside the ISR.

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +44,43 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
 
 static void thermalMgr(void *pvParameters) {
   /* Implement this task */
+  thermal_mgr_event_t task;
+  error_code_t errCode;
+  float tempReading = 0.0;
```
dgobalak: Try to limit the scope of variables. Since tempReading isn't used outside the while loop, you shouldn't declare it outside the while loop.

dgobalak (changes requested):
Good first pass-through. Just a few things to clean up

lm75bd/lm75bd.c
```diff
@@ -26,8 +26,15 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
 }
 
 error_code_t readTempLM75BD(uint8_t devAddr, float *temp) {
-  /* Implement this driver function */
-  
+  error_code_t errCode = ERR_CODE_INVALID_ARG;
+  if (temp == NULL) return errCode;
```
dgobalak: Don't initialize errCode to a non-success error code. Otherwise, it's easy for this to introduce a bug in the future when someone refactors the code. I'd just do:
```
if (temp == NULL) return ERR_CODE_INVALID_ARG;
```

lm75bd/lm75bd.c
```diff
@@ -26,8 +26,15 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
 }
 
 error_code_t readTempLM75BD(uint8_t devAddr, float *temp) {
-  /* Implement this driver function */
-  
+  error_code_t errCode = ERR_CODE_INVALID_ARG;
+  if (temp == NULL) return errCode; 
+  uint8_t readBuf[2] = {0}; 
+  uint8_t buf = LM75BD_DEV_OP_MODE_NORMAL;
```
dgobalak: LM75BD_DEV_OP_MODE_NORMAL isn't the correct macro for this. That macro is used to represent one of the operating modes of the device. You should be transmitting the temperature register address. You can define a macro for this (`LM75BD_REG_TEMP`) at the top of this file (next to the config register macro)

lm75bd/lm75bd.c
```diff
@@ -26,8 +26,15 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
...
+  RETURN_IF_ERROR_CODE(i2cSendTo(devAddr, &buf,1));
+  RETURN_IF_ERROR_CODE(i2cReceiveFrom(devAddr, readBuf, 2));
+int16_t val = (readBuf[0] << 8) | readBuf[1];
+*temp = (float)(val >> 5) * 0.125;
```
dgobalak: nit: indentation is off.

this isn't a big issue since we run an auto-formatter on our actual OBC-firmware codebase that would fix things like this.

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +44,42 @@ void initThermalSystemManager(lm75bd_config_t *config) {
 
 error_code_t thermalMgrSendEvent(thermal_mgr_event_t *event) {
   /* Send an event to the thermal manager queue */
-
+  // If no event is given arguments are bad
+  if (event == NULL) return ERR_CODE_INVALID_ARG;
+  // If queue handle doesnt describe a queue, error in state
+  if (thermalMgrQueueHandle == NULL) return ERR_CODE_INVALID_STATE;
+  // If queue is full, cannot add event, error
+  if (xQueueSend(thermalMgrQueueHandle, event, 10) == errQUEUE_FULL) return ERR_CODE_QUEUE_FULL;
```
dgobalak: Since `thermalMgrSendEvent` is called from `osHandlerLM75BD`, it's executed in an ISR. You should never block within an ISR, so the block time for `xQueueSend` should be 0.

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +44,42 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
   return ERR_CODE_SUCCESS;
 }
 
 void osHandlerLM75BD(void) {
-  /* Implement this function */
+  thermal_mgr_event_t interrupt = {THERMAL_MGR_INTERRUPT_EVENT};
```
dgobalak: C99 allows you to set elements of a struct by name when initializing. So it's more readable to do something like: `thermal_mgr_event_t interrupt = { .type=THERMAL_MGR_INTERRUPT_EVENT }`.

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +44,42 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+        addTemperatureTelemetry(tempReading);
+      }
+      else if (event.type == THERMAL_MGR_INTERRUPT_EVENT){
+        //Imported the logging functions to use this function
+        LOG_IF_ERROR_CODE(readTempLM75BD(LM75BD_OBC_I2C_ADDR, &tempReading));
+        if (tempReading>=80)
+            overTemperatureDetected();
+        else if (tempReading<=75)
+            safeOperatingConditions();
+      }
```
dgobalak: If `readTempLM75BD` fails, you shouldn't use the `tempReading` value since it's invalid.

services/thermal_mgr/thermal_mgr.h
```diff
@@ -5,6 +5,7 @@
 
 typedef enum {
   THERMAL_MGR_EVENT_MEASURE_TEMP_CMD,
+  THERMAL_MGR_INTERRUPT_EVENT,
```
dgobalak: nit: to match the naming convention of this enum, this should be `THERMAL_MGR_EVENT_INTERRUPT`

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +44,42 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+        LOG_IF_ERROR_CODE(readTempLM75BD(LM75BD_OBC_I2C_ADDR, &tempReading));
+        addTemperatureTelemetry(tempReading);
+      }
+      else if (event.type == THERMAL_MGR_INTERRUPT_EVENT){
+        //Imported the logging functions to use this function
+        LOG_IF_ERROR_CODE(readTempLM75BD(LM75BD_OBC_I2C_ADDR, &tempReading));
+        if (tempReading>=80)
+            overTemperatureDetected();
+        else if (tempReading<=75)
+            safeOperatingConditions();
```
dgobalak: Instead of hard-coding 80 and 75, use the LM75BD_DEFAULT_OT_THRESH and LM75BD_DEFAULT_HYST_THRESH macros that have been defined for you.

dgobalak (changes requested):

Navtajh04 (commented):
LGTM, ill add you to the github and then you can send me your email you I can add you to the notion and you can take a look through the FW task board and let me know a couple of tasks that interest you

PR #116, opened 2024-01-02
Firmware Onboarding

HermanG05:
# Purpose
Completed the OBC-Firmware-Onboarding

# New Changes
- Explain new changes

# Testing
Passes all tests

# Outstanding Changes


cmake-build-debug/CMakeFiles/3.25.2/CMakeCCompiler.cmake
```diff
@@ -0,0 +1,72 @@
+set(CMAKE_C_COMPILER "C:/Program Files/JetBrains/CLion 2023.1.4/bin/mingw/bin/gcc.exe")
```
Navtajh04, line 1: integration test output doesnt look right

lm75bd/lm75bd.c
```diff
@@ -9,7 +9,7 @@
 
 /* LM75BD Registers (p.8) */
 #define LM75BD_REG_CONF 0x01U  /* Configuration Register (R/W) */
-
+#define temp_addr 0x00
```
Navtajh04: macro names should be all caps

lm75bd/lm75bd.c
```diff
@@ -26,7 +26,28 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
...
 
 error_code_t readTempLM75BD(uint8_t devAddr, float *temp) {
-  /* Implement this driver function */
+
+  uint8_t temp1 = temp_addr;
+  i2cSendTo(devAddr, &temp1, 1);
+
+  if (temp == NULL) {
+    return ERR_CODE_INVALID_ARG; 
+  }
```
Navtajh04, line 32: this should be checked at the beginning of the function since if the temp is invalid then we shouldn't be setting the pointer register

services/thermal_mgr/thermal_mgr.c
```diff
@@ -44,17 +44,57 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
 
-  return ERR_CODE_SUCCESS;
+  if (xQueueSend(thermalMgrQueueHandle, (void *) &event, 5000/portTICK_PERIOD_MS) == pdPASS)
+  {
+    return ERR_CODE_SUCCESS;
+  }  
+  if (thermalMgrQueueHandle == NULL || event == NULL)
+  {
+    return ERR_CODE_INVALID_ARG; 
+  }
```
Navtajh04, line 50: you should always be checking the validity of the parameters at the beginning of the function sine you should not do anything if they are invalid

services/thermal_mgr/thermal_mgr.c
```diff
@@ -44,17 +44,57 @@ void initThermalSystemManager(lm75bd_config_t *config) {
 error_code_t thermalMgrSendEvent(thermal_mgr_event_t *event) {
   /* Send an event to the thermal manager queue */
 
-  return ERR_CODE_SUCCESS;
+  if (xQueueSend(thermalMgrQueueHandle, (void *) &event, 5000/portTICK_PERIOD_MS) == pdPASS)
```
Navtajh04: since you didnt check if thermalMgrQUeueHandle or event was NULL before this this could cause a segfault

services/thermal_mgr/thermal_mgr.c
```diff
@@ -44,17 +44,57 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+  {
+    return ERR_CODE_SUCCESS;
+  }  
+  if (thermalMgrQueueHandle == NULL || event == NULL)
+  {
+    return ERR_CODE_INVALID_ARG; 
+  }
+
+  return ERR_CODE_UNKNOWN;
```
Navtajh04: only way we would get here is if the args were valid and the queuesend did not pass which means that the queue was full. ERR_CODE_UNKNOWN should be avoided as much as possible since it does not give much info

services/thermal_mgr/thermal_mgr.c
```diff
@@ -44,17 +44,57 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
   /* Implement this function */
+  thermal_mgr_event_t interrupt;
+  interrupt.type = OVER_TEMPERATURE_SHUTDOWN;
+  thermalMgrSendEvent(&interrupt);
 }
 
 static void thermalMgr(void *pvParameters) {
   /* Implement this task */
+  thermal_mgr_event_t event;
+  void *const pvBuffer = &event;
```
Navtajh04: you can just pass &event into xQueueReceive instead of creating this extra variable since this hurts readability in this case

services/thermal_mgr/thermal_mgr.c
```diff
@@ -44,17 +44,57 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+  
+  
   while (1) {
-    
+    if (xQueueReceive(thermalMgrQueueHandle, pvBuffer, 5000/portTICK_PERIOD_MS) == pdPASS)
+    {
+        if (event.type == THERMAL_MGR_EVENT_MEASURE_TEMP_CMD)
+        {
+            float currtemp;
+            readTempLM75BD(data.devAddr,&currtemp);
```
Navtajh04: check return value since if this was an error code the temperature returned would be invalid and should not be logged

services/thermal_mgr/thermal_mgr.c
```diff
@@ -44,17 +44,57 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+            addTemperatureTelemetry(currtemp);
+        }
+        if (event.type == OVER_TEMPERATURE_SHUTDOWN)
+        {
+          float currtemp;
+          readTempLM75BD(data.devAddr, &currtemp);
```
Navtajh04: same as above

services/thermal_mgr/thermal_mgr.c
Navtajh04: since you are using this function in an ISR you should also never be blocking so the block time should be changed

Navtajh04 (changes requested):

PR #122, opened 2024-01-14
Orbital FW Onboarding Task

ThenujaL:
Completed onboarding Tasks.


lm75bd/lm75bd.h
```diff
@@ -5,7 +5,7 @@
 #include <stdint.h>
 
 /* LM75BD I2C Device Address */
-#define LM75BD_OBC_I2C_ADDR /* Define the address here */
+#define LM75BD_OBC_I2C_ADDR 0b1001111
```
dgobalak: We usually prefer hex over binary for readability.
```suggestion
#define LM75BD_OBC_I2C_ADDR 0x4F
```

lm75bd/lm75bd.c
```diff
@@ -26,8 +27,31 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
...
+
+  /* Positive if 1st MSB is 0 */
+  if(buff[0] < 0b10000000){
+    readValue = readValue >> 5;
+    *temp = ((float) readValue) * 0.125;
+  } else {
+    readValue = ~readValue + 1;
+    readValue = readValue >> 5;
+    *temp =  ((float) readValue) * -0.125;
+  }
```
dgobalak: This can be simplified to:
```suggestion
  int16_t regVal = (buff[0] << 8) | buff[1]; 
  *temp = (float)(regVal >> 5) * 0.125f;
```

This is because the register data is already in 2s complement format.

lm75bd/lm75bd.c
```diff
@@ -26,8 +27,31 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
...
+
+  /* Selecting temperature register */
+  uint8_t sendData = 0;
+  errCode = i2cSendTo(devAddr, &sendData, 1);
+  if (errCode != ERR_CODE_SUCCESS) return errCode;
+
+  /* Receive temp reading */
+  uint8_t buff[2] = {0, 0};
+  errCode = i2cReceiveFrom(devAddr, buff, 2);
+  if (errCode != ERR_CODE_SUCCESS) return errCode;
```
dgobalak: RETURN_IF_ERROR_CODE can be used to log and return the error code.
```suggestion
  /* Selecting temperature register */
  uint8_t sendData = 0;
  RETURN_IF_ERROR_CODE(i2cSendTo(devAddr, &sendData, 1));

  /* Receive temp reading */
  uint8_t buff[2] = {0, 0};
  RETURN_IF_ERROR_CODE(i2cReceiveFrom(devAddr, buff, 2));
```

lm75bd/lm75bd.c
```diff
@@ -26,8 +27,31 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
 }
 
 error_code_t readTempLM75BD(uint8_t devAddr, float *temp) {
-  /* Implement this driver function */
-  
+  error_code_t errCode;
+
+  /* Selecting temperature register */
+  uint8_t sendData = 0;
```
dgobalak: Define the temperature register (`LM75BD_REG_TEMP`) under LM75BD_REG_CONF at the top of this file. Use that instead of hard-coding 0 here.

lm75bd/lm75bd.c
```diff
@@ -5,6 +5,7 @@
 
 #include <stdint.h>
 #include <string.h>
+#include <stdio.h>
```
dgobalak: I don't think this include is necessary?

lm75bd/lm75bd.c
```diff
@@ -26,8 +27,31 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
...
-  
+  error_code_t errCode;
+
+  /* Selecting temperature register */
+  uint8_t sendData = 0;
+  errCode = i2cSendTo(devAddr, &sendData, 1);
+  if (errCode != ERR_CODE_SUCCESS) return errCode;
+
+  /* Receive temp reading */
+  uint8_t buff[2] = {0, 0};
```
dgobalak: It's easier to initialize a buffer to 0 by doing this.
```suggestion
  uint8_t buff[2] = {0};
```

If a buffer is partially initialized, the rest of the elements will be initialized to 0.

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +43,38 @@ void initThermalSystemManager(lm75bd_config_t *config) {
 
 error_code_t thermalMgrSendEvent(thermal_mgr_event_t *event) {
```
dgobalak, line 45: Check if `event == NULL` before using it. If it's null, return ERR_CODE_INVALID_ARG

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +43,38 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
   /* Send an event to the thermal manager queue */
-
+  if(xQueueSend(thermalMgrQueueHandle, event, (TickType_t) 10) != pdPASS) return ERR_CODE_INVALID_QUEUE_MSG;
```
dgobalak: Before using `thermalMgrQueueHandle`, check if `thermalMgrQueueHandle == NULL`. If it is, then return ERR_CODE_INVALID_STATE since it'd mean the firmware is in a weird state.

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +43,38 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
 }
 
 static void thermalMgr(void *pvParameters) {
   /* Implement this task */
+  thermal_mgr_event_t rxEvent;
+  float temp;
+
   while (1) {
-    
+    if(xQueueReceive(thermalMgrQueueHandle, &rxEvent, (TickType_t) 10) == pdPASS){
```
dgobalak: This task doesn't do anything unless it receives an event. So just block indefinitely until we receive an event. Otherwise, this task will unblock every 10 ticks (i.e. every 10ms) for no reason which wastes CPU time

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +43,38 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+  } else {
+    safeOperatingConditions();
+  }
+
 }
 
 static void thermalMgr(void *pvParameters) {
   /* Implement this task */
+  thermal_mgr_event_t rxEvent;
+  float temp;
```
dgobalak: Limit scope of these variables. They aren't used outside the while loop, so they don't need to be declared outside of it.

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +43,38 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+
+  /* Get temp */
+  float temp;
+  readTempLM75BD(LM75BD_OBC_I2C_ADDR, &temp);
+
+  if(temp >= LM75BD_DEFAULT_HYST_THRESH){
+    overTemperatureDetected();
+  } else {
+    safeOperatingConditions();
+  }
```
dgobalak: Since `osHandlerLM75BD` is called from an ISR, there are 2 problems with this approach:
- I2C transfers are slow and ISRs must execute quickly (otherwise other interrupts can get delayed). So you can't read the temperature from inside this ISR.
- Internally, readTempLM75BD will attempt to take a mutex with a non-zero block time. This is problematic because you can't block from an ISR.

A potential solution is to send an event to the thermal mgr telling it that this interrupt occurred. And then the thermal mgr will take care of it next time it runs.

dgobalak (changes requested):
Looks good. Just a few things to fix

lm75bd/lm75bd.c
ThenujaL: ah yes, that's a better way of doing that!

lm75bd/lm75bd.c
ThenujaL: yes, good catch

lm75bd/lm75bd.c
ThenujaL: Done, will push change.

services/thermal_mgr/thermal_mgr.c
ThenujaL: Done, will push change.

services/thermal_mgr/thermal_mgr.c
ThenujaL: Done, will push change.

services/thermal_mgr/thermal_mgr.c
ThenujaL: I made this change to address this: set INCLUDE_vTaskSuspend to 1 in freeRTOS.h, then set the tick count in this api call to portMAX_DELAY. Will push the change

services/thermal_mgr/thermal_mgr.c
ThenujaL: Done, will push change.

services/thermal_mgr/thermal_mgr.c
ThenujaL: ah yes, good point. I have added a new event type `THERMAL_MGR_EVENT_OS_HANDLE`, called `thermalMgrSendEvent(&event)` from the  `osHandlerLM75BD ` and exit right away. I modified the `thermalMgr` with some logic to handle the overtemp shutdown. Does this sound right?

freertos/include/FreeRTOS.h
```diff
@@ -125,7 +125,7 @@ extern "C" {
 #endif
 
 #ifndef INCLUDE_vTaskSuspend
-#define INCLUDE_vTaskSuspend 0
+#define INCLUDE_vTaskSuspend 1
```
dgobalak: INCLUDE_vTaskSuspend is defined as 1 in FreeRTOSConfig.h, so this isn't needed

Testing/Temporary/LastTest.log
dgobalak, line 1: nit: delete this file

Testing/Temporary/CTestCostData.txt
dgobalak, line 1: nit: delete this file

lm75bd/lm75bd.c
```diff
@@ -9,6 +9,7 @@
 
 /* LM75BD Registers (p.8) */
 #define LM75BD_REG_CONF 0x01U  /* Configuration Register (R/W) */
+#define LM75BD_REG_TEMP 0 /* Select temp register */
```
dgobalak: nit: It's good to write whether the register is read-only or if it allows writes.
```suggestion
#define LM75BD_REG_TEMP 0x00U /* Temperature Register (R) */
```

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +43,39 @@ void initThermalSystemManager(lm75bd_config_t *config) {
 
 error_code_t thermalMgrSendEvent(thermal_mgr_event_t *event) {
   /* Send an event to the thermal manager queue */
-
+  if(event == NULL) return ERR_CODE_INVALID_ARG;
+  if(thermalMgrQueueHandle == NULL) return ERR_CODE_INVALID_STATE;
+  if(xQueueSend(thermalMgrQueueHandle, event, (TickType_t) 10) != pdPASS) return ERR_CODE_INVALID_QUEUE_MSG;
```
dgobalak: Since thermalMgrSendEvent is called from an ISR (i.e. osHandler), the block time must be 0.

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +43,39 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
   return ERR_CODE_SUCCESS;
 }
 
 void osHandlerLM75BD(void) {
-  /* Implement this function */
+  thermal_mgr_event_t event;
+  event.type = THERMAL_MGR_EVENT_OS_HANDLE;
```
dgobalak: nit: This ensures any other fields of the struct get initialized to 0.
```suggestion
  thermal_mgr_event_t event = {.type = THERMAL_MGR_EVENT_OS_HANDLE};
```

dgobalak (changes requested):
Just a few more things to fix

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +43,38 @@ void initThermalSystemManager(lm75bd_config_t *config) {
 
 error_code_t thermalMgrSendEvent(thermal_mgr_event_t *event) {
   /* Send an event to the thermal manager queue */
-
+  if(event == NULL) return ERR_CODE_INVALID_ARG;
+  if(thermalMgrQueueHandle == NULL) return ERR_CODE_INVALID_STATE;
+  if(xQueueSend(thermalMgrQueueHandle, event, (TickType_t) 0) != pdPASS) return ERR_CODE_INVALID_QUEUE_MSG;
```
Navtajh04: only way this does not return pdPASS is if the queue is full so we should use that error code instead. xQueueSend has no way of knowing if the message you pass in is valid or invalid, so `ERR_CODE_INVALID_QUEUE_MSG` is mainly used when you receive from a queue and see that the message is invalid

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +43,38 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
 static void thermalMgr(void *pvParameters) {
   /* Implement this task */
   while (1) {
-    
+    thermal_mgr_event_t rxEvent;
+    if(xQueueReceive(thermalMgrQueueHandle, &rxEvent, portMAX_DELAY ) == pdPASS){
+      /* Check event type */
+      float temp;
+      if(rxEvent.type == THERMAL_MGR_EVENT_MEASURE_TEMP_CMD){
+        if(readTempLM75BD(LM75BD_OBC_I2C_ADDR, &temp) == ERR_CODE_SUCCESS){
```
Navtajh04: this should be logged if it fails

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +43,38 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+          addTemperatureTelemetry(temp);
+        }
+      } else if (rxEvent.type == THERMAL_MGR_EVENT_OS_HANDLE){
+        if(readTempLM75BD(LM75BD_OBC_I2C_ADDR, &temp) == ERR_CODE_SUCCESS){
```
Navtajh04: this should be logged if it fails

Navtajh04 (changes requested):
Looks good just a couple changes

services/thermal_mgr/thermal_mgr.c
ThenujaL: ah yes good point. I have changed it to return `ERR_CODE_QUEUE_FULL`

services/thermal_mgr/thermal_mgr.c
ThenujaL: Done! Reformatted this function so it does 1 read operation, checks the error/logs it, and if no error handles it accordingly.

services/thermal_mgr/thermal_mgr.c
ThenujaL: Done! Reformatted this function so it does 1 read operation, checks the error/logs it, and if no error handles it accordingly.

Navtajh04:
LGTM

PR #124, opened 2024-01-23
 Onboarding: Darwin 

loqoman:
Completion of bootcamp challenge :)

### DIFF
--- a/lm75bd/lm75bd.c
+++ b/lm75bd/lm75bd.c
@@ -47,7 +47,7 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
 error_code_t readTempLM75BD(uint8_t devAddr, float *temp) {    
     // Check that temp has been initalized
     if ( temp == NULL ) {
-        return ERR_CODE_UNKNOWN
+        return ERR_CODE_UNKNOWN;  
     }
     uint8_t targetRegister = LM75BD_POINTER_TEMP;
     error_code_t errCode;
@@ -61,7 +61,7 @@ error_code_t readTempLM75BD(uint8_t devAddr, float *temp) {
     // Check for out-of-range error  
     if ( (readTempBuf[1] | 0xF8) > 1 ) {  
         // Any of the top 5 bits were set when they shouldn't have been
-        return ERR_CODE_UNKNOWN 
+        return ERR_CODE_UNKNOWN;
     }
     // MSB is top half of readTempBuf  
     uint16_t tempData =  readTempBuf[0] << 3 |      
--- a/lm75bd/lm75bd.c
+++ b/lm75bd/lm75bd.c
@@ -59,7 +59,7 @@ error_code_t readTempLM75BD(uint8_t devAddr, float *temp) {
     uint8_t readTempBuf[2] = {0};   
     RETURN_IF_ERROR_CODE(i2cReceiveFrom(devAddr, readTempBuf, sizeof(readTempBuf)));  
     // Check for out-of-range error  
-    if ( (readTempBuf[1] | 0xF8) > 1 ) {  
+    if ( (readTempBuf[1] & 0xF8) > 0 ) {  
         // Any of the top 5 bits were set when they shouldn't have been
         return ERR_CODE_UNKNOWN;
     }
@@ -68,16 +68,17 @@ error_code_t readTempLM75BD(uint8_t devAddr, float *temp) {
                         (readTempBuf[1] >> 5);    
     
     /*  uint16_t representation
-     *  ________________________________________________________________________________
+     *  readTempBuf[1]                        readTempBuf[0]
+     *  _______________________________________________________________________________
      *  15 | 14 | 13 | 12 | 11 | 10 | 9 | 8 || B7 | B6 | B5 | B4 | B3 | B2 | B1 | B0 ||
-     *  --------------------------------------------------------------------------------
+     *  -------------------------------------------------------------------------------
      *  0    0    0    0    0    D10  D9  D8   D7   D6   D5   D4   D3   D2   D1   D0
      *  --------------------------------------------------------------------------------
      *                           ^ LM75BD_TEMP_RA_MSB_MASK
      *                            |-----------LM75BD_TEMP_RA_10B_MASK--------------|
      */
-    // Two's Complement
-   float result = 0.0;
+    // Two's Complement  
+    float result = 0.0;
     if ((tempData & LM75BD_TEMP_RA_MSB_MASK) > 0) {
         // Result is negative, invert it and mask off top 5 bits    
         result = -1.0 * (~tempData & LM75BD_TEMP_RA_10B_MASK) * LM75BD_TEMP_LSB_TO_C;   // Calculate degrees C
--- a/lm75bd/lm75bd.c
+++ b/lm75bd/lm75bd.c
@@ -59,7 +59,7 @@ error_code_t readTempLM75BD(uint8_t devAddr, float *temp) {
      *  0    0    0    0    0    D10  D9  D8   D7   D6   D5   D4   D3   D2   D1   D0
      *  --------------------------------------------------------------------------------
      *                           ^ LM75BD_TEMP_RA_MSB_MASK
-     *
+     *                            |-----------LM75BD_TEMP_RA_10B_MASK--------------|
      */
     // Two's Complement
     if ((tempData & LM75BD_TEMP_RA_MSB_MASK) > 0) {
--- a/lm75bd/lm75bd.c
+++ b/lm75bd/lm75bd.c
@@ -54,16 +54,15 @@ error_code_t readTempLM75BD(uint8_t devAddr, float *temp) {
     // Perform an I2C read of the device
     uint8_t readTempBuf[2] = {0};   
     RETURN_IF_ERROR_CODE(i2cReceiveFrom(devAddr, readTempBuf, sizeof(readTempBuf)));  
-    // MSB is top half of readTempBuf
+    // Check for out-of-range error  
+    if ( (readTempBuf[1] | 0xF8) > 1 ) {  
+        // Any of the top 5 bits were set when they shouldn't have been
+        return ERR_CODE_UNKNOWN 
+    }
+    // MSB is top half of readTempBuf  
     uint16_t tempData =  readTempBuf[0] << 3 |      
                         (readTempBuf[1] >> 5);    
     
-    // Check for out-of-range error  
-    // TODO: Change to left shift  
-    if (tempData > (pow(2, 11) - 1)) {
-        // Somehow we got erronious I2C data
-        return ERR_CODE_UNKNOWN;
-    }
     /*  uint16_t representation
      *  ________________________________________________________________________________
      *  15 | 14 | 13 | 12 | 11 | 10 | 9 | 8 || B7 | B6 | B5 | B4 | B3 | B2 | B1 | B0 ||
--- a/lm75bd/lm75bd.c
+++ b/lm75bd/lm75bd.c
@@ -25,15 +25,63 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
   return ERR_CODE_SUCCESS;
 }
 
-error_code_t readTempLM75BD(uint8_t devAddr, float *temp) {
-  /* Implement this driver function */
-  
-  return ERR_CODE_SUCCESS;
+/*
+ * @brief Reads temperature from LM75BD
+ *
+ * @param devAddr - Address of the device to communicate to  
+ * @param temp - Float to store temperature in
+ *
+ */
+error_code_t readTempLM75BD(uint8_t devAddr, float *temp) {  
+    float result = 0.0;
+    uint8_t targetRegister = LM75BD_POINTER_TEMP;
+    error_code_t errCode;
+    // Select sensors internal temperature register using pointer register    
+    LOG_IF_ERROR_CODE(i2cSendTo(devAddr, &targetRegister, sizeof(targetRegister)));
+    // XXX: p.8 of datasheet suggests first temperature read is always incorrect
+    
+    // Perform an I2C read of the device
+    uint8_t readTempBuf[2] = {0};   
+    LOG_IF_ERROR_CODE(i2cReceiveFrom(devAddr, readTempBuf, sizeof(readTempBuf)));  
+    // MSB is top half of readTempBuf
+    uint16_t tempData =  readTempBuf[0] << 3 |      // TODO: Double check the byte ordering here 
+                        (readTempBuf[1] >> 5);    
+    
+    // Check for out-of-range error
+    if (tempData > (pow(2, 11) - 1)) {
+        // Somehow we got erronious I2C data
+        return ERR_CODE_UNKNOWN;
+    }
+    /*  uint16_t representation
+     *  ________________________________________________________________________________
+     *  15 | 14 | 13 | 12 | 11 | 10 | 9 | 8 || B7 | B6 | B5 | B4 | B3 | B2 | B1 | B0 ||
+     *  --------------------------------------------------------------------------------
+     *  0    0    0    0    0    D10  D9  D8   D7   D6   D5   D4   D3   D2   D1   D0
+     *  --------------------------------------------------------------------------------
+     *  
+     */
+    // Two's Complement
+    if ((tempData & LM75BD_TEMP_RA_MSB_MASK) > 0) {
+        // Result is negative, invert it and mask off top 5 bits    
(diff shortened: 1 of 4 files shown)


lm75bd/lm75bd.c
```diff
@@ -25,15 +25,64 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
   return ERR_CODE_SUCCESS;
```
Navtajh04, line 37: looks like theres something wrong in the integration test since its not ending

lm75bd/lm75bd.c
```diff
@@ -25,15 +25,64 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
...
-  return ERR_CODE_SUCCESS;
+/*
+ * @brief Reads temperature from LM75BD
+ *
+ * @param devAddr - Address of the device to communicate to  
+ * @param temp - Float to store temperature in
+ *
+ */
+error_code_t readTempLM75BD(uint8_t devAddr, float *temp) {  
+    float result = 0.0;
```
Navtajh04: nit: since this variable isnt used until the end of the function I would declare it down there instead of up here

lm75bd/lm75bd.c
```diff
@@ -25,15 +25,64 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
...
+    uint8_t targetRegister = LM75BD_POINTER_TEMP;
+    error_code_t errCode;
+    // Select sensors internal temperature register using pointer register    
+    LOG_IF_ERROR_CODE(i2cSendTo(devAddr, &targetRegister, sizeof(targetRegister)));
```
Navtajh04: if this fails we should just return since if we have not set the pointer register successfully, we won't be able to read the temperature. Also, since there is no return following this either, there is a chance the function continues and overwrites errCode to success somewhere else in the function and you return success even though the temp is wrong, so `RETURN_IF_ERROR_CODE` would be better

lm75bd/lm75bd.c
```diff
@@ -25,15 +25,64 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
...
+    // XXX: p.8 of datasheet suggests first temperature read is always incorrect
+    
+    // Perform an I2C read of the device
+    uint8_t readTempBuf[2] = {0};   
+    LOG_IF_ERROR_CODE(i2cReceiveFrom(devAddr, readTempBuf, sizeof(readTempBuf)));
```
Navtajh04: same as above

lm75bd/lm75bd.h
```diff
@@ -17,11 +19,19 @@
 #define LM75BD_OS_OP_MODE_COMP 0x00U
 #define LM75BD_OS_OP_MODE_INT 0x01U
 
+/* LM75BD Register Mappings (p.8) */ 
+#define LM75BD_POINTER_TEMP 0x00U   /* Temperature register (RO) */
+#define LM75BD_POINTER_CONF 0x01U   /* Configuration register (R/W) */
```
Navtajh04: since these macros are just used for the driver and don't need to be accessible to other modules, it would be better to put them in the .c file to limit their scope

lm75bd/lm75bd.h
```diff
@@ -17,11 +19,19 @@
...
+
 // Default temperature thresholds
 #define LM75BD_DEFAULT_OT_THRESH 80.0f
 #define LM75BD_DEFAULT_HYST_THRESH 75.0f
 
-/**
+// Used for temperature conversions
+#define LM75BD_TEMP_LSB_TO_C 0.125f  
+#define LM75BD_TEMP_RA_10B_MASK 0x7FF  
+#define LM75BD_TEMP_RA_MSB_MASK 0x400
```
Navtajh04: same as above

services/thermal_mgr/thermal_mgr.c
```diff
@@ -41,21 +43,77 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+ */
 error_code_t thermalMgrSendEvent(thermal_mgr_event_t *event) {
-  /* Send an event to the thermal manager queue */
-
-  return ERR_CODE_SUCCESS;
+    /* Send an event to the thermal manager queue */  
+    if( event == NULL ) {
+        return ERR_CODE_INVALID_ARG;
+    }
+
```
Navtajh04, line 61: I would check to make sure that the queue has already been created (i.e. the queue handle is not null)

services/thermal_mgr/thermal_mgr.c
```diff
@@ -41,21 +43,77 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+    BaseType_t errorCode;
+    errorCode = xQueueSend( thermalMgrQueueHandle, (void *) event, 0 );
+    switch( errorCode ) {
```
Navtajh04: xQueueSend only ever returns either pdTRUE or errQUEUE_FULL, so I would just get rid of the errCode variable and use an if statement for concision

services/thermal_mgr/thermal_mgr.c
```diff
@@ -41,21 +43,77 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
-
-void osHandlerLM75BD(void) {
-  /* Implement this function */
+/* 
+ * @brief Handles overtemperature events from temperature sensor. It is assumed a
+ *          read event will cause the interrupt to be called (I.E cachedTemperatureReading
+ *          is guarenteed to be updated)
+ */
+void osHandlerLM75BD() {     
+    if( cachedTemperatureReading > 80.0  ) {
```
Navtajh04: if we are using an old temperature reading collected routinely by thermal mgr, then we would always read safe operating conditions since the interrupt would happen when we go from safe to unsafe conditions but our last reading would be from safe conditions. It would be better to read the current temp after the interrupt in thermal mgr

Navtajh04 (changes requested):

services/thermal_mgr/thermal_mgr.c
loqoman: Ah, I see. I had assumed that the LM75BD only asserts the interrupt after a peripheral has performed a read. If so, then we can guarnetee that the cached reading is reflective of where we are in the overtemperature cycle. 

I'll implement a solution where the interrupt handeler sets a flag that tells the periodic sampler to set the operating mode after it's made a reading.

loqoman: Mark as resolved if new solution is acceptable.

lm75bd/lm75bd.c
loqoman: Poked around a little bit on this and not sure what's going on here. Any reason you pointed out this line in particular? Is this something on my end or something that's broken in the CI/CD?

Navtajh04: Likely something on your end since this has happened before with other PRs where there were errors

Navtajh04: Nothing in particular about this line lol mb

lm75bd/lm75bd.c
```diff
@@ -25,15 +37,64 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
...
-  return ERR_CODE_SUCCESS;
+/*
+ * @brief Reads temperature from LM75BD
+ *
+ * @param devAddr - Address of the device to communicate to  
+ * @param temp - Float to store temperature in
+ *
+ */
+error_code_t readTempLM75BD(uint8_t devAddr, float *temp) {  
+    uint8_t targetRegister = LM75BD_POINTER_TEMP;
```
Navtajh04, line 52: check to make sure temp is not NULL pointer

services/thermal_mgr/thermal_mgr.c
```diff
@@ -41,21 +43,86 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+ */
+static void thermalMgr(void *pvParameters) {      
+    // XXX: It is assumed we don't have to initialize temperature sensor
+    lm75bd_config_t *tempSensorConfig;  
+    error_code_t errCode;
+    tempSensorConfig = (lm75bd_config_t *) pvParameters;   
+
+    while (1) {  
+        thermal_mgr_event_t *event_recieved;
+        if ( xQueueReceive( thermalMgrQueueHandle, (void *) event_recieved, 0) == pdTRUE ) {
```
Navtajh04: Since you are initializing `event_recieved` as a pointer to a thermal_mgr_event_t struct, there is memory allocated on the stack for the pointer, but it has not been initialized so it is pointing to NULL (forbidden memory). When you pass this into xQueueRecieve it will see that it is pointing to NULL (equivalent to 0) and try writing to memory address 0 which is what is likely causing the integration test to run forever. Initializing it as a regular struct and then using & would solve this issue as there would be memory allocated for where to store the received event. @SaqAsh here is an example of where the timeout bug pops up

services/thermal_mgr/thermal_mgr.c
```diff
@@ -41,21 +43,86 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+
+    while (1) {  
+        thermal_mgr_event_t *event_recieved;
+        if ( xQueueReceive( thermalMgrQueueHandle, (void *) event_recieved, 0) == pdTRUE ) {    
+            // We got an event from the thermalMgr queue, parse it
+            switch( event_recieved->type ){  
+                case THERMAL_MGR_EVENT_MEASURE_TEMP_CMD:    
+                        // Read temperature 
+                        float temperatureResult = 0.0;   
+                        errCode = readTempLM75BD( tempSensorConfig->devAddr, &temperatureResult );
```
Navtajh04: Check this error code before continuing

services/thermal_mgr/thermal_mgr.c
```diff
@@ -41,21 +43,86 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+                        // Only change overtemperature state if we had an interrupt raised recently
+                        if (recomputeOTStateFlag) {  
+
+                            if ( temperatureResult > 80.0 ) { 
+                                overTemperatureDetected();
+                            } else {
+                                safeOperatingConditions();
+                            }  
+                            // We've handled the flag, reset it
+                            recomputeOTStateFlag = 0;
```
Navtajh04: Since there could be an interrupt between the if statement and when this line of code is executed, if we had an OT interrupt and then while in this if condition it returns to safe operating conditions and triggers another interrupt before this line is executed, the ISR will write a value of one (which is already the value of the flag) but this line will reset the flag to 0 anyways after that and we'll be put in a state of OT even though we have returned to safe conditions. For this reason it would be better to use something like a semaphore or queue message sent from the ISR. This approach with a counter instead of a flag to make sure you don't miss any interrupts handling might also work but is more complex solution

services/thermal_mgr/thermal_mgr.c
```diff
@@ -41,21 +43,86 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+ * @param pvParameters - Configuration struct for the LM75BD temp sensor 
+ */
+static void thermalMgr(void *pvParameters) {      
+    // XXX: It is assumed we don't have to initialize temperature sensor
+    lm75bd_config_t *tempSensorConfig;  
+    error_code_t errCode;
+    tempSensorConfig = (lm75bd_config_t *) pvParameters;   
+
+    while (1) {  
+        thermal_mgr_event_t *event_recieved;
```
Navtajh04: Nit: camelCase

Navtajh04 (changes requested):
Looks good just a couple more changes

services/thermal_mgr/thermal_mgr.c
```diff
@@ -41,21 +43,86 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+                            } else {
+                                safeOperatingConditions();
+                            }  
+                            // We've handled the flag, reset it
+                            recomputeOTStateFlag = 0;
+                        }
+
+                        addTemperatureTelemetry( temperatureResult );
+                    break; 
+                deafult:
```
dgobalak:
```suggestion
                default:
```

It's weird that the compiler didn't catch this.

lm75bd/lm75bd.c
```diff
@@ -25,15 +37,64 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
...
+    RETURN_IF_ERROR_CODE(i2cReceiveFrom(devAddr, readTempBuf, sizeof(readTempBuf)));  
+    // MSB is top half of readTempBuf
+    uint16_t tempData =  readTempBuf[0] << 3 |      
+                        (readTempBuf[1] >> 5);    
+    
+    // Check for out-of-range error
+    if (tempData > (pow(2, 11) - 1)) {
+        // Somehow we got erronious I2C data
+        return ERR_CODE_UNKNOWN;
+    }
```
dgobalak: Let's say we did receive bad data like `0xFF 0xFF` (the highest possible value stored in 2 bytes). Then `tempData` would be `0xFF << 3 | 0xFF >> 5` which is `0b11111111000 | 0b111` which is `0x7FF` (i.e. 2^11 - 1). So tempData would never be greater than (2^11)-1 right?

Also, for future reference, it's generally faster to do 2^x using a left-shift instead of the pow function. I.e. `1 << 11` typically involves fewer instructions than `pow(2,11)` (technically depends on the compiler and any compiler optimizations)

lm75bd/lm75bd.c
```diff
@@ -25,15 +37,64 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
...
+        // Result is negative, invert it and mask off top 5 bits    
+        result = -1.0 * (~tempData & LM75BD_TEMP_RA_10B_MASK) * LM75BD_TEMP_LSB_TO_C;   // Calculate degrees C
+        result -= LM75BD_TEMP_LSB_TO_C;     // We lose a MSB due to 2s Complement
+    } else {
+        // Result is positive
+        result = (tempData & LM75BD_TEMP_RA_10B_MASK) * LM75BD_TEMP_LSB_TO_C;   // Convert into float
+
+    }  
+    // Store result;
+    *temp = result;
```
dgobalak: Since the register data is already in 11-bit 2s complement form, the conversion could be written something like:
```
int16_t regVal = readTempBuf[0] << 8 | readTempBuf[1];
```

services/thermal_mgr/thermal_mgr.c
dgobalak: Since this task doesn't do anything except respond to events, it can just block indefinitely on the queue. This would reduce this task's CPU usage

dgobalak (changes requested):
Just a few comments

lm75bd/lm75bd.c
loqoman: Good catch, this check should be directly after the i2c read, and only one the first 5 bits of the most significant byte.

lm75bd/lm75bd.c
```diff
@@ -25,15 +37,69 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
...
+ * @brief Reads temperature from LM75BD
+ *
+ * @param devAddr - Address of the device to communicate to  
+ * @param temp - Float to store temperature in
+ *
+ */
+error_code_t readTempLM75BD(uint8_t devAddr, float *temp) {    
+    // Check that temp has been initalized
+    if ( temp == NULL ) {
+        return ERR_CODE_UNKNOWN;
```
Navtajh04:
```suggestion
        return ERR_CODE_INVALID_ARG;  
```

lm75bd/lm75bd.c
```diff
@@ -25,15 +37,69 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
...
+    error_code_t errCode;
+    // Select sensors internal temperature register using pointer register    
+    RETURN_IF_ERROR_CODE(i2cSendTo(devAddr, &targetRegister, sizeof(targetRegister)));
+    // XXX: p.8 of datasheet suggests first temperature read is always incorrect
+    
+    // Perform an I2C read of the device
+    uint8_t readTempBuf[2] = {0};   
+    RETURN_IF_ERROR_CODE(i2cReceiveFrom(devAddr, readTempBuf, sizeof(readTempBuf)));  
+    // Check for out-of-range error  
+    if ( (readTempBuf[1] & 0xF8) > 0 ) {
```
Navtajh04: shouldn't you be checking the lower 5 bits? so `0x1F`?

Navtajh04: since those are the ones expected to be 0

Navtajh04 (changes requested):

lm75bd/lm75bd.c
loqoman: Good catch, twisted myself up between looking at the integration tests, my ascii art, and the datasheet :sweat_smile:

Navtajh04 (commented):
LGTM send me your email and ill add you to the notion

PR #127, opened 2024-02-18
Firmware Onboarding Challenge Pull Request

Cathouwu:
# Purpose
Firmware Onboarding Challenge Pull Request

# New Changes
- Implemented temperature sensor driver function, thermal management task and OS handler

# Testing
-Unit test passed
-Integration test runs


lm75bd/lm75bd.c
```diff
@@ -7,6 +7,10 @@
 #include <string.h>
 #include <math.h>
 
+#define BOOL char
+#define FALSE 0
+#define TRUE 1
```
dgobalak: We usually just include `stdbool.h` from the standard library (instead of defining these ourselves) if we need booleans.

services/thermal_mgr/thermal_mgr.c
```diff
@@ -9,6 +9,10 @@
 
 #include <string.h>
 
+#define BOOL char
+#define FALSE 0
+#define TRUE 1
```
dgobalak: We usually just include `stdbool.h` from the standard library (instead of defining these ourselves) if we need booleans.

lm75bd/lm75bd.c
```diff
@@ -27,7 +31,26 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
 
 error_code_t readTempLM75BD(uint8_t devAddr, float *temp) {
   /* Implement this driver function */
+  uint8_t tempReg[1]= {0};
+  i2cSendTo(LM75BD_OBC_I2C_ADDR, tempReg, 1);
+  uint8_t tempData[2] = {0,0};
```
dgobalak: nit: The short form for initializing an array to 0s is the following. This is particularly useful if you have a larger array.

```suggestion
  uint8_t tempData[2] = {0};
```

In C, if you partially initialize an array (i.e. you specify the value for some of the elements in the array), then the rest of the elements will be given a value of 0.

lm75bd/lm75bd.c
```diff
@@ -27,7 +31,26 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
 
 error_code_t readTempLM75BD(uint8_t devAddr, float *temp) {
   /* Implement this driver function */
+  uint8_t tempReg[1]= {0};
```
dgobalak: Instead of hard coding the temperature register address (`0`) here, create a macro for it. For example, at the top of this file, we have a macro for the config register (`LM75BD_REG_CONF`). This just improves readability. Otherwise, it's hard to tell why we're writing 0 here.

lm75bd/lm75bd.c
```diff
@@ -27,7 +31,26 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
...
+      negative = TRUE;
+  }
+  for (uint16_t i=1; i < (1 << 10); i <<= 1){
+      tempDec += !!(tempBin & i)*i;
+  }
+  if (negative){
+      tempDec *= -1;
+  }
+  //convert to Celsius
+  *temp = tempDec * 0.125;
```
dgobalak: This logic is a bit more complicated than it needs to be. The register holds the 11-bit temperature data in 2s complement form already. 

So this conversion can be done like:
```
int16_t regVal = tempData[0] << 8 | tempData[1];
```

services/thermal_mgr/thermal_mgr.c
```diff
@@ -38,23 +42,47 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+  osHandlerEvent.type = OS_HANDLER_EVENT;
+  thermalMgrSendEvent(&osHandlerEvent);
 }
 
 static void thermalMgr(void *pvParameters) {
   /* Implement this task */
+  thermal_mgr_event_type_t queueBuffer;
   while (1) {
-    
+    if( xQueueReceive(thermalMgrQueueHandle, &(queueBuffer),( TickType_t ) 10 ) == pdPASS ){
```
dgobalak: Since this task doesn't do anything unless it receives a queue msg, you can just block indefinitely (instead of using a block time of 10 ticks). This would reduce CPU usage.

services/thermal_mgr/thermal_mgr.c
```diff
@@ -38,23 +42,47 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+  if (event->type == OS_HANDLER_EVENT){
+    xQueueSendToFront(thermalMgrQueueHandle, event, ( TickType_t ) 0);
+  }  
+  else if (event->type == THERMAL_MGR_EVENT_MEASURE_TEMP_CMD){
+    xQueueSend(thermalMgrQueueHandle, event, ( TickType_t ) 10 );
+  }
   return ERR_CODE_SUCCESS;
 }
 
+BOOL readTempState = FALSE;
```
dgobalak: Is this global flag necessary? The task will know that the event came from osHandler if the type was OS_HANDLER_EVENT.

Testing/Temporary/CTestCostData.txt
dgobalak, line 1: This file shouldn't have been committed. I'd just delete the file and then commit the change. Afterwards, if it keeps getting re-generated, you can try adding the filename to the `.gitignore` file.

Testing/Temporary/LastTest.log
dgobalak, line 1: This file shouldn't have been committed. I'd just delete the file and then commit the change. Afterwards, if it keeps getting re-generated, you can try adding the filename to the `.gitignore` file.

services/thermal_mgr/thermal_mgr.c
```diff
@@ -38,23 +42,47 @@ void initThermalSystemManager(lm75bd_config_t *config) {
   thermalMgrQueueHandle = xQueueCreateStatic(
     THERMAL_MGR_QUEUE_LENGTH, THERMAL_MGR_QUEUE_ITEM_SIZE,
     thermalMgrQueueStorageArea, &thermalMgrQueueBuffer);
-
 }
 
 error_code_t thermalMgrSendEvent(thermal_mgr_event_t *event) {
   /* Send an event to the thermal manager queue */
-
```
dgobalak, line 46: I'd add a few checks here just to be safe

```
if (event == NULL) {
  return ERR_CODE_INVALID_ARG;
}

if (thermalMgrQueueHandle == NULL) { // Check if queue has been created
  return ERR_CODE_INVALID_STATE;
}
```

services/thermal_mgr/thermal_mgr.c
```diff
@@ -38,23 +42,47 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+  if (event->type == OS_HANDLER_EVENT){
+    xQueueSendToFront(thermalMgrQueueHandle, event, ( TickType_t ) 0);
+  }  
+  else if (event->type == THERMAL_MGR_EVENT_MEASURE_TEMP_CMD){
+    xQueueSend(thermalMgrQueueHandle, event, ( TickType_t ) 10 );
+  }
```
dgobalak, line 54: If someone adds a new event type in the future, this function won't send the event and will still return success. This would make it an annoying bug to solve. I'd either return an error code if an invalid event type is passed in or have a default block time for all other messages.

services/thermal_mgr/thermal_mgr.h
```diff
@@ -5,7 +5,7 @@
 
 typedef enum {
   THERMAL_MGR_EVENT_MEASURE_TEMP_CMD,
-  
+  OS_HANDLER_EVENT,
```
dgobalak: nit: prefix with `THERMAL_MGR_EVENT_`. When things are defined in header files, we try to be careful with the naming so it doesn't conflict with definitions in other files.

services/thermal_mgr/thermal_mgr.c
```diff
@@ -38,23 +42,47 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+    if( xQueueReceive(thermalMgrQueueHandle, &(queueBuffer),( TickType_t ) 10 ) == pdPASS ){
+      float currentTemp = 0.0;
+      readTempLM75BD(LM75BD_OBC_I2C_ADDR, &currentTemp); 
+      if (readTempState== TRUE){
+        if (currentTemp >= 80.0){
+        overTemperatureDetected();
+      }
+      else if(currentTemp <= 75.0){
+        safeOperatingConditions();
+      }
```
dgobalak, line 81: nit: indentation is a bit off. Not a big deal since we run an auto-formatter on our actual codebase.

```suggestion
        if (currentTemp >= 80.0) {
          overTemperatureDetected();
        }
        else if(currentTemp <= 75.0) {
          safeOperatingConditions();
        }
```

dgobalak (changes requested):
Looks good. Just have a few comments

freertos/include/FreeRTOS.h
```diff
@@ -125,7 +125,7 @@ extern "C" {
 #endif
 
 #ifndef INCLUDE_vTaskSuspend
-#define INCLUDE_vTaskSuspend 0
+#define INCLUDE_vTaskSuspend 1
```
dgobalak: FreeRTOS.h holds the default values for these definitions (that's why there's an `ifndef`) so we usually never touch this file. The application-specific settings are in FreeRTOSConfig.h. There, you'll see that vTaskSuspend is already enabled.

Since FreeRTOS.h includes FreeRTOSConfig.h, if something is defined in the config file, it won't be redefined here due to the ifndef.

lm75bd/lm75bd.c
```diff
@@ -6,9 +6,11 @@
 #include <stdint.h>
 #include <string.h>
 #include <math.h>
+#include <stdbool.h>
 
 /* LM75BD Registers (p.8) */
 #define LM75BD_REG_CONF 0x01U  /* Configuration Register (R/W) */
+#define TEMP_REG_ADDR 0x00U
```
dgobalak: nit: use the same naming style as the config reg macro name
```suggestion
#define LM75BD_REG_TEMP 0x00U /* Temperature Register (R) */
```

dgobalak (commented):
lgtm. Send me your email on Discord and I'll add you to the Notion and GitHub.

I made a few more comments, so take a look at those. No need to fix them. Just keep them in mind when working on future tasks.

PR #128, opened 2024-02-27
Firmware Onboarding Challenge Pull Request

AkshayaT1:
# Purpose

# New Changes
- Completed onboarding tasks

# Testing
- Unit Tests
- Integration Test

# Outstanding Changes

### DIFF
--- a/lm75bd/lm75bd.c
+++ b/lm75bd/lm75bd.c
@@ -9,7 +9,7 @@
 
 /* LM75BD Registers (p.8) */
 #define LM75BD_REG_CONF 0x01U  /* Configuration Register (R/W) */
-#define LM75BD_TEMP_REG 0x00U
+#define LM75BD_REG_TEMP 0x00U
 
 error_code_t lm75bdInit(lm75bd_config_t *config) {
   error_code_t errCode;
--- a/lm75bd/lm75bd.c
+++ b/lm75bd/lm75bd.c
@@ -34,7 +34,7 @@ error_code_t readTempLM75BD(uint8_t devAddr, float *temp) {
 
   error_code_t errCode;
 
-  uint8_t tempRegister = LM75BD_TEMP_REG;
+  uint8_t tempRegister = LM75BD_REG_TEMP;
 
   //read from sensor
   RETURN_IF_ERROR_CODE(i2cSendTo(devAddr, &tempRegister, 1));
--- a/lm75bd/lm75bd.c
+++ b/lm75bd/lm75bd.c
@@ -9,6 +9,7 @@
 
 /* LM75BD Registers (p.8) */
 #define LM75BD_REG_CONF 0x01U  /* Configuration Register (R/W) */
+#define LM75BD_TEMP_REG 0x00U
 
 error_code_t lm75bdInit(lm75bd_config_t *config) {
   error_code_t errCode;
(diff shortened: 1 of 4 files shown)


lm75bd/lm75bd.c
```diff
@@ -9,6 +9,7 @@
 
 /* LM75BD Registers (p.8) */
 #define LM75BD_REG_CONF 0x01U  /* Configuration Register (R/W) */
+#define LM75BD_TEMP_REG 0x00U
```
dgobalak: nit: follow the same naming style as the config reg
```suggestion
#define LM75BD_REG_TEMP 0x00U
```

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +44,55 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+  thermal_mgr_event_t sendEvent;
+  sendEvent.type = THERMAL_MGR_EVENT_INTERRUPT_TRIGGERED;
+  thermalMgrSendEvent(&sendEvent);
 }
 
 static void thermalMgr(void *pvParameters) {
   /* Implement this task */
   while (1) {
-    
+    thermal_mgr_event_t recieved;
```
dgobalak: nit 
```suggestion
    thermal_mgr_event_t received;
```

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +44,55 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+            safeOperatingConditions();
+          }
+        }
+      }
+      else{
+        LOG_ERROR_CODE(errCode);
+      }
+    }
+    else{
+      LOG_ERROR_CODE(ERR_CODE_QUEUE_FULL);
```
dgobalak: A few problems here:
- Since you use portMAX_DELAY as the block time for xQueueReceive, this task will be suspended indefinitely until it receives an event. I.e. if xQueueReceive returns, you can be certain that an event was received. See https://www.freertos.org/a00118.html (we have INCLUDE_vTaskSuspend set to 1). I'm fine with still checking if it returned pdPASS, but this log is unnecessary.
- If xQueueReceive used a finite block time, then it'd only fail if nothing was in the queue. I.e. xQueueReceive would never fail if the queue was full. This is a misleading error code to log.

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +44,55 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+  thermalMgrSendEvent(&sendEvent);
 }
 
 static void thermalMgr(void *pvParameters) {
   /* Implement this task */
   while (1) {
-    
+    thermal_mgr_event_t recieved;
+    float temp;
+    error_code_t errCode = readTempLM75BD(LM75BD_OBC_I2C_ADDR, &temp);
```
dgobalak: This should be done only after receiving an event. Imagine you read the temperature and then wait for an event. You might not receive an event for some time. So you'd be using an old value for telemetry and overtemperature detection.

Also, you should log an error right after it occurs. Right now, you're only logging a readTemp error after an event is received. This is an unnecessary delay and could lead to misleading logs.

dgobalak (changes requested):
Looks good so far. Just have a few comments

dgobalak:
Also could you pull the latest changes from the Orbital `level3` branch to the branch on your forked repo? I just pushed a change that should fix the failing GitHub workflows

dgobalak (commented):
lgtm. Send me your email on Discord and I'll add you to the Notion and GitHub.

PR #131, opened 2024-03-31
Jordyn's Onboarding Challenge

jordynmindorff:
# Purpose
Completed the onboarding! Roast away.

# New Changes
- Implemented required driver and task functions.

# Testing
- Ran provided unit & integration tests, all passed.

# Outstanding Changes
- N/A

### DIFF
--- a/lm75bd/lm75bd.c
+++ b/lm75bd/lm75bd.c
@@ -10,6 +10,9 @@
 /* LM75BD Registers (p.8) */
 #define LM75BD_REG_CONF 0x01U  /* Configuration Register (R/W) */
 
+#define LM75BD_TEMP_MULTIPLIER 0.125
+#define LM75BD_UNUSED_BITS 5
+
 error_code_t lm75bdInit(lm75bd_config_t *config) {
   error_code_t errCode;
 
@@ -41,7 +44,7 @@ error_code_t readTempLM75BD(uint8_t devAddr, float *temp) {
   raw <<= 8; 
   raw |= buff[1];
 
-  *temp = (float)(raw >> 5) * 0.125;
+  *temp = (float)(raw >> LM75BD_UNUSED_BITS) * LM75BD_TEMP_MULTIPLIER;
 
   return ERR_CODE_SUCCESS;
 }
--- a/lm75bd/lm75bd.c
+++ b/lm75bd/lm75bd.c
@@ -26,33 +26,23 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
 }
 
 error_code_t readTempLM75BD(uint8_t devAddr, float *temp) {
-  /* Implement this driver function */
+  if(temp == NULL) {
+    return ERR_CODE_INVALID_ARG;
+  }
+
   error_code_t errCode;
   uint8_t buff[2] = {0x0};
-  uint16_t signMask = 0x0400; // The 11th bit from right (aka bit 10) in our raw int will represent the sign
-  uint16_t elevenBitMask = 0x07ff; // To do bitwise operations to basically treat a uint16_t like it only has 11 bits
 
   RETURN_IF_ERROR_CODE(i2cSendTo(devAddr, buff, 1)); // Set pointer register for temp read
   RETURN_IF_ERROR_CODE(i2cReceiveFrom(devAddr, buff, 2)); // Recieve 2 temperature bytes
 
   // Collect 11 meaningful bits into a 16-bit integer
-  uint16_t raw = buff[0];
-  raw <<= 3; 
-  raw |= buff[1] >> 5;
+  int16_t raw = buff[0];
+  raw <<= 8; 
+  raw |= buff[1];
 
-  // Assign a value to the provided float, dependent on sign
-  if (raw & signMask) {
-    // Two's complement on raw -> positive representation
-    raw = raw ^ elevenBitMask;
-    raw += 1;
+  *temp = (float)(raw >> 5) * 0.125;
 
-    raw &= elevenBitMask; // Ignore any potential overflow to the 12th bit from right
-
-    *temp = -(raw) * 0.125f;
-  } else {
-    *temp = (raw) * 0.125f;
-  }
-  
   return ERR_CODE_SUCCESS;
 }
 
--- a/lm75bd/lm75bd.c
+++ b/lm75bd/lm75bd.c
@@ -27,6 +27,32 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
 
 error_code_t readTempLM75BD(uint8_t devAddr, float *temp) {
   /* Implement this driver function */
+  error_code_t errCode;
+  uint8_t buff[2] = {0x0};
+  uint16_t signMask = 0x0400; // The 11th bit from right (aka bit 10) in our raw int will represent the sign
+  uint16_t flipMask = 0x07ff; // To do a binary NOT on only 11 bits
+  uint16_t overflowMask = ~0xf800; // To get rid of undesired overflow (chop off bits other than the 11 relevant ones)
+
+  RETURN_IF_ERROR_CODE(i2cSendTo(devAddr, buff, 1)); // Set pointer register for temp read
+  RETURN_IF_ERROR_CODE(i2cReceiveFrom(devAddr, buff, 2)); // Recieve 2 temperature bytes
+
+  // Collect 11 meaningful bits into a 16-bit integer
+  uint16_t raw = buff[0];
+  raw = raw << 3; 
+  raw |= buff[1] >> 5;
+
+  // Assign a value to the provided float, dependent on sign
+  if (raw & signMask) {
+    // Two's complement on raw -> positive representation
+    raw = raw ^ flipMask;
+    raw += 1;
+
+    raw &= overflowMask; // Ignore any potential overflow to the 12th bit from right
+
+    *temp = -(raw) * 0.125f;
+  } else {
+    *temp = (raw) * 0.125f;
+  }
   
   return ERR_CODE_SUCCESS;
 }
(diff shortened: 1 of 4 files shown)


lm75bd/lm75bd.c
```diff
@@ -27,6 +27,31 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
 
 error_code_t readTempLM75BD(uint8_t devAddr, float *temp) {
```
Yarik-Popov, line 31: Check that temp is not null

lm75bd/lm75bd.c
```diff
@@ -27,6 +27,31 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
...
+    // Two's complement on raw -> positive representation
+    raw = raw ^ elevenBitMask;
+    raw += 1;
+
+    raw &= elevenBitMask; // Ignore any potential overflow to the 12th bit from right
+
+    *temp = -(raw) * 0.125f;
+  } else {
+    *temp = (raw) * 0.125f;
+  }
```
Yarik-Popov: U don't need this just use int16_t to store the combination of data from the buffer. Then multiply by 0.125 after performing the necessary bit shifting

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +44,42 @@ void initThermalSystemManager(lm75bd_config_t *config) {
 
 error_code_t thermalMgrSendEvent(thermal_mgr_event_t *event) {
```
Yarik-Popov, line 45: Check that the event is not null

Yarik-Popov: Also check that the queue handle is not null, return invalid state error if it is

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +44,42 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
 
 static void thermalMgr(void *pvParameters) {
-  /* Implement this task */
+  thermal_mgr_event_t queueMsg;
+  
   while (1) {
-    
+    if(xQueueReceive(thermalMgrQueueHandle, &queueMsg, 0) == pdPASS) {
+      float temp;
+      readTempLM75BD(LM75BD_OBC_I2C_ADDR, &temp);
```
Yarik-Popov: Reading temperature might fail so check the return. In general always check the return of functions that have our error code as a return type

Yarik-Popov (changes requested):
Overall good, some minor changes otherwise

services/thermal_mgr/thermal_mgr.c
```diff
@@ -42,19 +43,54 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
-  /* Implement this task */
+  thermal_mgr_event_t queueMsg;
+  
   while (1) {
-    
+    if(xQueueReceive(thermalMgrQueueHandle, &queueMsg, 0) == pdPASS) {
+      float temp;
+      
+      error_code_t result = readTempLM75BD(LM75BD_OBC_I2C_ADDR, &temp);
+      result = ERR_CODE_INVALID_STATE;
```
Yarik-Popov: Why are u setting the result to invalid state? Also we often use errCode for the var name of the error code

services/thermal_mgr/thermal_mgr.c
```diff
@@ -42,19 +43,54 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+      
+      if (result != ERR_CODE_SUCCESS) {
+        LOG_ERROR_CODE(result);
```
Yarik-Popov: Add a continue statement after this. The rest of the code in the loop shouldn't get executed

Yarik-Popov (changes requested):

services/thermal_mgr/thermal_mgr.c
jordynmindorff: rip sorry that was me testing locally and forgetting to change it back

services/thermal_mgr/thermal_mgr.c
```diff
@@ -42,19 +43,54 @@ void initThermalSystemManager(lm75bd_config_t *config) {
 }
 
 error_code_t thermalMgrSendEvent(thermal_mgr_event_t *event) {
-  /* Send an event to the thermal manager queue */
+  if (event == NULL) {
+    return ERR_CODE_INVALID_ARG;
+  } else if (thermalMgrQueueHandle == NULL) {
```
PratyushMakkar: Its good practice to make the two null checks separate since there is no dependence between then two.

services/thermal_mgr/thermal_mgr.c
```diff
@@ -42,19 +43,54 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+    return ERR_CODE_INVALID_STATE;
+  }
 
-  return ERR_CODE_SUCCESS;
+  if(xQueueSend(thermalMgrQueueHandle, (void *)event, 0) == pdPASS) {
```
PratyushMakkar: nit: You may want to check for failures instead of success. So for example, doing, 
(xQueueSend(...) != pdPASS) { 
// ... handle failure ...return if needed
}

contintue execution,

services/thermal_mgr/thermal_mgr.c
```diff
@@ -42,19 +43,54 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+  } else {
+    return ERR_CODE_QUEUE_FULL;
+  }
 }
 
+// triggered on an interrupt sent
+// interrupt means either we've exceeded the overtemp threshold or gone back down below it (hysteresis)
 void osHandlerLM75BD(void) {
-  /* Implement this function */
+  thermal_mgr_event_t interruptEvent;
```
PratyushMakkar: initialize the object. if the object contains any pointers, you want to catch those pointer bugs

services/thermal_mgr/thermal_mgr.c
```diff
@@ -42,19 +43,54 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
   while (1) {
-    
+    if(xQueueReceive(thermalMgrQueueHandle, &queueMsg, 0) == pdPASS) {
+      float temp;
+      
+      error_code_t errCode = readTempLM75BD(LM75BD_OBC_I2C_ADDR, &temp);
+      
+      if (errCode != ERR_CODE_SUCCESS) {
+        LOG_ERROR_CODE(errCode);
+        continue;
```
PratyushMakkar: maybe yield here

services/thermal_mgr/thermal_mgr.c
```diff
@@ -42,19 +43,54 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+      }
+
+      if(queueMsg.type == THERMAL_MGR_EVENT_MEASURE_TEMP_CMD) {
```
PratyushMakkar: nit: try using continue instead of deeper nested else ifs

lm75bd/lm75bd.c
```diff
@@ -26,8 +26,23 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
...
+
+  RETURN_IF_ERROR_CODE(i2cSendTo(devAddr, buff, 1)); // Set pointer register for temp read
+  RETURN_IF_ERROR_CODE(i2cReceiveFrom(devAddr, buff, 2)); // Recieve 2 temperature bytes
+
+  // Collect 11 meaningful bits into a 16-bit integer
+  int16_t raw = buff[0];
+  raw <<= 8; 
+  raw |= buff[1];
+
+  *temp = (float)(raw >> 5) * 0.125;
```
PratyushMakkar: 0.125 and 5 should be macro constants

services/thermal_mgr/thermal_mgr.h
```diff
@@ -5,25 +5,50 @@
...
-  
+  THERMAL_MGR_EVENT_HANDLE_INTERRUPT
 } thermal_mgr_event_type_t;
 
 typedef struct {
   thermal_mgr_event_type_t type;
 } thermal_mgr_event_t;
 
+#define OS_TH_TEMP 80
+#define OS_HYS_TEMP 75
```
Yarik-Popov: These are already defined inside of `lm75bd/lm75bd.h` so use those

Yarik-Popov (changes requested):

PratyushMakkar (changes requested):
only small suggestions

services/thermal_mgr/thermal_mgr.c
jordynmindorff: confused what you mean here. Do you mean a switch statement?

PratyushMakkar: So you could do,

if (.....) {
 // handle 
continue;
}

if (.....) {
// handle second condition 
continue;
}
and then for else block you dont need a if statement

lm75bd/lm75bd.h
```diff
@@ -5,7 +5,7 @@
 #include <stdint.h>
 
 /* LM75BD I2C Device Address */
-#define LM75BD_OBC_I2C_ADDR /* Define the address here */
+#define LM75BD_OBC_I2C_ADDR 0x4f  /* Define the address here */
```
PratyushMakkar: I got confused by this small f for a second because I have seen it used for floats, extreme nit: but use F and then add a U for clarity I guess

PratyushMakkar (approved):
looks gtm for all intensive purposes

lm75bd/lm75bd.h
jordynmindorff: Fair

Yarik-Popov:
Lgtm, dm me your email and I'll add u to the notion. Today's our last meeting for the term

PR #134, opened 2024-05-01
Mihir's OBC Firmware Bootcamp

mgupta27:
# Purpose
- My submission for Orbital's firmware bootcamp

# New Changes
- Completed task 1
- Completed task 2
- Completed task 3

# Testing
- Ran unit tests
- Verified integration test works as expected

# Outstanding Changes
- N/A

### DIFF
--- a/lm75bd/lm75bd.c
+++ b/lm75bd/lm75bd.c
@@ -50,36 +50,22 @@ error_code_t readTempLM75BD(uint8_t devAddr, float *temp) {
   // 1: Least SignificantByte (LSB)
   uint8_t readBuff[TEMP_READ_BUFF_SIZE] = {0};
   
-  // Stores the data read from the temperature register 
-  uint16_t tempRegValue = 0x00U;
-
-  // Stores if the temperature is positive or negative
-  uint8_t isTempPositive;
-
   RETURN_IF_ERROR_CODE(i2cSendTo(LM75BD_OBC_I2C_ADDR, &writeBuff, TEMP_WRITE_BUFF_SIZE));
 
   RETURN_IF_ERROR_CODE(i2cReceiveFrom(LM75BD_OBC_I2C_ADDR, readBuff, TEMP_READ_BUFF_SIZE));
+
+  // Stores the data read from the temperature register 
+  int16_t tempRegValue = 0;
 
   // Assemble 16-bit value
   tempRegValue |= (readBuff[0] << 8);
   tempRegValue |= readBuff[1];
 
-  // Poll D10 (15th bit) to determine if temperature value is positive or negative
-  // D10 = 0: Temperature is positive
-  // D10 = 1: Temperature is negative
-  isTempPositive = !((tempRegValue >> 15) & 0x01);
-
   // Store the 11-bit value by removing the 5 unused bits
   tempRegValue = tempRegValue >> 5;
 
   // Calculate and store temperature value
-  if (isTempPositive) {
-    *temp = tempRegValue * TEMP_CONVERSION_VAL;
-  } else {
-    tempRegValue = (~tempRegValue) + 0x01;  // Perform 2's complement
-    tempRegValue &= ELEVEN_BIT_MASK;        // Take the first 11 bits starting from the lsb
-    *temp = -1 * tempRegValue * TEMP_CONVERSION_VAL;
-  }
+  *temp = (float) tempRegValue * TEMP_CONVERSION_VAL;
 
   return ERR_CODE_SUCCESS;
 }
--- a/lm75bd/lm75bd.c
+++ b/lm75bd/lm75bd.c
@@ -15,6 +15,11 @@
 #define TEMP_CONVERSION_VAL 0.125f  /* Digital to Analog conversion rate */
 #define ELEVEN_BIT_MASK 0x07FFU     /* Mask to take first 11 bits from lsb side -> 0000 0111 1111 1111 */
 
+/* Macros for readTempLM75BD function */
+#define TEMP_WRITE_BUFF_SIZE 1U     /* Size of buffer for writing temperature data request */
+#define TEMP_READ_BUFF_SIZE 2U      /* Size of buffer for reading temperature data */
+#define TEMP_CONVERSION_VAL 0.125f  /* Conversion value from raw temperature data to temperature in celsisus */
+
 error_code_t lm75bdInit(lm75bd_config_t *config) {
   error_code_t errCode;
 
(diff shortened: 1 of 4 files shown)


services/thermal_mgr/thermal_mgr.c
```diff
@@ -38,23 +38,53 @@ void initThermalSystemManager(lm75bd_config_t *config) {
   thermalMgrQueueHandle = xQueueCreateStatic(
     THERMAL_MGR_QUEUE_LENGTH, THERMAL_MGR_QUEUE_ITEM_SIZE,
     thermalMgrQueueStorageArea, &thermalMgrQueueBuffer);
-
 }
 
 error_code_t thermalMgrSendEvent(thermal_mgr_event_t *event) {
```
Yarik-Popov, line 44: Check that the event is not null and that the queue is valid returning invalid args and invalid state respectively. When using ptrs always check that they aren't null

services/thermal_mgr/thermal_mgr.c
```diff
@@ -38,23 +38,53 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+  float temperature;
+
   while (1) {
-    
+    if (xQueueReceive(thermalMgrQueueHandle, &event, 0) != pdTRUE) {
+      continue;
+    }
+      
+    if (event.type == THERMAL_MGR_EVENT_MEASURE_TEMP_CMD) { 
+      errCode = readTempLM75BD(LM75BD_OBC_I2C_ADDR, &temperature);
```
Yarik-Popov: Log the error code if it isn't success

services/thermal_mgr/thermal_mgr.c
```diff
@@ -38,23 +38,53 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+      
+      if (errCode == ERR_CODE_SUCCESS) {
+        addTemperatureTelemetry(temperature);
+      }
+    } 
+    else if (event.type == THERMAL_MGR_EVENT_OS_TRIGGERED) {
+      errCode = readTempLM75BD(LM75BD_OBC_I2C_ADDR, &temperature);
```
Yarik-Popov: Same thing here for logging the error code

services/thermal_mgr/thermal_mgr.c
```diff
@@ -38,23 +38,53 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+  thermal_mgr_event_t event;
+  float temperature;
+
   while (1) {
-    
+    if (xQueueReceive(thermalMgrQueueHandle, &event, 0) != pdTRUE) {
+      continue;
+    }
+      
+    if (event.type == THERMAL_MGR_EVENT_MEASURE_TEMP_CMD) {
```
Yarik-Popov, line 81: I would add a condition that logs an error if some other event was sent

lm75bd/lm75bd.c
```diff
@@ -1,73 +1,122 @@
...
+  // Assume that the overtemperature and hysteresis thresholds are already set
+  // Hysteresis: 75 degrees Celsius
+  // Overtemperature: 80 degrees Celsius
+
+  return ERR_CODE_SUCCESS;
+}
+
+#define TEMP_WRITE_BUFF_SIZE 1U
+#define TEMP_READ_BUFF_SIZE 2U
+#define TEMP_CONVERSION_VAL 0.125f
```
Yarik-Popov: U also defined macros above so i would put them all in one place so move these macros to the other ones u defined

lm75bd/lm75bd.c
```diff
@@ -1,73 +1,122 @@
...
+  tempRegValue = tempRegValue >> 5;
+
+  // Calculate and store temperature value
+  if (isTempPositive) {
+    *temp = tempRegValue * TEMP_CONVERSION_VAL;
+  } else {
+    tempRegValue = (~tempRegValue) + 0x01;  // Perform 2's complement
+    tempRegValue &= ELEVEN_BIT_MASK;        // Take the first 11 bits starting from the lsb
+    *temp = -1 * tempRegValue * TEMP_CONVERSION_VAL;
+  }
```
Yarik-Popov: Most of this code isnt needed. Just use int16_t, left shift it then multiply

lm75bd/lm75bd.c
```diff
@@ -1,73 +1,122 @@
...
+  // Hysteresis: 75 degrees Celsius
+  // Overtemperature: 80 degrees Celsius
+
+  return ERR_CODE_SUCCESS;
+}
+
+#define TEMP_WRITE_BUFF_SIZE 1U
+#define TEMP_READ_BUFF_SIZE 2U
+#define TEMP_CONVERSION_VAL 0.125f
+error_code_t readTempLM75BD(uint8_t devAddr, float *temp) {
```
Yarik-Popov, line 38: Check that the temp is a valid ptr before using it

lm75bd/lm75bd.c
```diff
@@ -1,73 +1,122 @@
...
+  uint16_t tempRegValue = 0x00U;
+
+  // Stores if the temperature is positive or negative
+  uint8_t isTempPositive;
+
+  errCode = i2cSendTo(LM75BD_OBC_I2C_ADDR, &writeBuff, TEMP_WRITE_BUFF_SIZE);
+  if (errCode != ERR_CODE_SUCCESS) return errCode;
+
+  errCode = i2cReceiveFrom(LM75BD_OBC_I2C_ADDR, readBuff, TEMP_READ_BUFF_SIZE);
+  if (errCode != ERR_CODE_SUCCESS) return errCode;
```
Yarik-Popov: I would use the `RETURN_IF_ERROR_CODE(errCode)` macro instead of the way u did it here

Yarik-Popov (changes requested):
Overall good, some changes are needed

lm75bd/lm75bd.c
mgupta27: Just for my own understanding how were you able to deduce this from the datasheet? Do I need to change the logic for the code here or is it ok to keep. The `isTempPositive` conditional statement was taken directly from the directions in the datasheet.

Yarik-Popov: The datasheet describes the way to implement operations using 2s complement which is the default way that signed ints in C work thus this code is unnecessarily complex. When I was doing my onboarding I also wrote something similar to what u did and was told to change it. It can be simplified to something like this `*temp = (float) tempRegValue * TEMP_CONVERSION_VAL;` if tempRegValue is a signed int (int16_t) instead

lm75bd/lm75bd.c
```diff
@@ -1,73 +1,122 @@
...
+
+  // Assemble 16-bit value
+  tempRegValue |= (readBuff[0] << 8);
+  tempRegValue |= readBuff[1];
```
Yarik-Popov, line 62: Better to declare this variable here as it isn't used above, u want to reduce the scope of variables as much as possible. Also this can be made into a single line

services/thermal_mgr/thermal_mgr.c
Yarik-Popov: You didnt add the check for the queue being initialized

Yarik-Popov (changes requested):

lm75bd/lm75bd.c
mgupta27: Makes sense.

Yarik-Popov:
lgtm, dm me your email and ill add u to the notion

PR #150, opened 2024-09-01
Orbital - Firmware Onboarding Challenge

AarnavNoble:
# Purpose
Complete the firmware onboarding challenge to join Orbital's Firmware subteam.

# New Changes
- Made changes to pass the unit and integration tests detailed in the onboarding document.

# Testing
- Passed both the unit and integration tests.


lm75bd/lm75bd.c
```diff
@@ -26,8 +26,21 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
...
 
 error_code_t readTempLM75BD(uint8_t devAddr, float *temp) {
-  /* Implement this driver function */
+  if (temp == NULL) {
+    return ERR_CODE_INVALID_ARG;
+  }
+
+  error_code_t errCode;
   
+  uint8_t buffer[2] = {0x0};
```
Navtajh04: Nit: would be more readable to just write 0 here instead of 0x0

lm75bd/lm75bd.c
```diff
@@ -26,8 +26,21 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
...
+
+	RETURN_IF_ERROR_CODE(i2cSendTo(devAddr, buffer, 1));
+  RETURN_IF_ERROR_CODE(i2cReceiveFrom(devAddr, buffer, 2));
```
Navtajh04: Better to use sizeof(buffer) instead of hardcoding 2 in case the size of buffer is ever changed so we don't end up with an out of bounds array access

services/thermal_mgr/thermal_mgr.c
```diff
@@ -42,19 +43,56 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
 }
 
 static void thermalMgr(void *pvParameters) {
-  /* Implement this task */
+  thermal_mgr_event_t event;
+  float temp;
+
   while (1) {
-    
+    if (xQueueReceive(thermalMgrQueueHandle, &event, 0) == pdPASS){
```
Navtajh04: I would set the block time to something larger than 0 since we don't have any reason to be polling the queue, it's better to stay blocked until there's something actually available in the queue

services/thermal_mgr/thermal_mgr.c
```diff
@@ -42,19 +43,56 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+      error_code_t errCode = readTempLM75BD(LM75BD_OBC_I2C_ADDR, &temp);
```
Navtajh04, line 74: We shouldn't be reading the temp until we know what the event received from the queue is. Ex: if the msg received is invalid we shouldn't have read the temp

Navtajh04 (changes requested):
Looks pretty good overall, just a couple of changes

services/thermal_mgr/thermal_mgr.c
```diff
@@ -42,19 +43,65 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+  thermal_mgr_event_t event;
+  float temp;
+
   while (1) {
-    
+    if (xQueueReceive(thermalMgrQueueHandle, &event, portMAX_DELAY) == pdPASS){
+      error_code_t errCode = readTempLM75BD(LM75BD_OBC_I2C_ADDR, &temp);
+
+      if (event.type == THERMAL_MGR_EVENT_MEASURE_TEMP_CMD) {
+        error_code_t errCode = readTempLM75BD(LM75BD_OBC_I2C_ADDR, &temp);
```
Navtajh04, line 77: nit: it's simpler to use `LOG_IF_ERROR_CODE` so that the following if statement can be removed since it will set `errCode` to the return val and also log it if it is an error code, but it's fine to not make this change since its minor

Navtajh04 (approved):
LGTM, DM one of the firmware leads with your github and email so that you can be added to the notion and github and ping us with a couple of tasks from the FW task board that you are interested in

PR #151, opened 2024-09-02
abdel/onboarding

Abdel-Salam-A:
# Purpose
Onboarding

# New Changes
- Explain new changes

# Testing
Unit tests, and integration test

# Outstanding Changes


lm75bd/lm75bd.c
```diff
@@ -27,7 +27,23 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
...
+  i2cSendTo(devAddr, &addr, 1);
+
+  i2cReceiveFrom(devAddr, buf, 2);
+
+  uint8_t b1 = buf[0];
+  uint8_t b2 = buf[1];
+
+  uint16_t result = ((b1 << 8) | b2) >> 5 ;
   
+    if (result & 0x0400) {
```
Navtajh04: You shouldn't need this if statement

lm75bd/lm75bd.c
```diff
@@ -27,7 +27,23 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
 
 error_code_t readTempLM75BD(uint8_t devAddr, float *temp) {
   /* Implement this driver function */
+  uint8_t buf[2];
+
+  uint8_t addr = 0x0U;
+  i2cSendTo(devAddr, &addr, 1);
+
+  i2cReceiveFrom(devAddr, buf, 2);
```
Navtajh04: I would use sizeof(buf) instead of hardcoding 2 so that if buf size ever changes this doesn't need to be changed as well

lm75bd/lm75bd.h
```diff
@@ -5,7 +5,7 @@
 #include <stdint.h>
 
 /* LM75BD I2C Device Address */
-#define LM75BD_OBC_I2C_ADDR /* Define the address here */
+#define LM75BD_OBC_I2C_ADDR 0b1001111  /* Define the address here */
```
Navtajh04: Nit: it's generally more readable to use hex instead of binary

services/thermal_mgr/thermal_mgr.c
```diff
@@ -42,19 +42,28 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
-  /* Send an event to the thermal manager queue */
+  xQueueSend(thermalMgrQueueHandle, event, portMAX_DELAY);
 
   return ERR_CODE_SUCCESS;
 }
 
 void osHandlerLM75BD(void) {
-  /* Implement this function */
+  float temp;
+  readTempLM75BD(LM75BD_OBC_I2C_ADDR, &temp);
```
Navtajh04: You should avoid reading from the temp sensor in an ISR for 2 reasons

1) usually I2c and other serial communication functions will be protected with a mutex and we should not be trying to take a mutex in an ISR since it is running outside of a task

2) I2C and consequently this function call are relatively slow and the ISR should be kept as short and fast as possible

It would be better to notify the thermal mgr task somehow that this interrupt has happened so that it can deal with it outside of an ISR context

services/thermal_mgr/thermal_mgr.c
```diff
@@ -42,19 +42,28 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+  if (temp > 75) overTemperatureDetected();
+  else safeOperatingConditions();
 }
 
 static void thermalMgr(void *pvParameters) {
   /* Implement this task */
+  lm75bd_config_t * data = (void *) pvParameters;
```
Navtajh04: This variable is unused so you can probably remove this line

Navtajh04 (changes requested):
Looks pretty good, requested a couple of changes

lm75bd/lm75bd.c
Navtajh04: I don't think this was addressed

services/thermal_mgr/thermal_mgr.c
```diff
@@ -42,19 +42,32 @@ void initThermalSystemManager(lm75bd_config_t *config) {
 }
 
 error_code_t thermalMgrSendEvent(thermal_mgr_event_t *event) {
-  /* Send an event to the thermal manager queue */
```
Navtajh04, line 45: check if `event` is NULL

Navtajh04: also check to make sure the queue has been created

services/thermal_mgr/thermal_mgr.c
```diff
@@ -42,19 +42,32 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+  xQueueSend(thermalMgrQueueHandle, event, portMAX_DELAY);
```
Navtajh04: since you are using this function in an ISR, block time should be `0`

Navtajh04: In general, always check the return value of non void functions for error codes

services/thermal_mgr/thermal_mgr.c
```diff
@@ -42,19 +42,32 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
 
   return ERR_CODE_SUCCESS;
 }
 
 void osHandlerLM75BD(void) {
-  /* Implement this function */
+  thermal_mgr_event_t event; event.type = THERMAL_MGR_EVENT_ISR_OVERTEMP_CHECK;
+  xQueueSend(thermalMgrQueueHandle, &event, portMAX_DELAY);
```
Navtajh04: better to use the `thermalMgrSendEvent` function here

services/thermal_mgr/thermal_mgr.c
```diff
@@ -42,19 +42,32 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
 error_code_t thermalMgrSendEvent(thermal_mgr_event_t *event) {
-  /* Send an event to the thermal manager queue */
+  xQueueSend(thermalMgrQueueHandle, event, portMAX_DELAY);
 
   return ERR_CODE_SUCCESS;
 }
 
 void osHandlerLM75BD(void) {
-  /* Implement this function */
+  thermal_mgr_event_t event; event.type = THERMAL_MGR_EVENT_ISR_OVERTEMP_CHECK;
```
Navtajh04: nit: `thermal_mgr_event_t event = {.type = THERMAL_MGR_EVENT_ISR_OVERTEMP_CHECK};`

services/thermal_mgr/thermal_mgr.c
```diff
@@ -42,19 +42,32 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
 
 static void thermalMgr(void *pvParameters) {
   /* Implement this task */
+  lm75bd_config_t *config = (void *) pvParameters;
+  thermal_mgr_event_t event;
   while (1) {
-    
+    if (xQueueReceive(thermalMgrQueueHandle, &event, portMAX_DELAY) == pdTRUE) {
+      float temp;
+      readTempLM75BD(LM75BD_OBC_I2C_ADDR, &temp);
```
Navtajh04: In general, always check the return value of non void functions for error codes

Navtajh04: Also check the event before reading temp. If the event is not valid (i.e. not check temp or THERMAL_MGR_EVENT_ISR_OVERTEMP_CHECK) we should not do anything

Navtajh04 (changes requested):
Couple more comments

lm75bd/lm75bd.c
Abdel-Salam-A: I'm multiplying in two's complement, so I have to check for the negative.

lm75bd/lm75bd.c
```diff
@@ -27,7 +27,23 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
...
+
+  uint8_t b1 = buf[0];
+  uint8_t b2 = buf[1];
+
+  uint16_t result = ((b1 << 8) | b2) >> 5 ;
   
+     if (result & 0x0400) { 
+        result = ~(result | 0xF800) + 1;
+        *temp = result * -0.125;
+     } else *temp = result * 0.125;
```
Yarik-Popov: This code is unnecessary and can be supplied to using a type of `int16_t` for the result variable instead of `uint16_t` and then multiplying by 0.125

services/thermal_mgr/thermal_mgr.c
```diff
@@ -42,19 +42,38 @@ void initThermalSystemManager(lm75bd_config_t *config) {
 }
 
 error_code_t thermalMgrSendEvent(thermal_mgr_event_t *event) {
-  /* Send an event to the thermal manager queue */
+  if(event == NULL) return ERR_CODE_INVALID_ARG;
+  if(thermalMgrQueueHandle == NULL) return ERR_CODE_INVALID_STATE;
+
+  if(xQueueSend(thermalMgrQueueHandle, event, 0)) return ERR_CODE_QUEUE_FULL;
```
Yarik-Popov: Explicitly check that the value for xQueueSend is pdPASS or some equivalent instead of having it converted to a boolean implicitly

services/thermal_mgr/thermal_mgr.h
```diff
@@ -4,7 +4,7 @@
 #include "errors.h"
 
 typedef enum {
-  THERMAL_MGR_EVENT_MEASURE_TEMP_CMD,
+  THERMAL_MGR_EVENT_MEASURE_TEMP_CMD, THERMAL_MGR_EVENT_ISR_OVERTEMP_CHECK
```
Yarik-Popov, line 7: In general we have enum values each seperate lines

lm75bd/lm75bd.c
```diff
@@ -27,7 +27,23 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
 
 error_code_t readTempLM75BD(uint8_t devAddr, float *temp) {
   /* Implement this driver function */
+  uint8_t buf[2];
```
Yarik-Popov: It is good practice to initialize buffers

lm75bd/lm75bd.c
```diff
@@ -27,7 +27,23 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
...
+
+  uint8_t addr = 0x0U;
+  i2cSendTo(devAddr, &addr, 1);
+
+  i2cReceiveFrom(devAddr, buf, sizeof(buf));
```
Yarik-Popov: Each function should be wrapped inside the `RETURN_IF_ERROR()` macro from the logging.h

.gitignore
```diff
@@ -1,6 +1,9 @@
 # Prerequisites
 *.d
 
+# Shell scripts
+.sh
```
Yarik-Popov: Not a part of the onboarding but this helps with the review. It should be `*.sh` and also add `_deps*` and `Testing/`.

lm75bd/lm75bd.c
```diff
@@ -27,7 +27,23 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
 
 error_code_t readTempLM75BD(uint8_t devAddr, float *temp) {
```
Yarik-Popov, line 28: Check that `temp` is not null

services/thermal_mgr/thermal_mgr.c
```diff
@@ -42,19 +42,38 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
 
 static void thermalMgr(void *pvParameters) {
   /* Implement this task */
+  lm75bd_config_t *config = (void *) pvParameters;
+  thermal_mgr_event_t event;
   while (1) {
-    
+    if (xQueueReceive(thermalMgrQueueHandle, &event, portMAX_DELAY) == pdTRUE) {
+      if (event.type != THERMAL_MGR_EVENT_MEASURE_TEMP_CMD 
+       && event.type != THERMAL_MGR_EVENT_ISR_OVERTEMP_CHECK) continue;
```
Yarik-Popov: Needs to log that the event is invalid

Yarik-Popov (changes requested):
Relatively minor changes

lm75bd/lm75bd.c
```diff
@@ -26,7 +26,19 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
...
+   if(temp == NULL) return ERR_CODE_INVALID_ARG;
+
+  uint8_t buf[2];
+  uint8_t addr = 0x0U;
+
+  i2cSendTo(devAddr, &addr, 1);
+  i2cReceiveFrom(devAddr, buf, sizeof(buf));
+
+  int8_t b1 = buf[0];
+  int8_t b2 = buf[1];
```
Yarik-Popov: Maybe the output is off (sent pic in dm on discord) because you are using an int8_t but the buffer is uint8_t

services/thermal_mgr/thermal_mgr.c
```diff
@@ -42,19 +42,40 @@ void initThermalSystemManager(lm75bd_config_t *config) {
 }
 
 error_code_t thermalMgrSendEvent(thermal_mgr_event_t *event) {
-  /* Send an event to the thermal manager queue */
+  if(event == NULL) return ERR_CODE_INVALID_ARG;
+  if(thermalMgrQueueHandle == NULL) return ERR_CODE_INVALID_STATE;
+
+  if(xQueueSend(thermalMgrQueueHandle, event, 0) == pdTRUE) return ERR_CODE_QUEUE_FULL;
```
Yarik-Popov: Should be `xQueueSend(thermalMgrQueueHandle, event, 0) != pdTRUE` instead. pdTRUE is the successfully sent value for the xQueueSend function

Yarik-Popov (changes requested):
Hopefully these changes will fix the issue

lm75bd/lm75bd.c
```diff
@@ -28,16 +28,16 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
 error_code_t readTempLM75BD(uint8_t devAddr, float *temp) {
    if(temp == NULL) return ERR_CODE_INVALID_ARG;
 
-  uint8_t buf[2];
+  int8_t buf[2];
```
kepler452b123: usually you should initialize the buffers

services/thermal_mgr/thermal_mgr.c
```diff
@@ -45,7 +45,7 @@ error_code_t thermalMgrSendEvent(thermal_mgr_event_t *event) {
   if(event == NULL) return ERR_CODE_INVALID_ARG;
```
kepler452b123, line 46: nit: no need for config variable line 61

kepler452b123, line 46: continue would skip the LOG_ERROR_CODE line 67

services/thermal_mgr/thermal_mgr.c
```diff
@@ -42,19 +42,40 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
   while (1) {
-    
+    if (xQueueReceive(thermalMgrQueueHandle, &event, portMAX_DELAY) == pdTRUE) {
+      if (event.type != THERMAL_MGR_EVENT_MEASURE_TEMP_CMD 
+       && event.type != THERMAL_MGR_EVENT_ISR_OVERTEMP_CHECK){
+        continue; LOG_ERROR_CODE(ERR_CODE_INVALID_STATE);
+      }
+      
+      float temp;
+      readTempLM75BD(LM75BD_OBC_I2C_ADDR, &temp);
```
kepler452b123: check if the readTempLM75BD returns success, log if it does not. ideally by doing LOG_IF_ERROR_CODE, since readTempLM75BD returns a error_code_t. more details about LOG_IF_ERROR_CODE can be found in the onboarding challenge document

kepler452b123: generally, if you are calling a function returning an error code you should always check if its successful, and log if it isn't. this is very important for debugging

kepler452b123 (changes requested):

services/thermal_mgr/thermal_mgr.c
```diff
@@ -42,19 +43,41 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
 
   return ERR_CODE_SUCCESS;
 }
 
 void osHandlerLM75BD(void) {
-  /* Implement this function */
+  error_code_t errCode;
+  thermal_mgr_event_t event = {.type = THERMAL_MGR_EVENT_ISR_OVERTEMP_CHECK};
+  
+  LOG_ERROR_CODE(thermalMgrSendEvent(&event));
```
Yarik-Popov: You shouldn't be logging here as this is an isr and the on-boarding challenge doesn't support it

services/thermal_mgr/thermal_mgr.c
```diff
@@ -42,19 +43,41 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
-    
+    if (xQueueReceive(thermalMgrQueueHandle, &event, portMAX_DELAY) == pdTRUE) {
+      if (event.type != THERMAL_MGR_EVENT_MEASURE_TEMP_CMD 
+       && event.type != THERMAL_MGR_EVENT_ISR_OVERTEMP_CHECK){
+        LOG_ERROR_CODE(ERR_CODE_INVALID_STATE);
+        continue;
+      }
+      
+      float temp;
+      LOG_IF_ERROR_CODE(readTempLM75BD(LM75BD_OBC_I2C_ADDR, &temp));
```
Yarik-Popov, line 73: U shouldn't proceed with the rest of the code if there is an error so add a check k to make sure that it continues if the errCode isn't success

Yarik-Popov (changes requested):
Implement the changes listed below and then u will be done with the on-boarding challenge:
- [ ] Add RETURN_IF_ERROR_CODE to lines 34 & 35 of lm75bd.c
- [x] Remove the logging call in the osHandler (we don't have logging in ISR in the challenge)
- [x] Add a check to if the errCode returned by line 73 of the thermal Mgr if it's not a success, continue

services/thermal_mgr/thermal_mgr.c
```diff
@@ -42,19 +43,44 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
-  /* Implement this task */
+  error_code_t errCode = ERR_CODE_SUCCESS;
+  thermal_mgr_event_t event;
   while (1) {
-    
+    if (xQueueReceive(thermalMgrQueueHandle, &event, portMAX_DELAY) == pdTRUE) {
+      if (event.type != THERMAL_MGR_EVENT_MEASURE_TEMP_CMD 
+       && event.type != THERMAL_MGR_EVENT_ISR_OVERTEMP_CHECK){
+        LOG_ERROR_CODE(ERR_CODE_INVALID_STATE);
+        return;
```
Yarik-Popov: You never want to return from a freeRTOS function, it will probably crash the program, so replace all return statements with continue

services/thermal_mgr/thermal_mgr.c
```diff
@@ -42,19 +43,44 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+      }
+      
+      float temp;
+      LOG_IF_ERROR_CODE(readTempLM75BD(LM75BD_OBC_I2C_ADDR, &temp));
+
+      if (errCode != ERR_CODE_SUCCESS) return;
```
Yarik-Popov: You never want to return from a freeRTOS function, it will probably crash the program, so replace all return statements with continue

lm75bd/lm75bd.c
```diff
@@ -26,7 +26,19 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
...
 
 error_code_t readTempLM75BD(uint8_t devAddr, float *temp) {
-  /* Implement this driver function */
+   if(temp == NULL) return ERR_CODE_INVALID_ARG;
+
+  int8_t buf[2] = {0,0};
+  uint8_t addr = 0x0U;
+
+  i2cSendTo(devAddr, &addr, 1);
+  i2cReceiveFrom(devAddr, buf, sizeof(buf));
```
Yarik-Popov: These 2 functions needs to be surrounded by the RETURN_IF_ERROR_CODE macro

Yarik-Popov (changes requested):
Just so you are aware, as I'm not sure if this is stated in the docs. You never want to return from a freeRTOS task, it might crash the program.

List of things to change:
- [ ] Add to lines 34 and 35 of lm75bd.c `RETURN_IF_ERROR_CODE` macros (requested previously but not handled)
- [ ] Change all `return` to `continue` inside of the while(1) of the thermal manager

PR #157, opened 2024-09-15

justinbax:
# Purpose
Finished implementing the temperature sensor driver and the thermal management task

# New Changes
- Driver function to retrieve temperature (`readTempLM75BD`)
- Thermal management task to call driver function on event
- Over-temperature interrupt handler

# Testing
All unit tests are passing, integration tests give expected behaviour

# Outstanding Changes
None

### DIFF
--- a/lm75bd/lm75bd.c
+++ b/lm75bd/lm75bd.c
@@ -26,7 +26,26 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
 }
 
 error_code_t readTempLM75BD(uint8_t devAddr, float *temp) {
-  /* Implement this driver function */
+  error_code_t errCode;
+
+  if (temp == NULL) return ERR_CODE_INVALID_ARG;
+
+  uint8_t ptrByte = 0x00;
+  RETURN_IF_ERROR_CODE(i2cSendTo(devAddr, &ptrByte, 1));
+
+  uint8_t dataBytes[2] = {0};
+  RETURN_IF_ERROR_CODE(i2cReceiveFrom(devAddr, dataBytes, 2));
+
+  uint16_t tempValue = (dataBytes[0] << 3) | (dataBytes[1] >> 5);
+  
+  if (tempValue & 0x400) {
+    // If MSbit is set, temperature is negative; take two's complement
+    tempValue ^= 0x7FF;
+    tempValue++;
+    *temp = tempValue * -0.125f;
+  } else {
+    *temp = tempValue * 0.125f;
+  }
   
   return ERR_CODE_SUCCESS;
 }
--- a/lm75bd/lm75bd.c
+++ b/lm75bd/lm75bd.c
@@ -30,22 +30,17 @@ error_code_t readTempLM75BD(uint8_t devAddr, float *temp) {
 
   if (temp == NULL) return ERR_CODE_INVALID_ARG;
 
-  uint8_t ptrByte = 0x00;
-  RETURN_IF_ERROR_CODE(i2cSendTo(devAddr, &ptrByte, 1));
+  // Sends pointer to access temperature register
+  uint8_t ptrBytes[1] = {0};
+  RETURN_IF_ERROR_CODE(i2cSendTo(devAddr, ptrBytes, 1));
 
+  // Receive temperature
   uint8_t dataBytes[2] = {0};
   RETURN_IF_ERROR_CODE(i2cReceiveFrom(devAddr, dataBytes, 2));
 
-  uint16_t tempValue = (dataBytes[0] << 3) | (dataBytes[1] >> 5);
-  
-  if (tempValue & 0x400) {
-    // If MSbit is set, temperature is negative; take two's complement
-    tempValue ^= 0x7FF;
-    tempValue++;
-    *temp = tempValue * -0.125f;
-  } else {
-    *temp = tempValue * 0.125f;
-  }
+  // Combines both bytes to get final value
+  int16_t tempValue = (dataBytes[0] << 8) | (dataBytes[1]);
+  *temp = (tempValue >> 5) * 0.125f;
   
   return ERR_CODE_SUCCESS;
 }
--- a/lm75bd/lm75bd.c
+++ b/lm75bd/lm75bd.c
@@ -32,11 +32,11 @@ error_code_t readTempLM75BD(uint8_t devAddr, float *temp) {
 
   // Sends pointer to access temperature register
   uint8_t ptrBytes[1] = {0};
-  RETURN_IF_ERROR_CODE(i2cSendTo(devAddr, ptrBytes, 1));
+  RETURN_IF_ERROR_CODE(i2cSendTo(devAddr, ptrBytes, sizeof(ptrBytes)));
 
   // Receive temperature
   uint8_t dataBytes[2] = {0};
-  RETURN_IF_ERROR_CODE(i2cReceiveFrom(devAddr, dataBytes, 2));
+  RETURN_IF_ERROR_CODE(i2cReceiveFrom(devAddr, dataBytes, sizeof(dataBytes)));
 
   // Combines both bytes to get final value
   int16_t tempValue = (dataBytes[0] << 8) | (dataBytes[1]);
--- a/lm75bd/lm75bd.h
+++ b/lm75bd/lm75bd.h
@@ -5,7 +5,7 @@
 #include <stdint.h>
 
 /* LM75BD I2C Device Address */
-#define LM75BD_OBC_I2C_ADDR /* Define the address here */
+#define LM75BD_OBC_I2C_ADDR 0b1001111
 
 /* LM75BD Configuration Values */
 #define LM75BD_DEV_OP_MODE_NORMAL 0x00U
--- a/lm75bd/lm75bd.h
+++ b/lm75bd/lm75bd.h
@@ -5,7 +5,7 @@
 #include <stdint.h>
 
 /* LM75BD I2C Device Address */
-#define LM75BD_OBC_I2C_ADDR 0b1001111
+#define LM75BD_OBC_I2C_ADDR 0x4F
 
 /* LM75BD Configuration Values */
 #define LM75BD_DEV_OP_MODE_NORMAL 0x00U
--- a/services/thermal_mgr/thermal_mgr.c
+++ b/services/thermal_mgr/thermal_mgr.c
@@ -49,6 +49,10 @@ error_code_t thermalMgrSendEvent(thermal_mgr_event_t *event) {
     return ERR_CODE_INVALID_ARG;
   }
 
+  if (!thermalMgrQueueHandle) {
+    return ERR_CODE_INVALID_STATE;
+  }
+
   if (xQueueSend(thermalMgrQueueHandle, event, 0) != pdTRUE) {
     return ERR_CODE_QUEUE_FULL;
   }
@@ -74,7 +78,7 @@ static void thermalMgr(void *pvParameters) {
 
       LOG_IF_ERROR_CODE(readTempLM75BD(LM75BD_OBC_I2C_ADDR, &temp));
 
-      if (event.type == THERMAL_MGR_EVENT_MEASURE_TEMP_CMD) {
+      if (event.type == THERMAL_MGR_EVENT_MEASURE_TEMP_CMD && errCode) {
         addTemperatureTelemetry(temp);
       } else if (event.type == THERMAL_MGR_EVENT_OS_INTERRUPT) {
         // Determine if it's over-temperature or hysteresis, and act accordingly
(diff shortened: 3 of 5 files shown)


lm75bd/lm75bd.h
```diff
@@ -5,7 +5,7 @@
 #include <stdint.h>
 
 /* LM75BD I2C Device Address */
-#define LM75BD_OBC_I2C_ADDR /* Define the address here */
+#define LM75BD_OBC_I2C_ADDR 0b1001111
```
kepler452b123: our standard is to use hex instead of binary for readability

lm75bd/lm75bd.c
```diff
@@ -26,7 +26,26 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
 }
 
 error_code_t readTempLM75BD(uint8_t devAddr, float *temp) {
-  /* Implement this driver function */
+  error_code_t errCode;
+
+  if (temp == NULL) return ERR_CODE_INVALID_ARG;
+
+  uint8_t ptrByte = 0x00;
```
kepler452b123: I think it's more clear to just make 2 buffers using uint8_t writeByte[1] = {0} (or just one buffer with length 2)

lm75bd/lm75bd.c
```diff
@@ -26,7 +26,26 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
...
+
+  uint8_t dataBytes[2] = {0};
+  RETURN_IF_ERROR_CODE(i2cReceiveFrom(devAddr, dataBytes, 2));
+
+  uint16_t tempValue = (dataBytes[0] << 3) | (dataBytes[1] >> 5);
+  
+  if (tempValue & 0x400) {
+    // If MSbit is set, temperature is negative; take two's complement
+    tempValue ^= 0x7FF;
+    tempValue++;
```
kepler452b123: We don't really ever expect it to be negative so you don't have to take 2s complement here

services/thermal_mgr/thermal_mgr.c
```diff
@@ -44,17 +45,61 @@ void initThermalSystemManager(lm75bd_config_t *config) {
 error_code_t thermalMgrSendEvent(thermal_mgr_event_t *event) {
   /* Send an event to the thermal manager queue */
 
+  if (event == NULL) {
+    return ERR_CODE_INVALID_ARG;
+  }
+
+  if (xQueueSend(thermalMgrQueueHandle, event, 10) != pdTRUE) {
+    return ERR_CODE_QUEUE_FULL;
+  }
```
kepler452b123: you could make it shorter than 10 ticks so it blocks for less time

services/thermal_mgr/thermal_mgr.c
```diff
@@ -44,17 +45,61 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+
   return ERR_CODE_SUCCESS;
 }
 
 void osHandlerLM75BD(void) {
-  /* Implement this function */
+  error_code_t errCode;
+  thermal_mgr_event_t event;
+  event.type = THERMAL_MGR_EVENT_OS_INTERRUPT;
+  LOG_IF_ERROR_CODE(thermalMgrSendEvent(&event));
```
kepler452b123: Good instinct to use LOG_IF_ERROR_CODE, but this is an exception where we don't want to log inside of an ISR.

services/thermal_mgr/thermal_mgr.c
```diff
@@ -44,17 +45,61 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+      // Switch case to easily allow for more events in the future
+      switch (event.type) {
+        case THERMAL_MGR_EVENT_MEASURE_TEMP_CMD:
+        case THERMAL_MGR_EVENT_OS_INTERRUPT:
+          // Both cases are practically the same, except we do more logging at the end if it's OS_INTERRUPT
+          if (pvParameters == NULL) {
+            LOG_ERROR_CODE(ERR_CODE_INVALID_ARG);
+            continue;
+          }
+          lm75bd_config_t config = *(lm75bd_config_t *)(pvParameters);
```
kepler452b123: This variable isn't really necessary for anything

services/thermal_mgr/thermal_mgr.c
```diff
@@ -44,17 +45,61 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+
+          if (event.type == THERMAL_MGR_EVENT_OS_INTERRUPT) {
+            if (temp >= config.overTempThresholdCelsius) {
+              overTemperatureDetected();
+            } else if (temp <= config.hysteresisThresholdCelsius) {
+              safeOperatingConditions();
+            } else {
+              LOG_ERROR_CODE(ERR_CODE_INVALID_STATE);
+            }
+          }
```
kepler452b123: I would try to reduce redundancy in this function, e.g. you are checking if its a THERMAL_MGR_EVENT_OS_INTERRUPT twice. Follow this structure: check if the queue receive passes, if it does then read from LM75BD, if reading from LM75BD passes then check if its just a measure cmd and handle that, then check if its a event handle interrupt and handle that

kepler452b123: Also log that an invalid queue message was received if the events are neither THERMAL_MGR_EVENT_OS_INTERRUPT or THERMAL_MGR_EVENT_MEASURE_TEMP_CMD

services/thermal_mgr/thermal_mgr.c
```diff
@@ -44,17 +45,61 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+          // Both cases are practically the same, except we do more logging at the end if it's OS_INTERRUPT
+          if (pvParameters == NULL) {
+            LOG_ERROR_CODE(ERR_CODE_INVALID_ARG);
+            continue;
+          }
+          lm75bd_config_t config = *(lm75bd_config_t *)(pvParameters);
+
+          float temp = 0.0f;
+          LOG_IF_ERROR_CODE(readTempLM75BD(config.devAddr, &temp));
+          addTemperatureTelemetry(temp);
```
kepler452b123: You don't need to call this if its an OS Interrupt, just call overTemperatureDectected and safeOperatingConditions

services/thermal_mgr/thermal_mgr.c
```diff
@@ -44,17 +45,61 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
 
 static void thermalMgr(void *pvParameters) {
-  /* Implement this task */
   while (1) {
-    
+    error_code_t errCode;
+    thermal_mgr_event_t event;
+    BaseType_t eventReceived = xQueueReceive(thermalMgrQueueHandle, &event, 10);
+
+    if (eventReceived == pdTRUE) {
```
kepler452b123: nitpick: you can just put xQueReceive inside the if and avoid creating a BaseType_t variable

services/thermal_mgr/thermal_mgr.c
```diff
@@ -44,17 +45,61 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+  thermal_mgr_event_t event;
+  event.type = THERMAL_MGR_EVENT_OS_INTERRUPT;
+  LOG_IF_ERROR_CODE(thermalMgrSendEvent(&event));
 }
 
 static void thermalMgr(void *pvParameters) {
-  /* Implement this task */
   while (1) {
-    
+    error_code_t errCode;
```
kepler452b123: You can create the errCode after xQueueReceive passes, since the first use for this is on the readLM75BD driver call

services/thermal_mgr/thermal_mgr.c
```diff
@@ -44,17 +45,61 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+    thermal_mgr_event_t event;
+    BaseType_t eventReceived = xQueueReceive(thermalMgrQueueHandle, &event, 10);
```
kepler452b123: You can drop the waiting time from 10

kepler452b123 (changes requested):
Few comments to improve clarity

services/thermal_mgr/thermal_mgr.c
```diff
@@ -44,17 +45,51 @@ void initThermalSystemManager(lm75bd_config_t *config) {
 error_code_t thermalMgrSendEvent(thermal_mgr_event_t *event) {
   /* Send an event to the thermal manager queue */
 
+  if (event == NULL) {
+    return ERR_CODE_INVALID_ARG;
+  }
+
+  if (xQueueSend(thermalMgrQueueHandle, event, 0) != pdTRUE) {
```
Yarik-Popov, line 56: Add a check that the queue handle isn't null. If it is return invalid state

services/thermal_mgr/thermal_mgr.c
```diff
@@ -44,17 +45,51 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
   while (1) {
-    
+    thermal_mgr_event_t event;
+
+    if (xQueueReceive(thermalMgrQueueHandle, &event, 0) == pdTRUE) {
+      // An event was received
+      error_code_t errCode;
+      float temp = 0.0f;
+
+      LOG_IF_ERROR_CODE(readTempLM75BD(LM75BD_OBC_I2C_ADDR, &temp));
```
Yarik-Popov: Add a check after this for the errCode of success. We only want successful temp reads to be used

services/thermal_mgr/thermal_mgr.c
```diff
@@ -44,17 +45,51 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+        if (temp >= LM75BD_DEFAULT_OT_THRESH) {
+          overTemperatureDetected();
+        } else if (temp <= LM75BD_DEFAULT_HYST_THRESH) {
+          safeOperatingConditions();
+        } else {
+          LOG_ERROR_CODE(ERR_CODE_INVALID_STATE);
+        }
+      } else {
+        // The event type is unknown
+        LOG_ERROR_CODE(ERR_CODE_INVALID_STATE);
```
Yarik-Popov: Probably beat to define a new error code for invalid events such as ERR_CODE_INVALID_EVENT

lm75bd/lm75bd.c
```diff
@@ -26,7 +26,21 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
...
+
+  if (temp == NULL) return ERR_CODE_INVALID_ARG;
+
+  // Sends pointer to access temperature register
+  uint8_t ptrBytes[1] = {0};
+  RETURN_IF_ERROR_CODE(i2cSendTo(devAddr, ptrBytes, 1));
+
+  // Receive temperature
+  uint8_t dataBytes[2] = {0};
+  RETURN_IF_ERROR_CODE(i2cReceiveFrom(devAddr, dataBytes, 2));
```
Yarik-Popov: Use sizeof(dataBytes) instead of 2 here

lm75bd/lm75bd.c
```diff
@@ -26,7 +26,21 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
...
 
 error_code_t readTempLM75BD(uint8_t devAddr, float *temp) {
-  /* Implement this driver function */
+  error_code_t errCode;
+
+  if (temp == NULL) return ERR_CODE_INVALID_ARG;
+
+  // Sends pointer to access temperature register
+  uint8_t ptrBytes[1] = {0};
+  RETURN_IF_ERROR_CODE(i2cSendTo(devAddr, ptrBytes, 1));
```
Yarik-Popov: Use sizeof(ptrBytes) instead of 1

Yarik-Popov (changes requested):
Overall quite good, minor changes otherwise

services/thermal_mgr/thermal_mgr.c
```diff
@@ -44,17 +45,55 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
   while (1) {
-    
+    thermal_mgr_event_t event;
+
+    if (xQueueReceive(thermalMgrQueueHandle, &event, 0) == pdTRUE) {
+      // An event was received
+      error_code_t errCode;
+      float temp = 0.0f;
+
+      LOG_IF_ERROR_CODE(readTempLM75BD(LM75BD_OBC_I2C_ADDR, &temp));
```
Yarik-Popov: You should only be reading the temperature when an event tells you to. So move this into the if and elif blocks

services/thermal_mgr/thermal_mgr.c
```diff
@@ -44,17 +45,55 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+
+      if (event.type == THERMAL_MGR_EVENT_MEASURE_TEMP_CMD && errCode) {
```
Yarik-Popov: Instead, it is better to add an if(errCode != success) continue; statement instead after each read

services/thermal_mgr/thermal_mgr.c
Yarik-Popov: It is more readable to use == NULL instead

Yarik-Popov (changes requested):
Some minor changes left

Yarik-Popov:
Great job. You finished the onboarding. Dm me your email and I'll add you to the notion

PR #159, opened 2024-09-25
Completed the Firmware Onboarding Challenge!

Artyom-G:
# Purpose
Completed the onboarding challenge with no failures after builds

image


lm75bd/lm75bd.c
```diff
@@ -26,8 +26,25 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
...
+
+  // The temp is only written in the first 11 bits
+  int16_t rawTemp = ((bufferFrom[0] << 8) | bufferFrom[1]) >> 5;
+ 
+
+  // Checking Two Complement, the next two lines were ChatGPT
+  // Checks the 10th bit for twos complement
+  if (rawTemp & 0x0400) {  
+    rawTemp |= 0xF800; // Correctly fills the twos complement in 16 bit int
+  }
```
Yarik-Popov, line 45: These lines aren't needed as you are using a int16_t so you can remove them

Yarik-Popov:
Thanks for adding a screenshot of the integration test results

lm75bd/lm75bd.c
```diff
@@ -26,8 +26,25 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
...
 error_code_t readTempLM75BD(uint8_t devAddr, float *temp) {
-  /* Implement this driver function */
-  
+  error_code_t errCode;  
+
+  uint8_t bufferTo[1] = {0x00U}; // Temperature Register
+  RETURN_IF_ERROR_CODE(i2cSendTo(devAddr, bufferTo, 1));
+
+  uint8_t bufferFrom[2] = {0x00U};
+  RETURN_IF_ERROR_CODE(i2cReceiveFrom(devAddr, bufferFrom, 2));
```
Yarik-Popov: Instead of hardcoding 1 and 2 for the i2cSendTo and i2cReceiveFrom its better to use sizeof(ptr) instead even if it likely will not change

services/thermal_mgr/thermal_mgr.c
```diff
@@ -42,19 +42,59 @@ void initThermalSystemManager(lm75bd_config_t *config) {
 }
 
 error_code_t thermalMgrSendEvent(thermal_mgr_event_t *event) {
-  /* Send an event to the thermal manager queue */
-
-  return ERR_CODE_SUCCESS;
+  if (xQueueSend(thermalMgrQueueHandle, event, 0) == pdPASS) {
```
Yarik-Popov: You need to check that the event and thermalMgrQueueHandle aren't null before this line. Also you should be checking against pdTRUE

services/thermal_mgr/thermal_mgr.c
```diff
@@ -42,19 +42,59 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
 void osHandlerLM75BD(void) {
-  /* Implement this function */
+  thermal_mgr_event_t event;
+  // This struct only has one member
+  event.type = THERMAL_MGR_EVENT_OS_INTERRUPT;
+
+  if(xQueueSend(thermalMgrQueueHandle, &event, 0) != pdPASS){
+    printConsole("xQueueSend Failed in osHandlerLM75BD\n");
+  }
+  return;
```
Yarik-Popov: Since this is an ISR, you want to keep it as short as possible. Thus, there is no need to print Console here it there is an error

services/thermal_mgr/thermal_mgr.c
```diff
@@ -42,19 +42,59 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
 }
 
 static void thermalMgr(void *pvParameters) {
-  /* Implement this task */
+  // Config will passed from initThermalSystemManager
+  lm75bd_config_t config = *(lm75bd_config_t *)pvParameters;
```
Yarik-Popov, line 68: Config isn't used so you can remove it

services/thermal_mgr/thermal_mgr.c
```diff
@@ -42,19 +42,59 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+    thermal_mgr_event_t event;
+
+    if(xQueueReceive(thermalMgrQueueHandle, &event, portMAX_DELAY) == pdPASS){
+      // If Event is to Check Temperature
+      if(event.type == THERMAL_MGR_EVENT_MEASURE_TEMP_CMD){
+        float tempC = 0;
+	if(readTempLM75BD(LM75BD_OBC_I2C_ADDR, &tempC) != ERR_CODE_SUCCESS){
+	  printConsole("Temperature Read Failed in thermalMgr under THERMAL_MGR_EVENT_MEASURE_TEMP_CMD\n");
+	  continue;
+	}
```
Yarik-Popov: You should be using LOG_IF_ERROR_CODE and an if(errCode != success) continue; line instead

services/thermal_mgr/thermal_mgr.c
```diff
@@ -42,19 +42,59 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+	addTemperatureTelemetry(tempC);
+      }
+      // If Event is an OS Interupt
+      if(event.type == THERMAL_MGR_EVENT_OS_INTERRUPT){ // This was added to the enum in the thermal_mgr.h file
+        // Implement: if temperature was recorded earlier then check that instead of calling a func
+	float tempC = 0;
+	if(readTempLM75BD(config.devAddr, &tempC)){
+	  printConsole("Temperature Read Failed in thermalMgr under THERMAL_MGR_EVENT_OS_INTERRUPT\n");
+	  continue;
```
Yarik-Popov: Same thing as the above message for LOG_IF_ERROR_CODE

services/thermal_mgr/thermal_mgr.c
```diff
@@ -42,19 +42,59 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+	}
+
+	// Determining which interupt it was 
+	if(tempC < config.hysteresisThresholdCelsius){
+	  safeOperatingConditions();
+	}
+	else{
+	  overTemperatureDetected();
+	}
+      }
```
Yarik-Popov, line 94: You should add an else clause here that checks if an invalid event was received. You will need to add a custom error code for that and log it

Yarik-Popov (changes requested):
Overall good, just some minor changes are needed. 

You had questions about the delay on the queue functions and they lgtm. 0 for the isr send as you want to return immediately if there is a failure. And portMAX_DELAY will suspended the task until it is needed

lm75bd/lm75bd.c
Artyom-G: i think they are required, because im storing a 11 bit int in a int16_t in the following way 00000XXXXXXXXXXX. the code above changes it if it detects a two compliment bit, it gets changed to 11111XXXXXXXXXXX (the compliment in 16 bits).

I also tried running the test case without that code and it fails.

services/thermal_mgr/thermal_mgr.c
Artyom-G: im using config for the devAddr and hysteresisThresholdCelsius (one of the two devAddr was hardcoded by accident)

PR #161, opened 2024-09-25
UW Onboarding Final

Artyom-G:
# Pull Request #2

# Purpose
Completed the onboarding challenge and implemented some suggestions 

image


services/thermal_mgr/thermal_mgr.c
```diff
@@ -42,19 +43,59 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
   return ERR_CODE_SUCCESS;
 }
 
 void osHandlerLM75BD(void) {
-  /* Implement this function */
+  thermal_mgr_event_t event;
+  // This struct only has one member
+  event.type = THERMAL_MGR_EVENT_OS_INTERRUPT;
+
+  xQueueSend(thermalMgrQueueHandle, &event, 0);
```
Yarik-Popov: You should be using thermalMgrSendEvent instead

services/thermal_mgr/thermal_mgr.c
```diff
@@ -42,19 +43,59 @@ void initThermalSystemManager(lm75bd_config_t *config) {
 }
 
 error_code_t thermalMgrSendEvent(thermal_mgr_event_t *event) {
-  /* Send an event to the thermal manager queue */
-
+  
+  if(event == NULL || thermalMgrQueueHandle == NULL){
+    return ERR_CODE_INVALID_QUEUE_MSG;
+  }
```
Yarik-Popov: You should return invalid argument if event == NULL and invalid state if thermal mgr queue is null

lm75bd/lm75bd.c
```diff
@@ -26,8 +26,25 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
...
+
+  // The temp is only written in the first 11 bits
+  int16_t rawTemp = ((bufferFrom[0] << 8) | bufferFrom[1]) >> 5;
+ 
+
+  // Checking Two Complement, the next two lines were ChatGPT
+  // Checks the 10th bit for twos complement
+  if (rawTemp & 0x0400) {  
+    rawTemp |= 0xF800; // Correctly fills the twos complement in 16 bit int
+  }
```
Yarik-Popov, line 45: These lines can be removed

Yarik-Popov (changes requested):
Minor changes left

services/thermal_mgr/thermal_mgr.c
Artyom-G: Fixed this

services/thermal_mgr/thermal_mgr.c
Artyom-G: Fixed this

lm75bd/lm75bd.c
Artyom-G: I tried removing this, it fails. I think it needs to be there since the type is 16-bit but its storing a 11-bit value so the complement is messed up otherwise.

Artyom-G:
this should be the final now

Yarik-Popov (commented):
Lgtm, dm me your email and I'll add you to the notion. 

We have a fw meeting at 8pm today. If you want you can look through the task board for tasks in the backlog, not on hold and not assigned to anyone you are interested in doing. Good first tasks are usually p3

PR #162, opened 2024-09-27
Added missing code to complete Onboarding

dbaitech:
# Purpose
To complete the onboarding challenge to join Orbital's firmware team!

# New Changes
Added code to complete the temperature sensor driver, thermal management task and OS Handler.

# Testing
Ran and passed all provided unit and integration tests.


services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +44,66 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
 }
 
 static void thermalMgr(void *pvParameters) {
   /* Implement this task */
+  thermal_mgr_event_t event;
+  error_code_t errCode;
+
+  if (pvParameters == NULL) {
+    LOG_ERROR_CODE(ERR_CODE_INVALID_ARG);
+    return;
```
Yarik-Popov: You should never return from a freeRTOS task, it can cause the program to crash. If you want to handle this error case replace the return with a while(1); so that it doesn't use this argument in the loop

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +44,66 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+  }
+
   while (1) {
-    
+    if (xQueueReceive(thermalMgrQueueHandle, &event, portMAX_DELAY) == pdTRUE) {
+      if ((event.type == THERMAL_MGR_EVENT_MEASURE_TEMP_CMD) ||
+          (event.type == THERMAL_MGR_EVENT_OS_INTERRUPT)) {
+
+        float temperature;
+        lm75bd_config_t config = *(lm75bd_config_t *) pvParameters;  // obtaining lm75db config
```
Yarik-Popov: Best to access this variable before the start of the loop as this argument will not change during the loop

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +44,66 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+          } else {
+            safeOperatingConditions();
+          }
+        }
+      } else {
+        LOG_ERROR_CODE(ERR_CODE_INVALID_STATE);
+      }
+    } else {
+      LOG_ERROR_CODE(ERR_CODE_INVALID_QUEUE_MSG);
+    }
```
Yarik-Popov: What is the purpose of this here?

lm75bd/lm75bd.c
```diff
@@ -25,9 +26,46 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
...
+
+  error_code_t errCode;
+
+  // Set up the buffer to store the temperature register address
+  // 0: Register address
+  uint8_t writeBuff[TEMP_REG_WRITE_BUFF_SIZE] = { LM75BD_REG_TEMP };
+
+  // Send the pointer to set up reading from the temperature register
+  errCode = i2cSendTo(devAddr, writeBuff, TEMP_REG_WRITE_BUFF_SIZE);
+  RETURN_IF_ERROR_CODE(errCode);
```
Navtajh04: you can just do `RETURN_IF_ERROR_CODE(i2cSendTo(devAddr, writeBuff, TEMP_REG_WRITE_BUFF_SIZE));`

lm75bd/lm75bd.c
```diff
@@ -25,9 +26,46 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
...
+
+  // Set up the buffer to store the temperature data received
+  uint8_t readBuff[TEMP_READ_BUFF_SIZE] = { 0, 0 };
```
Navtajh04: nit: you can just do `= {0}` since by C standard this will set all of the other array elements to 0 as well

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +44,66 @@ void initThermalSystemManager(lm75bd_config_t *config) {
 
 error_code_t thermalMgrSendEvent(thermal_mgr_event_t *event) {
   /* Send an event to the thermal manager queue */
+  if (thermalMgrQueueHandle == NULL) {
+    return ERR_CODE_INVALID_QUEUE_MSG;
```
Navtajh04:
```suggestion
    return ERR_CODE_INVALID_STATE;
```

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +44,66 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
   /* Implement this function */
+  thermal_mgr_event_t event = { .type =  THERMAL_MGR_EVENT_OS_INTERRUPT };
+  error_code_t errCode = thermalMgrSendEvent(&event);
+  LOG_IF_ERROR_CODE(errCode);
 }
 
 static void thermalMgr(void *pvParameters) {
   /* Implement this task */
+  thermal_mgr_event_t event;
+  error_code_t errCode;
```
Navtajh04: nit: in general try to limit the scope of variables to the smallest possible

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +44,66 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+
+        if (event.type == THERMAL_MGR_EVENT_OS_INTERRUPT) {
+          if (temperature > config.hysteresisThresholdCelsius) {
+            overTemperatureDetected();
+          } else {
+            safeOperatingConditions();
+          }
+        }
+      } else {
+        LOG_ERROR_CODE(ERR_CODE_INVALID_STATE);
```
Navtajh04: We only reach here if the message we received was not either of the defined values, so it makes more sense to use `ERR_CODE_INVALID_QUEUE_MSG`

Navtajh04 (changes requested):
Looks really good overall, just a couple of minor changes

lm75bd/lm75bd.c
```diff
@@ -25,9 +26,46 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
...
+  // 1. Extract the temperature data (D10 to D0)
+  // | D10 D9 D8 D7 D6 D5 D4 D3 | D2 D1 D0 X X X X X |
+  int16_t temp_data = (readBuff[0] << 3) & 0x7F8;
+  temp_data |= (readBuff[1] >> 5) & 0x07;
+
+  // 2. Checking if D10 is 0 (positive) or 1 (negative)
+  // and sign extending to allow for two's complement if D10 is 1
+  if (temp_data & (1 << 10)) {
+    temp_data |= ~0x7FF;
+  }
```
Yarik-Popov: This code is needless complicated. You can just use ( (readBuff[0] << 8) | readBuff[1]) >> 5 instead

Yarik-Popov (changes requested):
Overall good start. Some changes are needed

services/thermal_mgr/thermal_mgr.c
dbaitech: I added a log in case there was a problem retrieving the item from the queue. Would it be better to create a new FreeRTOS error that better describes this?

Navtajh04 (commented):
LGTM, send me your email and I'll add you to the notion then you can take a look through the FW task board on Notion and ping me with a couple of tasks that interest you

PR #163, opened 2024-09-30
Tyler Chen Onboarding

zjteachen:


lm75bd/lm75bd.c
```diff
@@ -25,9 +26,22 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
...
 }
 
+#define CONF_WRITE_POINTER_BUFF_SIZE 1U
 error_code_t readTempLM75BD(uint8_t devAddr, float *temp) {
   /* Implement this driver function */
-  
+  error_code_t errCode;
+  uint8_t buff[CONF_WRITE_POINTER_BUFF_SIZE] = {0};
+  errCode = i2cSendTo(devAddr, buff, CONF_WRITE_POINTER_BUFF_SIZE);
+  if (errCode != ERR_CODE_SUCCESS) return errCode;
```
Navtajh04, line 35: You can use `RETURN_IF_ERROR_CODE` to simplify this

lm75bd/lm75bd.c
```diff
@@ -25,9 +26,22 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
...
+
+  uint8_t temp_bytes[2U] = {0}; // first interpret data as unsigned
+  errCode = i2cReceiveFrom(devAddr, temp_bytes, 2);
```
Navtajh04, line 38: use `sizeof(temp_bytes)` rather than hardcoding 2

services/thermal_mgr/thermal_mgr.c
```diff
@@ -38,23 +40,43 @@ void initThermalSystemManager(lm75bd_config_t *config) {
   thermalMgrQueueHandle = xQueueCreateStatic(
     THERMAL_MGR_QUEUE_LENGTH, THERMAL_MGR_QUEUE_ITEM_SIZE,
     thermalMgrQueueStorageArea, &thermalMgrQueueBuffer);
-
 }
 
 error_code_t thermalMgrSendEvent(thermal_mgr_event_t *event) {
   /* Send an event to the thermal manager queue */
-
```
Navtajh04, line 46: check to make sure event is not `NULL` and that the queue has been created already

services/thermal_mgr/thermal_mgr.c
```diff
@@ -38,23 +40,43 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+  xQueueSend(thermalMgrQueueHandle, event, portMAX_DELAY);
```
Navtajh04, line 47: Always check the return value for non-void functions

Navtajh04: Also, since you are calling this function in an ISR, the block time should be 0 since you should never be blocking in an ISR

lm75bd/lm75bd.c
Yarik-Popov, line 35: Better to use RETURN_IF_ERROR_CODE macro instead

lm75bd/lm75bd.c
```diff
@@ -25,9 +26,22 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
...
+  if (errCode != ERR_CODE_SUCCESS) return errCode;
```
Yarik-Popov, line 39: Better to use the RETURN_IF_ERROR_CODE macro instead

lm75bd/lm75bd.c
Yarik-Popov, line 38: Use sizeof(temp_bytes) instead of 2. Also you should use the naming convention of camelCase

services/thermal_mgr/thermal_mgr.c
Yarik-Popov, line 47: Add error handling for this:
- check that the event is not null, return invalid argument otherwise
- check that the thermalMgrQueueHandle is not null, otherwise return invalid state
- check the return value of the xQueueSend

Also this should be 0 and not PORTMAX_DELAY as this will block the task which might not be the intended behavior especially in ISR

services/thermal_mgr/thermal_mgr.c
```diff
@@ -38,23 +40,43 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+  event.type = THERMAL_MGR_EVENT_OS_INTERRUPT;
+  thermalMgrSendEvent(&event);
 }
 
 static void thermalMgr(void *pvParameters) {
   /* Implement this task */
   while (1) {
-    
+    thermal_mgr_event_t event;
+    if (xQueueReceive(thermalMgrQueueHandle, &event, portMAX_DELAY)) {
```
Yarik-Popov, line 62: Explicitly check against pdTRUE

services/thermal_mgr/thermal_mgr.c
```diff
@@ -38,23 +40,43 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+      const lm75bd_config_t data = *(lm75bd_config_t *) pvParameters;
```
Yarik-Popov, line 63: This should be outside of the loop so we don't constantly re create it the data and better to use a pointer instead of a regular struct

services/thermal_mgr/thermal_mgr.c
```diff
@@ -38,23 +40,43 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+      float temperature;
+      readTempLM75BD(data.devAddr, &temperature);
```
Yarik-Popov, line 65: In general, you should be checking the return of non-void functions for error codes. This should be wrapped inside of a LOG_IF_ERROR_CODE and if(errCode!= success) continue;
Also this should be moved to each event that needs it. You don't want to read the temp on all events

lm75bd/lm75bd.c
```diff
@@ -25,9 +26,22 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
...
+
+  int data = temp_bytes[0];
+  data -= 2 * ((1 << 7) & data); // handle signed bit
+  data = (data << 3) | (temp_bytes[1] >> 5); // get LSbyte
```
Yarik-Popov, line 43: This code can be simplified to `int16_t data = ( (tempBytes[0] << 8) | tempBytes[1]) >> 5;`

lm75bd/lm75bd.c
```diff
@@ -25,9 +26,22 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
...
+#define CONF_WRITE_POINTER_BUFF_SIZE 1U
 error_code_t readTempLM75BD(uint8_t devAddr, float *temp) {
   /* Implement this driver function */
-  
+  error_code_t errCode;
+  uint8_t buff[CONF_WRITE_POINTER_BUFF_SIZE] = {0};
+  errCode = i2cSendTo(devAddr, buff, CONF_WRITE_POINTER_BUFF_SIZE);
+  if (errCode != ERR_CODE_SUCCESS) return errCode;
+
+  uint8_t temp_bytes[2U] = {0}; // first interpret data as unsigned
```
Yarik-Popov, line 37: If you are going to define a macro for the length above, probably best to define a macro here as well

Yarik-Popov (changes requested):
Overall good, mainly you're code just needs better error handling

lm75bd/lm75bd.c
```diff
@@ -25,9 +26,22 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
   return ERR_CODE_SUCCESS;
 }
 
+#define CONF_WRITE_POINTER_BUFF_SIZE 1U
 error_code_t readTempLM75BD(uint8_t devAddr, float *temp) {
```
Yarik-Popov, line 30: Check that the temp is not null before use

Yarik-Popov (changes requested):

Navtajh04 (changes requested):

Adityya-K:
Closed due to a lot of time passing.

PR #167, opened 2024-10-10
Diba_Darooneh Onboarding Challenge

Dibadi1384:


sys/errors.h
```diff
@@ -1,4 +1,3 @@
-#pragma once
```
Yarik-Popov, line 1: You remove this line. This must always be included at the top of header files so that there aren't multiple redefinitions

Yarik-Popov (changes requested):
The error was due to you removing #pragma once statements in the sys/errors.h file. Once upon are done, pin @Firmware Lead for a review

PR #181, opened 2024-10-11
Diba_Darooneh_Onboarding

Dibadi1384:


lm75bd/lm75bd.c
```diff
@@ -26,9 +26,29 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
 }
 
 error_code_t readTempLM75BD(uint8_t devAddr, float *temp) {
```
Yarik-Popov, line 28: Check that temp is not null before proceeding

lm75bd/lm75bd.c
```diff
@@ -26,9 +26,29 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
...
+    error_code_t errCode;  
+
+    uint8_t buf[2] = {0};   
+    uint8_t pointerRegister = 0x00;  
+
+    // Send a pointer to the temperature register
+    LOG_IF_ERROR_CODE(i2cSendTo(devAddr, &pointerRegister, 1));
+
+    // Read from the temperature register
+    LOG_IF_ERROR_CODE(i2cReceiveFrom(devAddr, buf, 2));
```
Yarik-Popov: These should be RETURN_IF_ERROR_CODE as it doesn't make sense to continue with either one if there was an error. 

Also replace the hard coded 1 and 2 with the sizeof() the buffers

services/thermal_mgr/thermal_mgr.c
```diff
@@ -42,20 +44,71 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+            if (event.type == THERMAL_MGR_EVENT_MEASURE_TEMP_CMD) {
+                // Variable to store the current temperature
+                float temperature;
+
+                // Read the current temperature from the LM75BD sensor
+                errCode = readTempLM75BD(config->devAddr, &temperature); 
+                LOG_IF_ERROR_CODE(errCode); 
+
+                // Check for read temperature error and handle accordingly
+                RETURN_IF_ERROR_CODE(errCode);
```
Yarik-Popov: Never return from a freeRTOS task, it might cause the program to crash when run on a board. You should do if(errCode != success) continue;

services/thermal_mgr/thermal_mgr.c
```diff
@@ -42,20 +44,71 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+
+                // Check temperature against thresholds
+                if (temperature > 80.0f) {
```
Yarik-Popov: Use macros defined instead of hardcoding the value

services/thermal_mgr/thermal_mgr.c
```diff
@@ -42,20 +44,71 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+                    // Over-temperature condition detected
+                    overTemperatureDetected();
+                } else if (temperature < 75.0f) {
```
Yarik-Popov: Same thing here

services/thermal_mgr/thermal_mgr.c
```diff
@@ -42,20 +44,71 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+    error_code_t errCode; 
+    lm75bd_config_t *config = (lm75bd_config_t *)pvParameters;
+
+    while (1) {
+        thermal_mgr_event_t event;
+
+        // Wait for the next event from the queue
+        if (xQueueReceive(thermalMgrQueueHandle, &event, portMAX_DELAY) == pdPASS) {
+            // Check if the event type is MEASURE_TEMP_CMD
+            if (event.type == THERMAL_MGR_EVENT_MEASURE_TEMP_CMD) {
```
Yarik-Popov: You aren't handling the over temperature event

services/thermal_mgr/thermal_mgr.c
```diff
@@ -17,6 +18,7 @@ static StackType_t thermalMgrTaskStack[THERMAL_MGR_STACK_SIZE];
 
 #define THERMAL_MGR_QUEUE_LENGTH 10U
 #define THERMAL_MGR_QUEUE_ITEM_SIZE sizeof(thermal_mgr_event_t)
+#define THERMAL_MGR_QUEUE_TIMEOUT (pdMS_TO_TICKS(100))  // 100 ms timeout
```
Yarik-Popov: Probably should be 0 instead as we don't want to block in isr

services/thermal_mgr/thermal_mgr.c
```diff
@@ -42,20 +44,71 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+        return ERR_CODE_INVALID_STATE;  
+    }
+
+    // Attempt to send the event to the thermal manager queue
+    if (xQueueSend(thermalMgrQueueHandle, event, THERMAL_MGR_QUEUE_TIMEOUT) == pdPASS) {
+        return ERR_CODE_SUCCESS;  
+
+    // Return error if sending the event fails
+    return ERR_CODE_QUEUE_FULL;  
+    }
```
Yarik-Popov: Based on the curly braces it looks as tho this is also inside the success case for the xQueueSend so move the return below the }

services/thermal_mgr/thermal_mgr.c
```diff
@@ -42,20 +44,71 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+                    // Over-temperature condition detected
+                    overTemperatureDetected();
+                } else if (temperature < 75.0f) {
+                    // Safe operating conditions restored
+                    safeOperatingConditions();
+                }
+
+                // Send the measured temperature as telemetry
+                addTemperatureTelemetry(temperature);
+            }
```
Yarik-Popov: You also need to log that an invalid event was received

sys/errors.h
```diff
@@ -16,6 +16,8 @@ typedef enum {
 
     /* Driver errors */
     ERR_CODE_I2C_TRANSFER_TIMEOUT = 200,
+    ERR_CODE_I2C_WRITE_FAIL = 201,  // Added for I2C write failure
+    ERR_CODE_I2C_READ_FAIL = 202,   // Added for I2C read failure
```
Yarik-Popov: Where are these used? If they are unused, remove them

Yarik-Popov (changes requested):
Overall good, some changes are needed mainly to the error and over temperature event handling

services/thermal_mgr/thermal_mgr.c
```diff
@@ -18,6 +19,12 @@ static StackType_t thermalMgrTaskStack[THERMAL_MGR_STACK_SIZE];
 #define THERMAL_MGR_QUEUE_LENGTH 10U
 #define THERMAL_MGR_QUEUE_ITEM_SIZE sizeof(thermal_mgr_event_t)
 
+///Added
+#define THERMAL_MGR_QUEUE_TIMEOUT (pdMS_TO_TICKS(0))  // 100 ms timeout
+#define OVER_TEMPERATURE_THRESHOLD 80.0f
+#define SAFE_TEMPERATURE_THRESHOLD 75.0f
```
Yarik-Popov: These are defined in the lm75bd.h file so use those instead

services/thermal_mgr/thermal_mgr.c
```diff
@@ -42,20 +49,77 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+                    // Safe operating conditions restored
+                    safeOperatingConditions();
+                }
+                
+                // Send the measured temperature as telemetry
+                addTemperatureTelemetry(temperature);           
+            }
+
+          // Log invalid or unknown event
+          LOG_ERROR_CODE(ERR_CODE_INVALID_STATE);
```
Yarik-Popov: Create a custom error code for an invalid event. This should be in the else case of the event if checks

services/thermal_mgr/thermal_mgr.c
```diff
@@ -42,20 +49,77 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+
+                osHandlerLM75BD();
+
+                // Check temperature against thresholds
+                if (temperature > OVER_TEMPERATURE_THRESHOLD) {
+                    // Over-temperature condition detected
+                    overTemperatureDetected();
+                } else if (temperature < SAFE_TEMPERATURE_THRESHOLD) {
+                    // Safe operating conditions restored
+                    safeOperatingConditions();
```
Yarik-Popov: This should be under its own else if (event == os) then reading its own temperature. Each event should handle its own logic and not depend on other events logic

Yarik-Popov (changes requested):
Over temperature events still need to be handled properly

services/thermal_mgr/thermal_mgr.c
```diff
@@ -42,20 +47,79 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+      if (xQueueReceive(thermalMgrQueueHandle, &event, portMAX_DELAY) == pdPASS) {
+
+        // Read the current temperature from the LM75BD sensor
+        errCode = readTempLM75BD(config->devAddr, &temperature); 
+        LOG_IF_ERROR_CODE(errCode); 
+
+        // Check for read temperature error, skip if temp not read
+        if (errCode != ERR_CODE_SUCCESS) {
+          continue;
+        }
```
Yarik-Popov: You should only be reading the temperature temperature when an event tells us to. Let's say we get an invalid event or in the future some other event doesn't need it, it would be reading when it shouldn't be

services/thermal_mgr/thermal_mgr.c
```diff
@@ -18,6 +19,10 @@ static StackType_t thermalMgrTaskStack[THERMAL_MGR_STACK_SIZE];
 #define THERMAL_MGR_QUEUE_LENGTH 10U
 #define THERMAL_MGR_QUEUE_ITEM_SIZE sizeof(thermal_mgr_event_t)
 
+///Added
+#define THERMAL_MGR_QUEUE_TIMEOUT (pdMS_TO_TICKS(0))  // 100 ms timeout
```
Yarik-Popov: Update comment

Yarik-Popov (changes requested):
You're integration test is not producing the correct output. It gets stuck when the first over temperature event occurs and only prints that over temperature was received but doesn't print the temperatures after that

Yarik-Popov:
This is the output for your simulation run action. You need to look into how to fix this

sys/errors.h
```diff
@@ -13,6 +13,8 @@ typedef enum {
     ERR_CODE_MUTEX_TIMEOUT = 100,
     ERR_CODE_QUEUE_FULL = 101,
     ERR_CODE_INVALID_QUEUE_MSG = 102,
+    ERR_CODE_INVALID_EVENT = 103,  // Added error code
```
Yarik-Popov: This shouldn't be in the freertos error section. Make this in the common section of the errors so in the 0XX range

lm75bd/lm75bd.c
```diff
@@ -26,9 +26,34 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
...
+    RETURN_IF_ERROR_CODE(i2cReceiveFrom(devAddr, buf, sizeof(buf)));
   
-  return ERR_CODE_SUCCESS;
+    // Extract 11-bit data from the 2 bytes received
+    int16_t tempRaw = (buf[0] << 3) | (buf[1] >> 5);
+
+    // Calculate two's complement for negative temperatures
+    if (tempRaw & (1 << 10)) {
+        tempRaw |= 0xF800;
+    }
```
Yarik-Popov: Look at the comment I posted to Luke's question this morning. The second solution is the recommended way of doing things

Yarik-Popov (changes requested):
Minor improvements but other requested changes still not addressed

Yarik-Popov (commented):

PR #185, opened 2024-10-22
Ahmed: Onboarding tasks implemented

elsaigh:
# Purpose
Would like to have my thermalMgr task reviewed as it is having some issue.

# New Changes
- Added the temperature sensor driver functions and the thermal management task that utilizes it.

# Testing
- Ran the integration test and failing to receive any items from the thermal queue.


lm75bd/lm75bd.c
```diff
@@ -27,6 +28,29 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
 
 error_code_t readTempLM75BD(uint8_t devAddr, float *temp) {
```
Yarik-Popov, line 29: Check that temp isn't null before use

lm75bd/lm75bd.c
```diff
@@ -27,6 +28,29 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
...
+  uint16_t LSByte = (uint16_t) receive_buf[1];
+
+  int16_t temp_reg_val = (MSByte << 3) + (LSByte >> 5);
+  // if 11 bit is set, sign is -ve, so convert the 11 bit val to its equivalent 16 bit val
+  // otherwise, sign is +ve, and the 11 bit val will be the same in 16 bits
+  if (temp_reg_val & 0x0400U) {
+    // by simply setting everything above the 11 bits to 1 (top 5 bits here)
+    temp_reg_val |= 0xF800U;
+  }
+  printf("Temperature Register Value: %i\n", temp_reg_val);
```
Yarik-Popov: Remove the printfs

lm75bd/lm75bd.c
```diff
@@ -27,6 +28,29 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
...
 error_code_t readTempLM75BD(uint8_t devAddr, float *temp) {
   /* Implement this driver function */
+
+  uint8_t temp_pointer_byte = 0x00U; // selects temp register at 00 LSB
+  uint8_t *send_buf = &temp_pointer_byte;
+
+  i2cSendTo(devAddr, send_buf, 1);
+
+  uint8_t receive_buf[2];
+  i2cReceiveFrom(devAddr, receive_buf, 2);
```
Yarik-Popov: Wrap both i2c calls inside of the RETURN_IF_ERROR_CODE macro as we don't want to proceed if they fail

lm75bd/lm75bd.c
```diff
@@ -27,6 +28,29 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
 
 error_code_t readTempLM75BD(uint8_t devAddr, float *temp) {
   /* Implement this driver function */
+
+  uint8_t temp_pointer_byte = 0x00U; // selects temp register at 00 LSB
+  uint8_t *send_buf = &temp_pointer_byte;
+
+  i2cSendTo(devAddr, send_buf, 1);
+
+  uint8_t receive_buf[2];
```
Yarik-Popov: Initialize the buffer

lm75bd/lm75bd.c
```diff
@@ -27,6 +28,29 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
 
 error_code_t readTempLM75BD(uint8_t devAddr, float *temp) {
   /* Implement this driver function */
+
+  uint8_t temp_pointer_byte = 0x00U; // selects temp register at 00 LSB
+  uint8_t *send_buf = &temp_pointer_byte;
+
+  i2cSendTo(devAddr, send_buf, 1);
```
Yarik-Popov: Use sizeof(send_buf) instead of 1. Also we use camelCase for variables on the team

lm75bd/lm75bd.c
Yarik-Popov: Use sizeof(receive_buf) instead of 2

lm75bd/lm75bd.c
```diff
@@ -27,6 +28,29 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
...
+  uint16_t MSByte = (uint16_t) receive_buf[0];
+  uint16_t LSByte = (uint16_t) receive_buf[1];
+
+  int16_t temp_reg_val = (MSByte << 3) + (LSByte >> 5);
+  // if 11 bit is set, sign is -ve, so convert the 11 bit val to its equivalent 16 bit val
+  // otherwise, sign is +ve, and the 11 bit val will be the same in 16 bits
+  if (temp_reg_val & 0x0400U) {
+    // by simply setting everything above the 11 bits to 1 (top 5 bits here)
+    temp_reg_val |= 0xF800U;
+  }
```
Yarik-Popov: This code if overly complicated. Doing th following will handle the twos complement conversion for you
```suggestion
int16_t tempRegVal = ((MSByte <<8)|LSByte)>>5;
```

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +44,43 @@ void initThermalSystemManager(lm75bd_config_t *config) {
 
 error_code_t thermalMgrSendEvent(thermal_mgr_event_t *event) {
   /* Send an event to the thermal manager queue */
+  xQueueSend(thermalMgrQueueHandle, event, (TickType_t) 10);
```
Yarik-Popov: - [ ] check that the queue handle isn't null before use, returning invalid state
- [ ] check that the event isn't null
- [ ] use 0 for block time instead of 10 as we don't want to block in an isr

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +44,43 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
 
-static void thermalMgr(void *pvParameters) {
+static void thermalMgr(void *pvParameters) { // what is pvParameters used for?
   /* Implement this task */
+  
+  thermal_mgr_event_t queueEventItemBuffer = {0};
+
   while (1) {
     
+    if (xQueueReceive(thermalMgrQueueHandle, &queueEventItemBuffer, (TickType_t) 10) == pdPASS) {
```
Yarik-Popov: Use portMAX_DELAY instead of 10 as we will block the task until an event is received which is a lot more efficient

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +44,43 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+      
+      float tempC;
+      readTempLM75BD(LM75BD_OBC_I2C_ADDR, &tempC);
```
Yarik-Popov: - [ ] only read when an event tells you too. Let's say we get an event that doesn't require reading then we would read when we shouldn't have
- [ ] in general, check and log the return error code from non void functions

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +44,43 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+
+      if (queueEventItemBuffer.type == THERMAL_MGR_EVENT_MEASURE_TEMP_CMD) {
+        addTemperatureTelemetry(tempC);
+      } else if (queueEventItemBuffer.type == THERMAL_MGR_EVENT_OS_CMD) {
+        if (tempC > 75.0) {
```
Yarik-Popov: Use the macro defined in the lm75bd.h file

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +44,43 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+          overTemperatureDetected();
+        } else {
+          safeOperatingConditions();
+        }
+        // To reset the OS interrupt
+        readTempLM75BD(LM75BD_OBC_I2C_ADDR, &tempC);
```
Yarik-Popov: You don't need to do this, so I would just get rid of these lines

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +44,43 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+  thermal_mgr_event_t queueEventItemBuffer = {0};
+
   while (1) {
     
+    if (xQueueReceive(thermalMgrQueueHandle, &queueEventItemBuffer, (TickType_t) 10) == pdPASS) {
+      
+      float tempC;
+      readTempLM75BD(LM75BD_OBC_I2C_ADDR, &tempC);
+
+      if (queueEventItemBuffer.type == THERMAL_MGR_EVENT_MEASURE_TEMP_CMD) {
```
Yarik-Popov, line 81: Add an else clause that logs that an invalid event was received

Yarik-Popov (changes requested):
Overall good start. Some changes are needed to improve the code quality

services/thermal_mgr/thermal_mgr.c
```diff
@@ -44,17 +45,56 @@ void initThermalSystemManager(lm75bd_config_t *config) {
 error_code_t thermalMgrSendEvent(thermal_mgr_event_t *event) {
   /* Send an event to the thermal manager queue */
 
+  if (thermalMgrQueueHandle == NULL || event == NULL) {
```
Yarik-Popov: The queue handle isn't an arg so it should return invalid state instead of invalid arg

services/thermal_mgr/thermal_mgr.c
```diff
@@ -44,17 +45,56 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+    return ERR_CODE_INVALID_ARG;
+  }
+
+  xQueueSend(thermalMgrQueueHandle, event, (TickType_t) 0);
```
Yarik-Popov: In general, you should always check the return value of non-void functions for an error code and handle it accordingly. In this case, return queue full if an error occurs

services/thermal_mgr/thermal_mgr.c
```diff
@@ -44,17 +45,56 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
 void osHandlerLM75BD(void) {
   /* Implement this function */
+  thermal_mgr_event_t osEvent = {0};
+  osEvent.type = THERMAL_MGR_EVENT_OS_CMD;
+
+  thermalMgrSendEvent(&osEvent);
 }
 
-static void thermalMgr(void *pvParameters) {
+static void thermalMgr(void *pvParameters) { // what is pvParameters used for?
```
Yarik-Popov: This is just a comment, you can resolve this once you read it. `pvParameters` isn't used but i can be used to send values to the task on start at runtime that can't be done via macros at compile time

services/thermal_mgr/thermal_mgr.c
```diff
@@ -44,17 +45,56 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
   while (1) {
     
+    if (xQueueReceive(thermalMgrQueueHandle, &queueEventItemBuffer, (TickType_t) portMAX_DELAY) == pdPASS) {
+
+      if (queueEventItemBuffer.type == THERMAL_MGR_EVENT_MEASURE_TEMP_CMD) {
+
+        float tempC;
+        LOG_IF_ERROR_CODE(readTempLM75BD(LM75BD_OBC_I2C_ADDR, &tempC));
+
+        addTemperatureTelemetry(tempC);
```
Yarik-Popov, line 87: You shouldn't get to this line if the readTemp fails. Add an `if(errCode != success) continue;` line before this

services/thermal_mgr/thermal_mgr.c
```diff
@@ -44,17 +45,56 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+      } else if (queueEventItemBuffer.type == THERMAL_MGR_EVENT_OS_CMD) {
+
+        float tempC;
+        LOG_IF_ERROR_CODE(readTempLM75BD(LM75BD_OBC_I2C_ADDR, &tempC));
+
+        if (tempC > LM75BD_DEFAULT_HYST_THRESH) {
+          overTemperatureDetected();
+        } else {
+          safeOperatingConditions();
+        }
```
Yarik-Popov, line 99: You shouldn't get to these lines if the readTemp fails. Add an `if(errCode != success) continue;` line before this

services/thermal_mgr/thermal_mgr.c
```diff
@@ -44,17 +45,56 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+
+  xQueueSend(thermalMgrQueueHandle, event, (TickType_t) 0);
+
   return ERR_CODE_SUCCESS;
 }
 
 void osHandlerLM75BD(void) {
   /* Implement this function */
+  thermal_mgr_event_t osEvent = {0};
+  osEvent.type = THERMAL_MGR_EVENT_OS_CMD;
```
Yarik-Popov:
```suggestion
  thermal_mgr_event_t osEvent = {.type=THERMAL_MGR_EVENT_OS_CMD};
```
Just so you know, you can also do this.

Yarik-Popov (changes requested):
Some changes are needed to improve error handling

lm75bd/lm75bd.c
```diff
@@ -27,6 +28,28 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
...
+  uint8_t tempPointerByte = 0x00U; // selects temp register at 00 LSB
+
+  uint8_t *sendBuf = &tempPointerByte;
+  RETURN_IF_ERROR_CODE(i2cSendTo(devAddr, sendBuf, sizeof(sendBuf)));
+
+  uint8_t receiveBuf[2] = {0};
+  RETURN_IF_ERROR_CODE(i2cReceiveFrom(devAddr, receiveBuf, sizeof(receiveBuf)));
+
+  uint16_t MSByte = (uint16_t) receiveBuf[0];
+  uint16_t LSByte = (uint16_t) receiveBuf[1];
```
Yarik-Popov, line 47: It might be because you are using uint16_t here and the bit shifting causes issues. Use uint8_t instead as I have seen this work for it before

services/thermal_mgr/thermal_mgr.c
```diff
@@ -44,17 +45,60 @@ void initThermalSystemManager(lm75bd_config_t *config) {
 error_code_t thermalMgrSendEvent(thermal_mgr_event_t *event) {
   /* Send an event to the thermal manager queue */
 
+  if (thermalMgrQueueHandle == NULL || event == NULL) {
```
Yarik-Popov: - [ ] return invalid state if thermal mgr queue handle is null
- [ ] return invalid argument if event is null

services/thermal_mgr/thermal_mgr.c
```diff
@@ -44,17 +45,60 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+    return ERR_CODE_INVALID_STATE;
+  }
+
+  error_code_t sendErrorCode = xQueueSend(thermalMgrQueueHandle, event, (TickType_t) 0);
```
Yarik-Popov: The xQueueSend function doesn't return an error_code_t but a different type of error. The easier way to handle this is just to move this so that its inside the if statement comparison so you don't need to care about the type of the return value

Yarik-Popov (changes requested):
There is only the thermal mgr send function that has requested changes left. Once those are addressed, you should be done the onboarding.

services/thermal_mgr/thermal_mgr.c
```diff
@@ -44,17 +45,63 @@ void initThermalSystemManager(lm75bd_config_t *config) {
 error_code_t thermalMgrSendEvent(thermal_mgr_event_t *event) {
   /* Send an event to the thermal manager queue */
 
+  if (thermalMgrQueueHandle == NULL || event == NULL) {
+    return ERR_CODE_INVALID_STATE;
+  }
```
Yarik-Popov: Event being null will still return invalid state

Yarik-Popov (changes requested):

Yarik-Popov:

PR #196, opened 2024-11-01
Fatma J Onboarding

TheRealGecko:
# Purpose
To check my onboarding

# New Changes
Did onboarding

# Testing
Looked at the test cases in the output

# Outstanding Changes
Idk


lm75bd/lm75bd.c
```diff
@@ -27,7 +27,17 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
 
 error_code_t readTempLM75BD(uint8_t devAddr, float *temp) {
   /* Implement this driver function */
-  
+  uint8_t sendBuf[1] = {0x0U};
+  uint8_t buffer[2];
```
Yarik-Popov: Initialize the buffer

lm75bd/lm75bd.c
```diff
@@ -27,7 +27,17 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
...
+  i2cSendTo (devAddr, sendBuf, 1U);
+  i2cReceiveFrom (devAddr, buffer, 2U);
```
Yarik-Popov: - [ ] wrap both lines inside of a RETURN_IF_ERROR_CODE 
- [ ] use sizeof(sendBuf) instead of 1
- [ ] use sizeof(buffer) instead of 2

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +43,32 @@ void initThermalSystemManager(lm75bd_config_t *config) {
 
 error_code_t thermalMgrSendEvent(thermal_mgr_event_t *event) {
   /* Send an event to the thermal manager queue */
+   xQueueSend(thermalMgrQueueHandle, event, 0); // why u throwing error
```
Yarik-Popov: - [ ] check to make sure that thermal mgr queue isn't null before use, returning invalid state
- [ ] check to make sure that the event isn't null
- [ ] in general, you should be checking the return value of non void functions for errors and handling them accordingly

lm75bd/lm75bd.c
```diff
@@ -27,7 +27,17 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
 
 error_code_t readTempLM75BD(uint8_t devAddr, float *temp) {
```
Yarik-Popov, line 28: Check that temp isn't null, always check that ptrs aren't null before use

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +43,32 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
 void osHandlerLM75BD(void) {
   /* Implement this function */
+  float temp;
+  readTempLM75BD(LM75BD_OBC_I2C_ADDR, &temp); 
+
+  if(temp > 80) {
+    overTemperatureDetected();
+  } else if(temp < 75) {
+    safeOperatingConditions();
+  }
```
Yarik-Popov: Do not read inside the os handler as it may block. Send an event and have the thermalMgr task function handle it

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +43,32 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
 }
 
 void osHandlerLM75BD(void) {
   /* Implement this function */
+  float temp;
+  readTempLM75BD(LM75BD_OBC_I2C_ADDR, &temp); 
+
+  if(temp > 80) {
+    overTemperatureDetected();
+  } else if(temp < 75) {
```
Yarik-Popov: Use the macros defined in the lm75bd.h instead of hardocding the values

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +43,32 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
 }
 
 static void thermalMgr(void *pvParameters) {
   /* Implement this task */
   while (1) {
-    
+    thermal_mgr_event_t data = *(thermal_mgr_event_t *) pvParameters;
+   if(xQueueReceive(thermalMgrQueueHandle, pvParameters, 0) == pdTRUE && data.type == THERMAL_MGR_EVENT_MEASURE_TEMP_CMD) { // ensure about the tick time + how to checl the type ajsdkfk
+      float temp;
+      readTempLM75BD(LM75BD_OBC_I2C_ADDR, &temp);
```
Yarik-Popov: Check the return value for an error code, log it if it exists and `continue`. we don't want to add a temp value that failed to read we don't know if temp will be a valid value

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +43,32 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+    safeOperatingConditions();
+  }
 }
 
 static void thermalMgr(void *pvParameters) {
   /* Implement this task */
   while (1) {
-    
+    thermal_mgr_event_t data = *(thermal_mgr_event_t *) pvParameters;
+   if(xQueueReceive(thermalMgrQueueHandle, pvParameters, 0) == pdTRUE && data.type == THERMAL_MGR_EVENT_MEASURE_TEMP_CMD) { // ensure about the tick time + how to checl the type ajsdkfk
```
Yarik-Popov: - [ ] Use portMAX_DELAY instead of 0
- [ ] you should be logging if an invalid event was received

lm75bd/lm75bd.c
```diff
@@ -27,7 +27,17 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
...
+  uint8_t buffer[2];
+  i2cSendTo (devAddr, sendBuf, 1U);
+  i2cReceiveFrom (devAddr, buffer, 2U);
+  uint8_t a = 1U;
+  uint8_t mask = a << 7;
+  if (!(buffer[0] & mask)) {
+    *temp = ((buffer[0] << 3) | (buffer[1] >> 5)) * 0.125;
+  } else {
+    *temp = -(((~((buffer[0] << 3) | (buffer[1] >> 5))) & (0b0000011111111111)) + 1) * 0.125;
+  }
```
Yarik-Popov: the code is overly complicated
```suggestion
    int16_t temperature = ((buffer[0] << 8) | buffer[1]) >> 5;
    *temp = temperature * 0.125;
```
You might need to do some casting to get this to work

Yarik-Popov (changes requested):
Overall good start, some changes are needed to improve the code quality

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +44,47 @@ void initThermalSystemManager(lm75bd_config_t *config) {
 
 error_code_t thermalMgrSendEvent(thermal_mgr_event_t *event) {
   /* Send an event to the thermal manager queue */
-
-  return ERR_CODE_SUCCESS;
+  if(thermalMgrQueueHandle != NULL && event != NULL) {
```
Yarik-Popov: - [ ] return invalid argument if event is null
- [ ] return invalid state if thermal mgr queue handle is null as its not an argument but a state variable

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +44,47 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+    xQueueSend(thermalMgrQueueHandle, event, 0);
```
Yarik-Popov: In general, you should check the return of non-void functions for error codes and handle them

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +44,47 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+  thermal_mgr_event_t event;
+  event.type = THERMAL_MGR_EVENT_OS_INTERRUPT;
+  thermalMgrSendEvent(&event);
 }
 
 static void thermalMgr(void *pvParameters) {
   /* Implement this task */
   while (1) {
-    
+    thermal_mgr_event_t data = *(thermal_mgr_event_t *) pvParameters;
```
Yarik-Popov: PvParameters is not of type thermal mgr event but of the lm75bd config so you shouldn't do this. Just `thermal_mgr_event_t event={0};`

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +44,47 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+
+    error_code_t errCodeEvent = xQueueReceive(thermalMgrQueueHandle, pvParameters, portMAX_DELAY);
```
Yarik-Popov: The return type of xQueueReceive isn't error_code_t. Only the functions we write on the team use that custom error code. Libraries have there own

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +44,47 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+      float temp;
+      if (data.type == THERMAL_MGR_EVENT_MEASURE_TEMP_CMD) {
+        error_code_t errCode = readTempLM75BD(LM75BD_OBC_I2C_ADDR, &temp); 
+        if(errCode != ERR_CODE_SUCCESS) {
+          LOG_ERROR_CODE(errCode);
+          continue;
+        }
+        addTemperatureTelemetry(temp);
+      } else if(data.type == THERMAL_MGR_EVENT_OS_INTERRUPT) {
+        if(temp > LM75BD_DEFAULT_OT_THRESH) {
```
Yarik-Popov, line 90: You should be reading the temperature when you get an over temperature event as if temp wasn't read for a while the value might be stale and not represent the current temperature

lm75bd/lm75bd.c
```diff
@@ -27,7 +27,22 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
...
+  RETURN_IF_ERROR_CODE(i2cSendTo (devAddr, sendBuf, sizeof(sendBuf)));
+  RETURN_IF_ERROR_CODE(i2cReceiveFrom (devAddr, buffer, sizeof(buffer)));
+  uint8_t a = 1U;
+  uint8_t mask = a << 7;
+  int16_t temperature = ((buffer[0] << 8) | buffer[1]) >> 5;
+  if (!(buffer[0] & mask)) {
+    *temp = temperature * 0.125;
+  } else {
+    *temp = -((~temperature & 0b0000011111111111) + 1) * 0.125;
+  }
```
Yarik-Popov: Out of these lines, you only need lines 40 and 42. You might need to cast the temperature before assign it to temp

Yarik-Popov (changes requested):
A step in the right direction, there are still some changes that need to be handled

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +44,60 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
-
-  return ERR_CODE_SUCCESS;
+  if(event == NULL) {
+    return ERR_CODE_INVALID_ARG;
+  } else if(thermalMgrQueueHandle == NULL) {
+    return ERR_CODE_INVALID_STATE;
+  } else {
+    BaseType_t errCode = xQueueSend(thermalMgrQueueHandle, event, 0);
+    if(errCode != pdTRUE) {
+      LOG_ERROR_CODE(errCode);
```
Yarik-Popov: Return error queue full instead of logging it. Also this errCode is a different type than the error_code_t

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +44,60 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+        }
+
+        if(temp > LM75BD_DEFAULT_OT_THRESH) {
+          overTemperatureDetected();
+        } else if(temp < LM75BD_DEFAULT_HYST_THRESH) {
+          safeOperatingConditions();
+        }
+      }
+    } else {
+        LOG_ERROR_CODE(errCodeEvent);
```
Yarik-Popov: You should be logging an invalid event (you will need to create or search for it in our sys/errors.h) as the type is different

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +44,60 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
 
 static void thermalMgr(void *pvParameters) {
   /* Implement this task */
   while (1) {
-    
+    thermal_mgr_event_t data = {0};
+
+    BaseType_t errCodeEvent = xQueueReceive(thermalMgrQueueHandle, pvParameters, portMAX_DELAY);
+
+    data = *(thermal_mgr_event_t *) pvParameters;
```
Yarik-Popov: - [ ] Don't use pvParameters here, just use &data
- [ ] Remove the data = ...; line

lm75bd/lm75bd.c
```diff
@@ -27,7 +27,17 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
 
 error_code_t readTempLM75BD(uint8_t devAddr, float *temp) {
   /* Implement this driver function */
+  error_code_t errCode;
   
+  if(temp == NULL) return ERR_CODE_INVALID_ARG;
+
+  uint8_t sendBuf[1] = {0x0U};
+  uint8_t buffer[2] = {0U, 0U};
```
Yarik-Popov: Just so you know you can just use {0} and it will zero out all the fields

Yarik-Popov (changes requested):
Should be last set of changes

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +44,58 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
-  return ERR_CODE_SUCCESS;
+  if(event == NULL) {
+    return ERR_CODE_INVALID_ARG;
+  } else if(thermalMgrQueueHandle == NULL) {
+    return ERR_CODE_INVALID_STATE;
+  } else {
+    BaseType_t errCode = xQueueSend(thermalMgrQueueHandle, event, 0);
+    if(errCode != pdTRUE) {
+      return ERR_CODE_QUEUE_FULL;
+    }
```
Navtajh04, line 55: nit:
```suggestion
    if(xQueueSend(thermalMgrQueueHandle, event, 0) != pdTRUE) {
      return ERR_CODE_QUEUE_FULL;
    }
```

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +44,58 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
 
 static void thermalMgr(void *pvParameters) {
   /* Implement this task */
   while (1) {
-    
+    thermal_mgr_event_t data = {0};
+
+    BaseType_t errCodeEvent = xQueueReceive(thermalMgrQueueHandle, &data, portMAX_DELAY);
+
+    if(errCodeEvent == pdTRUE) {
```
Navtajh04, line 74: nit: in general if you arent going to use the errCode anywhere outside of the if condition, its better to just call the function in the if condition and compare it there rather than using another variable. Like in the comment shown above

Navtajh04 (commented):
LGTM, left some minor comments as an FYI but you dont need to make those changes. You can send me your email and ill add you to the notion then u can take a look at the FW task board for a task that interests you

PR #197, opened 2024-11-02
Seth Onboarding

Seth-Winton:
Just finished and it passed all the automated tests.

### DIFF
--- a/lm75bd/lm75bd.c
+++ b/lm75bd/lm75bd.c
@@ -9,6 +9,7 @@
 
 /* LM75BD Registers (p.8) */
 #define LM75BD_REG_CONF 0x01U  /* Configuration Register (R/W) */
+#define LM75BD_REG_TEMP 0x00  /* Temperature register*/
 
 error_code_t lm75bdInit(lm75bd_config_t *config) {
   error_code_t errCode;
@@ -26,7 +27,24 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
 }
 
 error_code_t readTempLM75BD(uint8_t devAddr, float *temp) {
-  /* Implement this driver function */
+  error_code_t errCode;
+  uint8_t tempReg = LM75BD_REG_TEMP;
+  errCode = i2cSendTo(devAddr, &tempReg, 1); /* Select temperature register*/
+  if (errCode != ERR_CODE_SUCCESS) return errCode;
+
+  uint8_t buff[2] = {0};
+  errCode = i2cReceiveFrom(devAddr, buff, 2); /* Read temp data */
+  if (errCode != ERR_CODE_SUCCESS) return errCode;
+  
+  uint16_t val = ((buff[0] << 3) + (buff[1] >> 5)); /* Combine MSB and LSB */
+  if (val >> 10) {
+    /* Negative */
+    *temp = ((~val + 1) & 0x7FF) * -0.125;
+  }
+  else {
+    /* Positive */
+    *temp = val * 0.125;
+  }
   
   return ERR_CODE_SUCCESS;
 }
--- a/lm75bd/lm75bd.c
+++ b/lm75bd/lm75bd.c
@@ -27,24 +27,18 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
 }
 
 error_code_t readTempLM75BD(uint8_t devAddr, float *temp) {
+  if (temp == NULL) {return ERR_CODE_INVALID_ARG;}
   error_code_t errCode;
+
   uint8_t tempReg = LM75BD_REG_TEMP;
-  errCode = i2cSendTo(devAddr, &tempReg, 1); /* Select temperature register*/
-  if (errCode != ERR_CODE_SUCCESS) return errCode;
+  RETURN_IF_ERROR_CODE(i2cSendTo(devAddr, &tempReg, sizeof(tempReg))); /* Select temperature register*/
 
   uint8_t buff[2] = {0};
-  errCode = i2cReceiveFrom(devAddr, buff, 2); /* Read temp data */
-  if (errCode != ERR_CODE_SUCCESS) return errCode;
+  RETURN_IF_ERROR_CODE(i2cReceiveFrom(devAddr, buff, sizeof(buff))); /* Read temp data */
   
-  uint16_t val = ((buff[0] << 3) + (buff[1] >> 5)); /* Combine MSB and LSB */
-  if (val >> 10) {
-    /* Negative */
-    *temp = ((~val + 1) & 0x7FF) * -0.125;
-  }
-  else {
-    /* Positive */
-    *temp = val * 0.125;
-  }
+  // TODO make this cleaner
+  int16_t val = (buff[0] << 8) | buff[1]; /* Combine MSB and LSB */
+  *temp = (float)(val >> 5) * 0.125;
   
   return ERR_CODE_SUCCESS;
 }
@@ -84,8 +78,7 @@ error_code_t writeConfigLM75BD(uint8_t devAddr, uint8_t osFaultQueueSize, uint8_
   buff[1] |= (osOperationMode << 1);
   buff[1] |= devOperationMode;
 
-  errCode = i2cSendTo(LM75BD_OBC_I2C_ADDR, buff, CONF_WRITE_BUFF_SIZE);
-  if (errCode != ERR_CODE_SUCCESS) return errCode;
+  RETURN_IF_ERROR_CODE(i2cSendTo(LM75BD_OBC_I2C_ADDR, buff, CONF_WRITE_BUFF_SIZE));
 
   return ERR_CODE_SUCCESS;
 }
--- a/lm75bd/lm75bd.c
+++ b/lm75bd/lm75bd.c
@@ -9,7 +9,7 @@
 
 /* LM75BD Registers (p.8) */
 #define LM75BD_REG_CONF 0x01U  /* Configuration Register (R/W) */
-#define LM75BD_REG_TEMP 0x00  /* Temperature register*/
+#define LM75BD_REG_TEMP 0x00U  /* Temperature register*/
 
 error_code_t lm75bdInit(lm75bd_config_t *config) {
   error_code_t errCode;
--- a/lm75bd/lm75bd.h
+++ b/lm75bd/lm75bd.h
@@ -5,7 +5,7 @@
 #include <stdint.h>
 
 /* LM75BD I2C Device Address */
-#define LM75BD_OBC_I2C_ADDR /* Define the address here */
+#define LM75BD_OBC_I2C_ADDR 0x4F
 
 /* LM75BD Configuration Values */
 #define LM75BD_DEV_OP_MODE_NORMAL 0x00U
--- a/lm75bd/lm75bd.h
+++ b/lm75bd/lm75bd.h
@@ -5,7 +5,7 @@
 #include <stdint.h>
 
 /* LM75BD I2C Device Address */
-#define LM75BD_OBC_I2C_ADDR 0x4F
+#define LM75BD_OBC_I2C_ADDR 0x4FU
 
 /* LM75BD Configuration Values */
 #define LM75BD_DEV_OP_MODE_NORMAL 0x00U
--- a/lm75bd/lm75bd.h
+++ b/lm75bd/lm75bd.h
@@ -5,7 +5,7 @@
 #include <stdint.h>
 
 /* LM75BD I2C Device Address */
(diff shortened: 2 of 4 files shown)


lm75bd/lm75bd.c
```diff
@@ -26,7 +27,24 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
 }
 
 error_code_t readTempLM75BD(uint8_t devAddr, float *temp) {
```
Yarik-Popov, line 29: Check that temp isn't null before use

lm75bd/lm75bd.c
```diff
@@ -26,7 +27,24 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
...
-  /* Implement this driver function */
+  error_code_t errCode;
+  uint8_t tempReg = LM75BD_REG_TEMP;
+  errCode = i2cSendTo(devAddr, &tempReg, 1); /* Select temperature register*/
+  if (errCode != ERR_CODE_SUCCESS) return errCode;
```
kepler452b123: nit: use RETURN_IF_ERROR_CODE here

lm75bd/lm75bd.c
```diff
@@ -26,7 +27,24 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
 }
 
 error_code_t readTempLM75BD(uint8_t devAddr, float *temp) {
-  /* Implement this driver function */
+  error_code_t errCode;
+  uint8_t tempReg = LM75BD_REG_TEMP;
+  errCode = i2cSendTo(devAddr, &tempReg, 1); /* Select temperature register*/
```
kepler452b123: use sizeof(tempReg) instead of 1 for more flexibility

lm75bd/lm75bd.c
```diff
@@ -26,7 +27,24 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
...
+  if (errCode != ERR_CODE_SUCCESS) return errCode;
+
+  uint8_t buff[2] = {0};
+  errCode = i2cReceiveFrom(devAddr, buff, 2); /* Read temp data */
```
kepler452b123: same here

lm75bd/lm75bd.h
```diff
@@ -5,7 +5,7 @@
 #include <stdint.h>
 
 /* LM75BD I2C Device Address */
-#define LM75BD_OBC_I2C_ADDR /* Define the address here */
+#define LM75BD_OBC_I2C_ADDR 0x4F
```
kepler452b123, line 8: nit make it 0x4FU for consistency (just how we do it thru out our codebase)

services/thermal_mgr/thermal_mgr.c
```diff
@@ -42,20 +43,46 @@ void initThermalSystemManager(lm75bd_config_t *config) {
 }
 
 error_code_t thermalMgrSendEvent(thermal_mgr_event_t *event) {
-  /* Send an event to the thermal manager queue */
-
+  // Send an event to the thermal manager queue
+  if (xQueueSend(thermalMgrQueueHandle, event, 1000) !=  pdTRUE) {
+    return ERR_CODE_QUEUE_FULL;
+  }
```
kepler452b123: make this 1 tick, since this is called from an ISR

services/thermal_mgr/thermal_mgr.c
```diff
@@ -42,20 +43,46 @@ void initThermalSystemManager(lm75bd_config_t *config) {
 }
 
 error_code_t thermalMgrSendEvent(thermal_mgr_event_t *event) {
-  /* Send an event to the thermal manager queue */
-
+  // Send an event to the thermal manager queue
```
kepler452b123: check if event is null. if it is, return invalid arg error. always check if pointers are null

kepler452b123: check if the thermal mgr queue exists. if it doesn't, return err code invalid state

services/thermal_mgr/thermal_mgr.c
```diff
@@ -42,20 +43,46 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
 }
 
 void osHandlerLM75BD(void) {
-  /* Implement this function */
+  // Send event to handle OS interrupt
+  thermal_mgr_event_t event;
+  event.type = THERMAL_MGR_EVENT_HANDLE_OS_CMD;
+  if (xQueueSendFromISR(thermalMgrQueueHandle, &event, NULL) != pdTRUE) {
+    LOG_ERROR_CODE(ERR_CODE_QUEUE_FULL);
+  }
```
kepler452b123: should just be calling thermalMgrSendEvent from this to send the event. avoid logging directly from ISRs, do this in thermal mgr send event function

services/thermal_mgr/thermal_mgr.c
```diff
@@ -42,20 +43,46 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
 }
 
 static void thermalMgr(void *pvParameters) {
-  /* Implement this task */
-  while (1) {
-    
-  }
+	while (1) {
+		if (xQueueReceive(thermalMgrQueueHandle, pvParameters, 0) == pdTRUE) {
```
kepler452b123: ticks to block when receiving should be portmax_delay. we want it to block indefinitely until an event is received

services/thermal_mgr/thermal_mgr.c
```diff
@@ -42,20 +43,46 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+      thermal_mgr_event_t event = *(thermal_mgr_event_t *) pvParameters;
+      
+      // Handle event received from queue
+      if (event.type == THERMAL_MGR_EVENT_MEASURE_TEMP_CMD) {
+        float temp;
+        readTempLM75BD(LM75BD_OBC_I2C_ADDR, &temp);
```
kepler452b123: you should be checking if the driver function returns err_code_success before continuing. add a catch all else statement which returns the appropriate error code and use break in these cases if the driver function does not return success

kepler452b123 (changes requested):

services/thermal_mgr/thermal_mgr.c
Yarik-Popov: This should be 0 as a block time of > 0 in an isr is undefined behavior

lm75bd/lm75bd.c
Yarik-Popov: I wouldn't consider this to be a nit as the RETURN_IF_ERROR_CODE macro also logs an error which is useful in debugging and is not done here

services/thermal_mgr/thermal_mgr.c
```diff
@@ -42,20 +43,46 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+        addTemperatureTelemetry(temp);
+      }
+      else if (event.type == THERMAL_MGR_EVENT_HANDLE_OS_CMD) {
```
Yarik-Popov, line 81: Add an else clause that logs if an invalid event was received

services/thermal_mgr/thermal_mgr.c
```diff
@@ -42,20 +43,46 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
 }
 
 static void thermalMgr(void *pvParameters) {
-  /* Implement this task */
-  while (1) {
-    
-  }
+	while (1) {
+		if (xQueueReceive(thermalMgrQueueHandle, pvParameters, 0) == pdTRUE) {
+      thermal_mgr_event_t event = *(thermal_mgr_event_t *) pvParameters;
```
Yarik-Popov: pvParamaters isn't an event type but the config for the lm75bd. Since you don't use the config I would get rid of this line

services/thermal_mgr/thermal_mgr.c
Yarik-Popov: But to use define an event structure instead of using pvParameters. See my comment below

lm75bd/lm75bd.c
```diff
@@ -26,7 +27,24 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
...
+  
+  uint16_t val = ((buff[0] << 3) + (buff[1] >> 5)); /* Combine MSB and LSB */
+  if (val >> 10) {
+    /* Negative */
+    *temp = ((~val + 1) & 0x7FF) * -0.125;
+  }
+  else {
+    /* Positive */
+    *temp = val * 0.125;
+  }
```
Yarik-Popov: Code is overly complex.
```suggestion
    int16_t val = ((buff[0] << 8) | buff[1]) >> 5; /* Combine MSB and LSB */
    *temp = val * 0.125;
```
int16_t will handle the conversion for twos complement for you

Yarik-Popov (changes requested):
Overall good start. Some other changes are needed.

services/thermal_mgr/thermal_mgr.c
Seth-Winton: Actually, I made it so this doesn't get called from the ISR (the osHandler ISR calls its own QueueSend function). Should it still be 0?

services/thermal_mgr/thermal_mgr.c
Seth-Winton: It looks like the xQueueSend function is not meant to be called through an ISR, so I used this instead. Maybe I should use xQueueSendFromISR inside thermalMgrSendEvent and then call thermalMgrSendEvent in this ISR?

Seth-Winton: Does this also mean that I shouldn't be logging anything in thermalMgrSendEvent since it would get called from the ISR?

lm75bd/lm75bd.c
Seth-Winton: Unfortunately, it seems this way doesn't pass the TestReadTempLm75bdNegTempSuccess. Is there still a way to clean this up while maintaining negative values?

Seth-Winton: I just figured this out actually

services/thermal_mgr/thermal_mgr.c
```diff
@@ -42,20 +43,63 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
   return ERR_CODE_SUCCESS;
 }
 
 void osHandlerLM75BD(void) {
-  /* Implement this function */
+  // Send event to handle OS interrupt
+  thermal_mgr_event_t event;
+  event.type = THERMAL_MGR_EVENT_HANDLE_OS_CMD;
+  if (xQueueSendFromISR(thermalMgrQueueHandle, &event, NULL) != pdTRUE) {
+    LOG_ERROR_CODE(ERR_CODE_QUEUE_FULL);
```
Yarik-Popov: You shouldn't be logging in an isr as logging may block

services/thermal_mgr/thermal_mgr.c
```diff
@@ -2,6 +2,7 @@
 #include "errors.h"
 #include "lm75bd.h"
 #include "console.h"
+#include "logging/logging.h"
```
Yarik-Popov: The way we do things on the subteam is just by including the last part of the include so `#include "logging.h"` is the expected way of doing things

services/thermal_mgr/thermal_mgr.c
```diff
@@ -42,20 +43,63 @@ void initThermalSystemManager(lm75bd_config_t *config) {
 }
 
 error_code_t thermalMgrSendEvent(thermal_mgr_event_t *event) {
-  /* Send an event to the thermal manager queue */
+  if (event == NULL) {return ERR_CODE_INVALID_ARG;}
+  if (thermalMgrQueueHandle == NULL) {return ERR_CODE_INVALID_STATE;}
 
+  // Send an event to the thermal manager queue
+  if (xQueueSend(thermalMgrQueueHandle, event, 1000) !=  pdTRUE) {
```
Yarik-Popov: I would define a macro for the 1000

lm75bd/lm75bd.h
```diff
@@ -5,7 +5,7 @@
 #include <stdint.h>
 
 /* LM75BD I2C Device Address */
-#define LM75BD_OBC_I2C_ADDR /* Define the address here */
+#define LM75BD_OBC_I2C_ADDR 0x4FU
```
Yarik-Popov: On the subteam we don't include the U ending

Yarik-Popov (changes requested):
Mainly style changes left. Once these are addressed, you should be done the onboarding

Yarik-Popov (commented):

PR #199, opened 2024-11-22
faris pull

Falmousl:
# Purpose
Creating PR for review. I wanted to mention that I have two files of this on my computer, even though they have the same code one gives me a build error so if there is an error in the build let me know because I have a working version on my computer but in that scenario I messed up the PR.

# New Changes
- Explain new changes

# Testing

# Outstanding Changes

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -9,6 +9,7 @@ jobs:
 
     - name: Setup
       run: |
+        sudo apt-get update
         sudo apt-get -y install build-essential cmake
 
     - name: Create binary directory
--- a/.github/workflows/run.yml
+++ b/.github/workflows/run.yml
@@ -9,6 +9,7 @@ jobs:
 
     - name: Setup
       run: |
+        sudo apt-get update
         sudo apt-get -y install build-essential cmake expect
 
     - name: Create binary directory


lm75bd/lm75bd.c
```diff
@@ -26,8 +26,15 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
 }
 
 error_code_t readTempLM75BD(uint8_t devAddr, float *temp) {
-  /* Implement this driver function */
-  
+  uint8_t pointer_byte = 0x00;
+  i2cSendTo(devAddr, &pointer_byte, 1);
```
kepler452b123: use sizeOf instead of hardcoding the size of the buffer

lm75bd/lm75bd.c
```diff
@@ -26,8 +26,15 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
...
+  uint8_t bufStore[2];
+  i2cReceiveFrom(LM75BD_OBC_I2C_ADDR, bufStore, 2);
```
kepler452b123: same as above

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +43,37 @@ void initThermalSystemManager(lm75bd_config_t *config) {
 
 error_code_t thermalMgrSendEvent(thermal_mgr_event_t *event) {
   /* Send an event to the thermal manager queue */
-
```
kepler452b123: check if the queue is null, if it is then send an invalid state error. also check if the event is null (generally good to check if any pointer args are null before performing actions with them). also handle if the xQueueSend does not work (this would happen if the queue is full, return valid error)

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +43,37 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+  xQueueSend(thermalMgrQueueHandle, event, portMAX_DELAY);
```
kepler452b123: think about what it would mean if we used portMAX_DELAY for xQueueSend, is this the behavior we want?

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +43,37 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
   /* Implement this task */
   while (1) {
-    
+    thermal_mgr_event_t event;
+     if (xQueueReceive(thermalMgrQueueHandle, &event, portMAX_DELAY) == pdTRUE){
+      float curTemp;
+      float interruptTemp;
+        switch (event.type) {
+          case THERMAL_MGR_EVENT_MEASURE_TEMP_CMD:
+            readTempLM75BD(LM75BD_OBC_I2C_ADDR, &curTemp);
```
kepler452b123: handle the case where readTempLM75BD does not return ERR_CODE_SUCCESS (log the error code).

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +43,37 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+            addTemperatureTelemetry(curTemp);
+          case THERMAL_MGR_EVENT_OS_INTERRUPT:
+            readTempLM75BD(LM75BD_OBC_I2C_ADDR, &interruptTemp); 
+            if (interruptTemp >= 80) {
+                overTemperatureDetected();
+            } else {
+                safeOperatingConditions();  
+            }
```
kepler452b123: add another case if the event does not match either of these types (log error code for invalid queue message)

lm75bd/lm75bd.c
```diff
@@ -26,8 +26,15 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
...
-  /* Implement this driver function */
-  
+  uint8_t pointer_byte = 0x00;
+  i2cSendTo(devAddr, &pointer_byte, 1);
+  uint8_t bufStore[2];
+  i2cReceiveFrom(LM75BD_OBC_I2C_ADDR, bufStore, 2); 
+  int16_t tempVal = (((bufStore[0] << 8) | bufStore[1]) >> 5); // Combine the two bytes and shift the value to get the temperature
+  if (((bufStore[0] >> 7) & 1) == 1) { // Check if the temperature is negative
+    tempVal |= 0xF800; 
+  }
```
kepler452b123: int16_t type should just convert the negative temp for you

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +43,37 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+     if (xQueueReceive(thermalMgrQueueHandle, &event, portMAX_DELAY) == pdTRUE){
+      float curTemp;
+      float interruptTemp;
+        switch (event.type) {
+          case THERMAL_MGR_EVENT_MEASURE_TEMP_CMD:
+            readTempLM75BD(LM75BD_OBC_I2C_ADDR, &curTemp);
+            addTemperatureTelemetry(curTemp);
+          case THERMAL_MGR_EVENT_OS_INTERRUPT:
+            readTempLM75BD(LM75BD_OBC_I2C_ADDR, &interruptTemp); 
+            if (interruptTemp >= 80) {
```
kepler452b123: check lm75bd.h for macros to use instead of the value 80

kepler452b123 (changes requested):
some changes for param checks, style, and error handling

PR #205, opened 2025-01-15
Harsukrit Pall's Onboarding (Changed)

HarsukritP:
# Purpose

Harsukrit Pall's Onboarding for the CubeSat Thermal Firmware, implemented and worked on the temperature sensor driver implementation, Thermal management task, OS interrupt handler, temperature monitoring and alerts with all tests passed.

### DIFF
--- a/lm75bd/lm75bd.c
+++ b/lm75bd/lm75bd.c
@@ -26,22 +26,24 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
 }
 
 error_code_t readTempLM75BD(uint8_t devAddr, float *temp) {
-    error_code_t errCode;
-    uint8_t tempReg[2];
-    uint8_t pointerReg = 0x00;
-
     if (temp == NULL) {
         return ERR_CODE_INVALID_ARG;
     }
+    error_code_t errCode;
+    uint8_t tempReg[2] = {0};
+    uint8_t pointerReg = 0x00;
 
-    RETURN_IF_ERROR_CODE(i2cSendTo(devAddr, &pointerReg, 1));
-
-    RETURN_IF_ERROR_CODE(i2cReceiveFrom(devAddr, tempReg, 2));
-
+    RETURN_IF_ERROR_CODE(i2cSendTo(devAddr, &pointerReg, sizeof(pointerReg)));
+    RETURN_IF_ERROR_CODE(i2cReceiveFrom(devAddr, tempReg, sizeof(tempReg)));
+    
+    // Combine 2 bytes into 16-bit value
     int16_t rawTemp = (tempReg[0] << 8) | tempReg[1];
     
+    // Temperature data is in 9-bit resolution (7 integer bits + 2 fractional bits)
+    // Right-shift by 7 to align with the temperature format
     rawTemp >>= 7;
 
+    // Convert to Celsius (each bit represents 0.5°C)
     *temp = rawTemp * 0.5f;
 
     return ERR_CODE_SUCCESS;
--- a/lm75bd/lm75bd.c
+++ b/lm75bd/lm75bd.c
@@ -26,9 +26,25 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
 }
 
 error_code_t readTempLM75BD(uint8_t devAddr, float *temp) {
-  /* Implement this driver function */
-  
-  return ERR_CODE_SUCCESS;
+    error_code_t errCode;
+    uint8_t tempReg[2];
+    uint8_t pointerReg = 0x00;
+
+    if (temp == NULL) {
+        return ERR_CODE_INVALID_ARG;
+    }
+
+    RETURN_IF_ERROR_CODE(i2cSendTo(devAddr, &pointerReg, 1));
+
+    RETURN_IF_ERROR_CODE(i2cReceiveFrom(devAddr, tempReg, 2));
+
+    int16_t rawTemp = (tempReg[0] << 8) | tempReg[1];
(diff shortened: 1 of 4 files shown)


lm75bd/lm75bd.c
```diff
@@ -26,9 +26,25 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
 }
 
 error_code_t readTempLM75BD(uint8_t devAddr, float *temp) {
-  /* Implement this driver function */
-  
-  return ERR_CODE_SUCCESS;
+    error_code_t errCode;
+    uint8_t tempReg[2];
```
kepler452b123: it's good practice to initialize arrays (just initialize to 0)

lm75bd/lm75bd.c
```diff
@@ -26,9 +26,25 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
...
+    uint8_t pointerReg = 0x00;
+
+    if (temp == NULL) {
+        return ERR_CODE_INVALID_ARG;
+    }
+
+    RETURN_IF_ERROR_CODE(i2cSendTo(devAddr, &pointerReg, 1));
+
+    RETURN_IF_ERROR_CODE(i2cReceiveFrom(devAddr, tempReg, 2));
```
kepler452b123: nit: use sizeOf here instead of hardcoding size as 1 and 2

lm75bd/lm75bd.c
```diff
@@ -26,9 +26,25 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
...
-  /* Implement this driver function */
-  
-  return ERR_CODE_SUCCESS;
+    error_code_t errCode;
+    uint8_t tempReg[2];
+    uint8_t pointerReg = 0x00;
+
+    if (temp == NULL) {
+        return ERR_CODE_INVALID_ARG;
+    }
```
kepler452b123, line 31: this check should be done at the start of the function

services/thermal_mgr/thermal_mgr.c
```diff
@@ -38,24 +38,46 @@ void initThermalSystemManager(lm75bd_config_t *config) {
   thermalMgrQueueHandle = xQueueCreateStatic(
     THERMAL_MGR_QUEUE_LENGTH, THERMAL_MGR_QUEUE_ITEM_SIZE,
     thermalMgrQueueStorageArea, &thermalMgrQueueBuffer);
-
 }
 
 error_code_t thermalMgrSendEvent(thermal_mgr_event_t *event) {
-  /* Send an event to the thermal manager queue */
-
-  return ERR_CODE_SUCCESS;
```
kepler452b123, line 47: check if event pointer is null (always check if pointer args are null). if it is, then return invalid arg

services/thermal_mgr/thermal_mgr.c
```diff
@@ -38,24 +38,46 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+    if (xQueueSend(thermalMgrQueueHandle, event, 0) != pdTRUE) {
+        return ERR_CODE_QUEUE_FULL;
+    }
```
kepler452b123, line 55: check if the queue exists. if it doesn't then return invalid state err code

services/thermal_mgr/thermal_mgr.c
```diff
@@ -38,24 +38,46 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+                        isOverTemp = 1;
+                    } else if (currentTemp < config->hysteresisThresholdCelsius && isOverTemp) {
+                        safeOperatingConditions();
+                        isOverTemp = 0;
+                    }
+                }
+            }
+        }
+    }
 }
```
kepler452b123: misunderstanding here:
we do not want to check for over temperature and add telemetry every time an event is received. there should be two events (add another one in thermal_mgr.c). the one that checks for overtemperature is an interrupt sent by the osHandler. the other one is just a command to measure and log the temperature as telemetry. every time you receive an event, you should check the event type and then do the corresponding process for the event

kepler452b123 (changes requested):

services/thermal_mgr/thermal_mgr.c
```diff
@@ -38,24 +38,57 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+    error_code_t errCode;
+    uint8_t isOverTemp = 0;
+
+    while (1) {
+        if (xQueueReceive(thermalMgrQueueHandle, &event, portMAX_DELAY) == pdTRUE) {
+            if (event.type == THERMAL_MGR_EVENT_MEASURE_TEMP_CMD) {
+                errCode = readTempLM75BD(config->devAddr, &currentTemp);
+                if (errCode == ERR_CODE_SUCCESS) {
+                    addTemperatureTelemetry(currentTemp);
+                }
```
kepler452b123, line 78: you should be using log if error code to log in case the driver function fails

services/thermal_mgr/thermal_mgr.c
```diff
@@ -38,24 +38,57 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+            } else if (event.type == THERMAL_MGR_EVENT_OS_INTERRUPT) {
+                errCode = readTempLM75BD(config->devAddr, &currentTemp);
```
kepler452b123: same here

kepler452b123 (changes requested):

kepler452b123 (approved):

PR #213, opened 2025-02-02
Felix Yang Onboarding

FxJYg:
# Purpose
- Finish implementing required functions for Onboarding

# New Changes
- Shown as above

# Testing
= Passed the tests for Reading temperature 
- Integration test seems to be fine

# Outstanding Changes
-none

### DIFF
--- a/lm75bd/lm75bd.h
+++ b/lm75bd/lm75bd.h
@@ -5,7 +5,7 @@
 #include <stdint.h>
 
 /* LM75BD I2C Device Address */
-#define LM75BD_OBC_I2C_ADDR 0b1001111U /* Define the address here */
+#define LM75BD_OBC_I2C_ADDR 0x4FU /* Define the address here */
 
 /* LM75BD Configuration Values */
 #define LM75BD_DEV_OP_MODE_NORMAL 0x00U
--- a/lm75bd/lm75bd.h
+++ b/lm75bd/lm75bd.h
@@ -5,7 +5,7 @@
 #include <stdint.h>
 
 /* LM75BD I2C Device Address */
-#define LM75BD_OBC_I2C_ADDR /* Define the address here */
+#define LM75BD_OBC_I2C_ADDR 0b1001111U /* Define the address here */
 
 /* LM75BD Configuration Values */
 #define LM75BD_DEV_OP_MODE_NORMAL 0x00U
--- a/services/thermal_mgr/thermal_mgr.c
+++ b/services/thermal_mgr/thermal_mgr.c
@@ -2,6 +2,7 @@
 #include "errors.h"
 #include "lm75bd.h"
 #include "console.h"
+#include "logging.h"
 
 #include <FreeRTOS.h>
 #include <os_task.h>
@@ -44,9 +45,14 @@ void initThermalSystemManager(lm75bd_config_t *config) {
 error_code_t thermalMgrSendEvent(thermal_mgr_event_t *event) {
   /* Send an event to the thermal manager queue */
   if(!event){
-    ERR_CODE_INVALID_ARG;
+    return ERR_CODE_INVALID_ARG;
   }
-  xQueueSend(thermalMgrQueueHandle, event, 0);
+  if(!thermalMgrQueueHandle){
+    return ERR_CODE_INVALID_STATE;
+  }
+  if(xQueueSend(thermalMgrQueueHandle, event, 0) != pdTRUE){
+    return ERR_CODE_QUEUE_FULL;
+  }
   return ERR_CODE_SUCCESS;
 }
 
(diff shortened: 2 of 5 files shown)


lm75bd/lm75bd.h
```diff
@@ -5,7 +5,7 @@
 #include <stdint.h>
 
 /* LM75BD I2C Device Address */
-#define LM75BD_OBC_I2C_ADDR /* Define the address here */
+#define LM75BD_OBC_I2C_ADDR 0b1001111U /* Define the address here */
```
kepler452b123: define this in hex, not binary. usually we define numeric values in hex because binary is hard to read, especially when we are working with memory addresses which are often 4+ bytes long

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +43,39 @@ void initThermalSystemManager(lm75bd_config_t *config) {
 
 error_code_t thermalMgrSendEvent(thermal_mgr_event_t *event) {
   /* Send an event to the thermal manager queue */
-
+  if(!event){
+    ERR_CODE_INVALID_ARG;
```
kepler452b123: should return this err code

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +43,39 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+  }
+  xQueueSend(thermalMgrQueueHandle, event, 0);
```
kepler452b123: check if the queue is null, pass invalid state error code if it is

kepler452b123: also check if this call fails. if it does (does not return pdPass). if it does fail, you need to return that the queue is full thru an error code

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +43,39 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+  handle.type = OS_HANDLER_EVENT;
+  thermalMgrSendEvent(&handle);
 }
 
 static void thermalMgr(void *pvParameters) {
   /* Implement this task */
   while (1) {
-    
+    thermal_mgr_event_t cur_event;
+    if(xQueueReceive(thermalMgrQueueHandle, &cur_event, 0) == pdTRUE){
```
kepler452b123: set the timeout for this to portmax delay. we want the queue to block indefinitely until an event is received

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +43,39 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+      if(cur_event.type == THERMAL_MGR_EVENT_MEASURE_TEMP_CMD){
+        float temp;
+        readTempLM75BD(LM75BD_OBC_I2C_ADDR ,&temp);
```
kepler452b123: you should check if the driver function actually works. if it doesn't you need to log the error code

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +43,39 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+        addTemperatureTelemetry(temp);
+      }else if(cur_event.type == OS_HANDLER_EVENT){
+        float temp;
+        readTempLM75BD(LM75BD_OBC_I2C_ADDR ,&temp);
```
kepler452b123: same as above

kepler452b123 (changes requested):
overall looks really good. just some conventional changes + error logging

services/thermal_mgr/thermal_mgr.c
FxJYg: got it

services/thermal_mgr/thermal_mgr.c
FxJYg: got it

services/thermal_mgr/thermal_mgr.c
FxJYg: got it

services/thermal_mgr/thermal_mgr.c
FxJYg: okay got it!

services/thermal_mgr/thermal_mgr.c
FxJYg: oops

lm75bd/lm75bd.h
FxJYg: got it!

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +44,45 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+  thermalMgrSendEvent(&handle);
 }
 
 static void thermalMgr(void *pvParameters) {
   /* Implement this task */
   while (1) {
-    
+    error_code_t errCode;
+    thermal_mgr_event_t cur_event;
+    if(xQueueReceive(thermalMgrQueueHandle, &cur_event, 1) == pdTRUE){
```
kepler452b123: we want freertos to block indefinitely until an event is received, so use a larger timeout (is there a way to block indefinitely until an event is placed into the queue?)

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +44,45 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+      if(cur_event.type == THERMAL_MGR_EVENT_MEASURE_TEMP_CMD){
+        float temp;
+        LOG_IF_ERROR_CODE(readTempLM75BD(LM75BD_OBC_I2C_ADDR ,&temp));
+        addTemperatureTelemetry(temp);
+      }else if(cur_event.type == OS_HANDLER_EVENT){
+        float temp;
+        LOG_IF_ERROR_CODE(readTempLM75BD(LM75BD_OBC_I2C_ADDR ,&temp));
+        if(temp >= 75){
+          overTemperatureDetected();
```
kepler452b123: consider the case where readTempLM75BD fails. the code would still continue executing addTemperatureTelemetry and the over temperature functions.

kepler452b123 (changes requested):

kepler452b123 (commented):
lgtm send me ur email for notion on discord

PR #214, opened 2025-02-02
Derek Feng onboarding

crankyjoke:
# Purpose

# New Changes
- Explain new changes

# Testing

# Outstanding Changes

### DIFF
--- a/services/thermal_mgr/thermal_mgr.c
+++ b/services/thermal_mgr/thermal_mgr.c
@@ -43,18 +43,39 @@ void initThermalSystemManager(lm75bd_config_t *config) {
 
 error_code_t thermalMgrSendEvent(thermal_mgr_event_t *event) {
   /* Send an event to the thermal manager queue */
-
+  if(event  == NULL || thermalMgrQueueHandle == NULL) return ERR_CODE_INVALID_ARG;
+  xQueueSend(thermalMgrQueueHandle, event, portMAX_DELAY);
   return ERR_CODE_SUCCESS;
 }
 
 void osHandlerLM75BD(void) {
   /* Implement this function */
+
+  thermal_mgr_event_t event; event.type = THERMAL_MGR_EVENT_INTERRUPT;
+  thermalMgrSendEvent(&event);
 }
 
 static void thermalMgr(void *pvParameters) {
-  /* Implement this task */
+  thermal_mgr_event_t event;
+  lm75bd_config_t data;
   while (1) {
-    
+
+    if (xQueueReceive(thermalMgrQueueHandle, &event, portMAX_DELAY)) {
+      data = *(lm75bd_config_t *)pvParameters;
+
+      float temp_c;
+      readTempLM75BD(data.devAddr, &temp_c);
+      if (event.type == THERMAL_MGR_EVENT_MEASURE_TEMP_CMD) {
+        addTemperatureTelemetry(temp_c);
+      } else if (event.type == THERMAL_MGR_EVENT_INTERRUPT) {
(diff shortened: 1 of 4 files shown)


crankyjoke:
Derek Feng onboarding

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +43,39 @@ void initThermalSystemManager(lm75bd_config_t *config) {
 
 error_code_t thermalMgrSendEvent(thermal_mgr_event_t *event) {
   /* Send an event to the thermal manager queue */
-
+  if(event  == NULL || thermalMgrQueueHandle == NULL) return ERR_CODE_INVALID_ARG;
```
kepler452b123: should return error code invalid state if thermalMgrQueueHandle is null

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +43,39 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+  xQueueSend(thermalMgrQueueHandle, event, portMAX_DELAY);
```
kepler452b123: check if xQueueSend returns pd pass. if it doesn't you need to return that the queue is full

kepler452b123: also consider the implications if you put the timeout as portMAX_DELAY and the queue is full for a long time

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +43,39 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+  thermal_mgr_event_t event;
+  lm75bd_config_t data;
   while (1) {
-    
+
+    if (xQueueReceive(thermalMgrQueueHandle, &event, portMAX_DELAY)) {
+      data = *(lm75bd_config_t *)pvParameters;
+
+      float temp_c;
+      readTempLM75BD(data.devAddr, &temp_c);
```
kepler452b123: you need to alter this part to check if readTempLM75BD is returns success. if it doesn't then you need to log the error code. if it does then continue with the logic for the corresponding event type

kepler452b123 (changes requested):

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +44,46 @@ void initThermalSystemManager(lm75bd_config_t *config) {
 
 error_code_t thermalMgrSendEvent(thermal_mgr_event_t *event) {
   /* Send an event to the thermal manager queue */
-
+  if(event  == NULL || thermalMgrQueueHandle == NULL) return ERR_CODE_INVALID_STATE;
```
kepler452b123: If event is null we want to return invalid argument. invalid state is for the queue being null

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +44,46 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+  if(xQueueSend(thermalMgrQueueHandle, event, portMAX_DELAY) != pdPASS) return ERR_CODE_QUEUE_FULL;
```
kepler452b123: the timeout should be 0. portMAX_DELAY means it will wait indefinitely for the queue to have space. since this might be called from an ISR, we want all the actions we do to be really quick

kepler452b123 (changes requested):
just 2 more quick changes

Adityya-K:
Closed as member has already started to contribute (seems to be approved in-person)

PR #216, opened 2025-02-09
Hi

keUeki:
# Purpose

# New Changes
- Explain new changes

# Testing

# Outstanding Changes


lm75bd/lm75bd.c
```diff
@@ -26,8 +26,21 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
 }
 
 error_code_t readTempLM75BD(uint8_t devAddr, float *temp) {
```
kepler452b123, line 28: check if temp is null since its a pointer argument

lm75bd/lm75bd.c
```diff
@@ -26,8 +26,21 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
...
-  /* Implement this driver function */
-  
+  uint8_t reqRegister = 0x00;
+  uint8_t retRegister[2] = {};
+  int16_t preTemp;
+  i2cSendTo(devAddr, &reqRegister, 1);
+  i2cReceiveFrom(devAddr, retRegister, 2);
```
kepler452b123: you should be checking if these return error code success

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +43,33 @@ void initThermalSystemManager(lm75bd_config_t *config) {
 
 error_code_t thermalMgrSendEvent(thermal_mgr_event_t *event) {
   /* Send an event to the thermal manager queue */
-
+  xQueueSend(thermalMgrQueueHandle, event, 0);
```
kepler452b123: error checks:
- Check if event is null, return invalid arg if it is
- Check if the queue is null, return invalid state if it is
- Ensure that the result of xQueueSend is pdpass. if it isnt then the queue is full

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +43,33 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
 }
 
 void osHandlerLM75BD(void) {
   /* Implement this function */
+  float temp = 0.0;
+  readTempLM75BD(LM75BD_OBC_I2C_ADDR, &temp);
+  if (temp >= 80.0) {
+    overTemperatureDetected();
+} else if (temp <= 75.0) {
+    safeOperatingConditions();
```
kepler452b123: we shouldn't be using driver functions directly from the ISR. just use your send function to send an event to the queue. you might need to make a new event type in thermal_mr.h. then in thermal manager check the event type and if it is from an ISR do the overTemperature handling

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +43,33 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
   /* Implement this task */
+  lm75bd_config_t config = *(lm75bd_config_t *)pvParameters;
+  thermal_mgr_event_t event;
+  float temp = 0.0;
   while (1) {
-    
+    if(xQueueReceive(thermalMgrQueueHandle, &event, 1000000) == pdPASS){
+      if(event.type == THERMAL_MGR_EVENT_MEASURE_TEMP_CMD){
+        readTempLM75BD(config.devAddr, &temp);
+        addTemperatureTelemetry(temp);
```
kepler452b123: see comment on osHandler function. good thinking on the queue receive timeout ticks, it should be high. just use portMax delay tho since its a macro and is basically saying to indefinitely block (which is what we want in this function)

kepler452b123 (dismissed):
good start, need to make some changes

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +43,47 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+    if(xQueueReceive(thermalMgrQueueHandle, &event, portMAX_DELAY) == pdPASS){
+      if(event.type == THERMAL_MGR_EVENT_MEASURE_TEMP_CMD){
+        if (readTempLM75BD(config.devAddr, &temp) == ERR_CODE_SUCCESS) {
+          addTemperatureTelemetry(temp);
+        }
+        if(event.type == THERMAL_MGR_EVENT_OVER_TEMPERATURE){
+          if (temp >= 80.0) {
+            overTemperatureDetected();
+          } else if (temp <= 75.0) {
+            safeOperatingConditions();
```
kepler452b123: if the error codes are not success you should be logging the codes for the driver function. also you have the over temperature event enclosed within the measure temp event. separate them and call the driver function for each event type. if it is not successful, log it. if it is, then do the corresponding actions (addTempTelemetry, overTempDetected, etc.)

kepler452b123 (changes requested):

kepler452b123 (approved):
lgtm send your notion email on discord

PR #223, opened 2025-03-28
Title: Completed level1 challenge - Andrew Wang

AndrewWang001:
Implemented solutions to q1–q13 in challenge.c.

### DIFF
--- a/challenge.c
+++ b/challenge.c
@@ -8,7 +8,7 @@
 // Question 0
 // Include the challenge.h header file
 //-------------------------------------------------------------------------
-
+#include "challenge.h"
 
 //-------------------------------------------------------------------------
 // Question 1
@@ -16,7 +16,8 @@
 // respectively. The value of `q1A` should be initialized to 0 and the value 
 // of `q1B`should be initialized to 1.
 //-------------------------------------------------------------------------
-
+int q1A = 0;
+int q1B = 1; //globally init
 
 //-------------------------------------------------------------------------
 // Question 2
@@ -25,6 +26,8 @@
 // named `Q2_ARRAY_SIZE`.
 //-------------------------------------------------------------------------
 
+#define Q2_ARRAY_SIZE 10 //macro define
+int q2Array[Q2_ARRAY_SIZE];
 
 //-------------------------------------------------------------------------
 // Question 3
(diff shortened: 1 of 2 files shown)


challenge.c
```diff
@@ -143,7 +185,18 @@ typedef struct {
 } q11_b_t;
 
 error_t q11(q11_a_t *a, q11_b_t *b){
-
+    if (a == NULL || b == NULL) return FAIL;
+    memcpy(&b->array[1], a->array, sizeof(uint16_t) * 50);
```
Yarik-Popov: Use `b->array + 1` and take the size of the array instead of doing math to find it

challenge.c
```diff
@@ -162,7 +217,13 @@ error_t q11(q11_a_t *a, q11_b_t *b){
 //-------------------------------------------------------------------------
 
 void *q13(uint32_t *ptr1, uint16_t *ptr2){
+    if (ptr1 == NULL || ptr2 == NULL) return (void *)-1; //edge case if one DNE
+
+    uintptr_t p1 = (uintptr_t)ptr1 +5; //unsigned int large enough to store pointer, inc by 5
+    uintptr_t p2 = (uintptr_t)ptr2 +5;
```
Yarik-Popov: `uintptr_t` has size 4 you want to use `uint8_t *` has size 1

challenge.c
```diff
@@ -162,7 +217,13 @@ error_t q11(q11_a_t *a, q11_b_t *b){
...
 
+    return (void *)((p1 < p2) ? p1 : p2); //return the smallest inc by 5
```
Yarik-Popov: You should use your `MIN` macro above

Yarik-Popov (changes requested):
Hopefully this fixes your issue

challenge.c
```diff
@@ -68,6 +68,9 @@ int32_t q4(uint8_t * array, uint32_t arrayLength) {
 
     int32_t sum = 0; //moved outside to be returned
     for (uint8_t i = 0; i < arrayLength; i++) { //only less than as terminating condition
+        if (!array[i]) {
+            break; //if array[i] DNE
+        }
```
Yarik-Popov: This will break out if the array has value 0 in a given spot which is not the intended behaviour. `array = {0, 1}` has sum 1 but will return 0

Yarik-Popov (changes requested):
This has bug if `array = {0, 1}`

Adityya-K:
Member has already contributed (The approval for this onboarding must have been in-person)

PR #224, opened 2025-05-04
Adityya Kaushal Onboarding Task

Adityya-K:
# Purpose
Complete the Onboarding Task for UW Orbital

# New Changes
Implemented all tasks with comments and some level of error handling


lm75bd/lm75bd.c
```diff
@@ -26,14 +29,49 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
 }
 
 error_code_t readTempLM75BD(uint8_t devAddr, float *temp) {
-  /* Implement this driver function */
-  
+  // Define errCode for error macros
+  error_code_t errCode;
+
```
Navtajh04, line 38: Check to make sure temp is not NULL

lm75bd/lm75bd.c
```diff
@@ -1,22 +1,25 @@
 #include "lm75bd.h"
```
Navtajh04, line 1: Add the unneeded files above to gitignore

lm75bd/lm75bd.c
```diff
@@ -26,14 +29,49 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
...
+  // temperature sensor
+  RETURN_IF_ERROR_CODE(i2cSendTo(devAddr, &tempReg, 1));
+  // Recieve data from the temperature sensor
+  RETURN_IF_ERROR_CODE(i2cReceiveFrom(devAddr, readBuff, 2));
+
+  // Start converting by taking the array, converting it to a large 16 bit
+  // signed integer and deleting off the last five bits as they are useless
+  int16_t tempVal = (int16_t)((readBuff[0] << 8) | readBuff[1]) >> 5;
+
+  printf("Val: %d", tempVal);
```
Navtajh04: printf won't work do anything on our system (std out is not defined).

In general we want to only log if there are errors and if there are any variables we want to log (like temperature of a specific board in this case) then we add it as telemetry data. This also saves us a lot of memory on our error logging.

For the onboarding though I wouldn't worry about logging temp

lm75bd/lm75bd.c
```diff
@@ -26,14 +29,49 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
...
+    // floats. So we take the original unsigned binary representation by using
+    // two's complement again. We convert the unsigned value to a float and then
+    // multiply by the required constant and negative which makes the float
+    // encoding correct!
+    *temp = (float)(~(tempVal) + 1) * -0.125;
+  } else {
+    // If the number was positive to begin with, then it is already represented
+    // as an unsigned and no coverting is needed.
+    *temp = (float)(tempVal) * 0.125;
+  }
```
Navtajh04: since `int16_t` is already in 2s compliment format, you should just be able to do `*temp = ((float)(tempVal)) * 0.125

lm75bd/lm75bd.h
```diff
@@ -5,7 +5,7 @@
 #include <stdint.h>
 
 /* LM75BD I2C Device Address */
-#define LM75BD_OBC_I2C_ADDR /* Define the address here */
+#define LM75BD_OBC_I2C_ADDR 0x04F
```
Navtajh04: nit: `0x4F` here would be a bit more clear here since I2C address are 7 bits so the extra 0 is kinda confusing

services/thermal_mgr/thermal_mgr.c
```diff
@@ -20,41 +23,104 @@ static StackType_t thermalMgrTaskStack[THERMAL_MGR_STACK_SIZE];
...
-    THERMAL_MGR_QUEUE_LENGTH, THERMAL_MGR_QUEUE_ITEM_SIZE,
-    thermalMgrQueueStorageArea, &thermalMgrQueueBuffer);
-
+  thermalMgrQueueHandle =
+      xQueueCreateStatic(THERMAL_MGR_QUEUE_LENGTH, THERMAL_MGR_QUEUE_ITEM_SIZE,
+                         thermalMgrQueueStorageArea, &thermalMgrQueueBuffer);
 }
 
 error_code_t thermalMgrSendEvent(thermal_mgr_event_t *event) {
   /* Send an event to the thermal manager queue */
```
Navtajh04, line 57: make sure queue has been created (handle is not NULL)

services/thermal_mgr/thermal_mgr.c
```diff
@@ -20,41 +23,104 @@ static StackType_t thermalMgrTaskStack[THERMAL_MGR_STACK_SIZE];
...
+  if (xQueueSend(thermalMgrQueueHandle, event, 0) == pdPASS) {
+    return ERR_CODE_SUCCESS;
+    // We check if the queue is full and return accordingly
+  } else if (xQueueIsQueueFullFromISR(thermalMgrQueueHandle) == pdFALSE) {
+    return ERR_CODE_QUEUE_FULL;
+    // If none of the previous cases occured, then we do not know the error and
+    // we return an ERR_CODE_UNKNOWN
+  } else {
+    return ERR_CODE_UNKNOWN;
+  }
```
Navtajh04: `xQueueSend` only returns fail if the queue is full so you can return that error code if it fails without needing the else if

Also adding that else if causes an issue where if the queue was full when we ran the if clause but emptied out by the time we ran the else if we will get `ERR_CODE_UNKNOWN` even though the failure was actually because the queue was full when we tried to send. I'd reccomend looking at RTOS 101 to understand better what I mean by race condition

services/thermal_mgr/thermal_mgr.c
```diff
@@ -20,41 +23,104 @@ static StackType_t thermalMgrTaskStack[THERMAL_MGR_STACK_SIZE];
...
+  error_code_t errCode;
+
+  // We create a new type of event to send into the queue since we want to
+  // minimize the execution time of this function, so let's offload the work to
+  // our thermal manager function
+  thermal_mgr_event_t osInterrupt;
+
+  // From the header we define a new type of event to look for in our thermal
+  // manager
+  osInterrupt.type = THERMAL_MGR_EVENT_HYS_TEMP;
```
Navtajh04: nit: `thermal_mgr_event_t osInterrupt = {.type = THERMAL_MGR_EVENT_HYS_TEMP};` is a bit cleaner

services/thermal_mgr/thermal_mgr.c
```diff
@@ -20,41 +23,104 @@ static StackType_t thermalMgrTaskStack[THERMAL_MGR_STACK_SIZE];
...
+
+        // Check for overtemperature and call the appropriate function
+        if (temp > 80) {
+          overTemperatureDetected();
+
+          // Check for hysteresis and call the appropriate function
+        } else if (temp < 75) {
+          safeOperatingConditions();
+        }
+      }
```
Navtajh04, line 124: return error code if event type was not either of the above

services/thermal_mgr/thermal_mgr.c
Navtajh04: Don't thinik this was handled

services/thermal_mgr/thermal_mgr.c
Navtajh04: ^

PR #230, opened 2025-06-30
Onboarding Rakshit Nair 

Rak256:
# Purpose
Finished the onboarding challenge automated tests.

# New Changes
- Implemented the readTempLM75BD(), Thermal Management Task, and the OS functions

# Testing
- Finished the automated tests provided in the Notion docs

# Outstanding Changes
- N/A

### DIFF
--- a/lm75bd/lm75bd.c
+++ b/lm75bd/lm75bd.c
@@ -7,6 +7,7 @@
 #include <string.h>
 #include <math.h>
 
+static uint8_t buffer[2] = {0x0U};
 /* LM75BD Registers (p.8) */
 #define LM75BD_REG_CONF 0x01U  /* Configuration Register (R/W) */
 
@@ -27,7 +28,28 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
 
 error_code_t readTempLM75BD(uint8_t devAddr, float *temp) {
   /* Implement this driver function */
+  //Send the pointer byte
+  i2cSendTo(devAddr, buffer, 1);
+  //Receive temp register data in the buffer
+  i2cReceiveFrom(devAddr, buffer,2);
   
+  //Shift MSB by 3 since last 5 bits are ignored. Shift LSB by 5 bits to ignore last 5 bits
+  uint16_t tmp_value = (buffer[0] << 3) + (buffer[1] >> 5); 
+
+  if ((buffer[0] & 0x80) != 0){
+    //2's complement calculation. Flip all 11 bits and then add 1
+    tmp_value ^= 0x7FF;
+    tmp_value += 0x1U;
+    *temp = -tmp_value*0.125;
+  }
+
+  else{
+    *temp = tmp_value * 0.125;
+  }
+
+  //reinitialize buffer
+  buffer[0] = 0x0U;
+  buffer[1] = 0x0U;
   return ERR_CODE_SUCCESS;
 }
 
--- a/lm75bd/lm75bd.c
+++ b/lm75bd/lm75bd.c
@@ -7,7 +7,7 @@
 #include <string.h>
 #include <math.h>
 
-static uint8_t buffer[2] = {0x0U};
+
 /* LM75BD Registers (p.8) */
 #define LM75BD_REG_CONF 0x01U  /* Configuration Register (R/W) */
 
@@ -27,11 +27,18 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
 }
 
 error_code_t readTempLM75BD(uint8_t devAddr, float *temp) {
+  if (temp == NULL){
+    return ERR_CODE_INVALID_ARG;
+  }
   /* Implement this driver function */
+  uint8_t buffer[2] = {0x0U};
+
+  error_code_t errCode;
+
   //Send the pointer byte
-  i2cSendTo(devAddr, buffer, 1);
+ LOG_IF_ERROR_CODE(i2cSendTo(devAddr, buffer, 1));
   //Receive temp register data in the buffer
-  i2cReceiveFrom(devAddr, buffer,2);
+ LOG_IF_ERROR_CODE(i2cReceiveFrom(devAddr, buffer,2));
   
   //Shift MSB by 3 since last 5 bits are ignored. Shift LSB by 5 bits to ignore last 5 bits
   uint16_t tmp_value = (buffer[0] << 3) + (buffer[1] >> 5); 
@@ -47,9 +54,6 @@ error_code_t readTempLM75BD(uint8_t devAddr, float *temp) {
     *temp = tmp_value * 0.125;
   }
 
-  //reinitialize buffer
-  buffer[0] = 0x0U;
-  buffer[1] = 0x0U;
   return ERR_CODE_SUCCESS;
 }
 
--- a/lm75bd/lm75bd.c
+++ b/lm75bd/lm75bd.c
@@ -35,10 +35,10 @@ error_code_t readTempLM75BD(uint8_t devAddr, float *temp) {
 
   error_code_t errCode;
 
-  //Send the pointer byte
- LOG_IF_ERROR_CODE(i2cSendTo(devAddr, buffer, 1));
-  //Receive temp register data in the buffer
- LOG_IF_ERROR_CODE(i2cReceiveFrom(devAddr, buffer,2));
+  //Send the pointer byte, break operation if an error is detected
+ RETURN_IF_ERROR_CODE(i2cSendTo(devAddr, buffer, 1));
+  //Receive temp register data in the buffer, break operation if an error is detected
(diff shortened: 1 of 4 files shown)


lm75bd/lm75bd.c
```diff
@@ -7,6 +7,7 @@
 #include <string.h>
 #include <math.h>
 
+static uint8_t buffer[2] = {0x0U};
```
kepler452b123: this does not need to be a static variable

lm75bd/lm75bd.c
```diff
@@ -27,7 +28,28 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
 
 error_code_t readTempLM75BD(uint8_t devAddr, float *temp) {
   /* Implement this driver function */
```
kepler452b123, line 33: check if temp is a null ptr (always do this in functions where you dereference pointers)

lm75bd/lm75bd.c
```diff
@@ -27,7 +28,28 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
...
+  //Send the pointer byte
+  i2cSendTo(devAddr, buffer, 1);
+  //Receive temp register data in the buffer
+  i2cReceiveFrom(devAddr, buffer,2);
```
kepler452b123: use the LOG_IF_ERR_CODE macro when calling these

lm75bd/lm75bd.c
```diff
@@ -27,7 +28,28 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
...
+    *temp = -tmp_value*0.125;
+  }
+
+  else{
+    *temp = tmp_value * 0.125;
+  }
+
+  //reinitialize buffer
+  buffer[0] = 0x0U;
+  buffer[1] = 0x0U;
```
kepler452b123: this is not needed if buffer is not a static variable

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +57,56 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
   /* Implement this function */
+  //Check if OS threshold was crossed
+  if (temp > LM75BD_DEFAULT_OT_THRESH){
+    thermalMgrSendEvent(&os);
+  }
+
+  //Check if hysteresis threshold was crossed
+  else if (temp < LM75BD_DEFAULT_HYST_THRESH ){
+    thermalMgrSendEvent(&safe);
+  }
```
kepler452b123: temp does not have a value here, don't make temp static. thermalMgr has a while(1) loop so you can put temp there and check for overtemp there.

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +57,56 @@ void initThermalSystemManager(lm75bd_config_t *config) {
 
 error_code_t thermalMgrSendEvent(thermal_mgr_event_t *event) {
   /* Send an event to the thermal manager queue */
+  if (xQueueSend(thermalMgrQueueHandle, event, (TickType_t) 0) == pdPASS){
+    if (event == THERMAL_MGR_EVENT_MEASURE_TEMP_CMD || event == THERMAL_MGR_EVENT_OS || event == THERMAL_MGR_EVENT_SAFE_OP){
+    return ERR_CODE_SUCCESS;
+    }
```
kepler452b123: you should check this before sending the event to the queue

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +57,56 @@ void initThermalSystemManager(lm75bd_config_t *config) {
 
 error_code_t thermalMgrSendEvent(thermal_mgr_event_t *event) {
   /* Send an event to the thermal manager queue */
+  if (xQueueSend(thermalMgrQueueHandle, event, (TickType_t) 0) == pdPASS){
```
kepler452b123: ticktime is good

services/thermal_mgr/thermal_mgr.c
```diff
@@ -22,6 +22,20 @@ static QueueHandle_t thermalMgrQueueHandle;
...
+static float temp;
+
+//OS event
+static thermal_mgr_event_t os ={
+  THERMAL_MGR_EVENT_OS
+};
+//Safe operations event
+static thermal_mgr_event_t safe = {
+  THERMAL_MGR_EVENT_SAFE_OP
+};
```
kepler452b123: events should be in the header file

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +57,56 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+  }
+
 }
 
 static void thermalMgr(void *pvParameters) {
   /* Implement this task */
   while (1) {
-    
+    //receive event from queue
+    if(xQueueReceive(thermalMgrQueueHandle, &messageBuffer, (TickType_t) 0) == pdPASS){
```
kepler452b123: making ticktype=0 does not leverage FreeRTOS properly. you can use a longer time and the task will block until an event is in the queue. this introduces overhead as it will repeatedly try xQueueReceive

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +57,56 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+      //update temp variable/reset OS 
+      readTempLM75BD(LM75BD_OBC_I2C_ADDR, &temp);
```
kepler452b123: LOG_IF_ERR_CODE for the driver function

kepler452b123 (changes requested):
good track/ideas. needs some cleanup though

services/thermal_mgr/thermal_mgr.c
Rak256: Just to make sure I understand the error, my current code keeps polling for the queue even though an event may not be in it for a long time. So, I should block the task for some arbitrary of time (like port_max_delay?) until the event occurs so that clock cycles aren't wasted. Is this correct?

Does this not apply to the queue send function because you can always send to the queue unless it is full?

lm75bd/lm75bd.c
```diff
@@ -26,8 +27,33 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
...
+  }
   /* Implement this driver function */
+  uint8_t buffer[2] = {0x0U};
+
+  error_code_t errCode;
+
+  //Send the pointer byte
+ LOG_IF_ERROR_CODE(i2cSendTo(devAddr, buffer, 1));
+  //Receive temp register data in the buffer
+ LOG_IF_ERROR_CODE(i2cReceiveFrom(devAddr, buffer,2));
```
kepler452b123: should use RETURN_IF_ERROR_CODE here, as otherwise the function will continue

services/thermal_mgr/thermal_mgr.c
```diff
@@ -42,19 +45,56 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+    //If queue is full
+    return ERR_CODE_QUEUE_FULL;
+  }
+
+  
 }
 
 void osHandlerLM75BD(void) {
   /* Implement this function */
+
```
kepler452b123: you can initiate an instance of the thermal_mgr_event_type_t here and pass that

services/thermal_mgr/thermal_mgr.c
```diff
@@ -42,19 +45,56 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
   /* Implement this task */
   while (1) {
-    
+    // variable to store temperature
+    float temp;
+    //receive event from queue
+    if(xQueueReceive(thermalMgrQueueHandle, &messageBuffer, portMAX_DELAY) == pdPASS){
+      error_code_t errCode;
+      //update temp variable/reset OS 
+      LOG_IF_ERROR_CODE(readTempLM75BD(LM75BD_OBC_I2C_ADDR, &temp));
```
kepler452b123, line 85: this would still continue the loop even if it fails (use continue or make sure errCode is success or something to skip the rest of the logic if this fails)

services/thermal_mgr/thermal_mgr.h
```diff
@@ -16,6 +17,10 @@ typedef struct {
 extern "C" {
 #endif
 
+static thermal_mgr_event_t os = {
+  THERMAL_MGR_EVENT_OS
+};
```
kepler452b123: don't need to make this a static variable, make a local variable in the osHandler

kepler452b123 (changes requested):
pretty much done, just a few things

kepler452b123 (commented):
lgtm, send email for notion and github to me on disc

PR #231, opened 2025-07-07
Onboarding Jack Dietrich

Jack-Dietrich:
# Purpose
Finished the onboarding challenge, looking for feedback.

# New Changes
Implemented the readTempLM75BD, Thermal Management Task, and osHandlerLM75BD

# Testing
Unit tests and Integration test ran.

# Outstanding Changes
N/A

### DIFF
--- a/lm75bd/lm75bd.c
+++ b/lm75bd/lm75bd.c
@@ -25,9 +25,51 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
   return ERR_CODE_SUCCESS;
 }
 
+#define CONF_WRITE_BUFF_SIZE 1U //should need to be a maximum of 2 as for setup we write dev addr then ptr byte, rest of the time, just write dev addr and get back 2 BYTES
+#define CONF_READ_BUFF_SIZE 2U
+#define LM75BD_PTR_BYTE 0x00U //should just be 0 as we want to read from the temp register. Thus this technically is not neccisary
+#define LM75BD_PTR_SEND 2U
+#define LM75BD_TEMP_READ 1U
 error_code_t readTempLM75BD(uint8_t devAddr, float *temp) {
   /* Implement this driver function */
   
+  /*Plan
+  -As per fig 11, we need to write the device addr then the poiner byte to indicate we are reading temp
+  -need to throw away first temp reading
+  -
+  
+  
+  */
+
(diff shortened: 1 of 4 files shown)


lm75bd/lm75bd.c
```diff
@@ -25,9 +25,60 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
...
+  readBuff[1] = 0x00;//Just 0 as we want to read
+
+  writeBuff[0] = 0x00;
+
+  LOG_IF_ERROR_CODE(i2cSendTo(LM75BD_OBC_I2C_ADDR, writeBuff, 1));  //Send the pointer byte, the address should already be sent by the function itsself
+
+  
+
+
+  LOG_IF_ERROR_CODE(i2cReceiveFrom(LM75BD_OBC_I2C_ADDR, readBuff, 2));//read in 2 bytes back from the LM75BD
```
kepler452b123: should return here as we don't want the function to continue if they fail

lm75bd/lm75bd.c
```diff
@@ -25,9 +25,60 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
...
+    preTemp++;//add 1 to pretemp
+    preTemp = preTemp >> 5;
+    convTemp = -(preTemp*0.125);//negate it
+  }else{
+    preTemp = preTemp >> 5;
+    convTemp = preTemp*0.125;
+  }
+
+  if(temp){//dereference only if not nullptr
+    *temp = convTemp;
```
kepler452b123: we always check if a pointer param is null at the start of a function (and return invalid_arg if it is)

services/thermal_mgr/thermal_mgr.c
```diff
@@ -44,18 +45,70 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+        if( temp >  LM75BD_DEFAULT_HYST_THRESH){
+
+          //interupt was called because we are over temp
+          overTemperatureDetected();
+
+        }else{//if the interrupt happened and we are not over temp, that means we just returned to regular temps
+
+          safeOperatingConditions();
+          
+        }
```
kepler452b123: good, but implement logic to skip the rest of the logic if the driver function fails. e.g. check if errcode is success before continuing, or use continue statement if errCode is false. we don't want to do the rest of the logic if the driver function fails

kepler452b123 (changes requested):

kepler452b123 (commented):
lgtm send email to me

PR #233, opened 2025-08-10
FW Onboarding task - Arjun

asterbot:
# Purpose
Onboarding task for FW team

# New Changes
Completed tasks mentioned in onboarding notion 

# Testing
Ran unit tests + integration tests

### DIFF
--- a/lm75bd/lm75bd.c
+++ b/lm75bd/lm75bd.c
@@ -25,8 +25,37 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
   return ERR_CODE_SUCCESS;
 }
 
+float convertToCelcius(uint16_t finalTemp){
+    finalTemp &= (1<<11) - 1; // keep lowest 11 bits
+
+    return (finalTemp & (1<<10))? 
+        -1*(~(finalTemp | ~(((1<<16) - 1)>>5)) + 1) * 0.125f :
+        finalTemp * 0.125f;
+}
+
 error_code_t readTempLM75BD(uint8_t devAddr, float *temp) {
   /* Implement this driver function */
+
+  if (!temp) return ERR_CODE_INVALID_ARG;
+
+  error_code_t errCode;
+
+  // Select sensor's internal register using pointer register
+  uint8_t pointerRegister = 0; // 00 for temperature register
+  RETURN_IF_ERROR_CODE(
+    i2cSendTo(devAddr, &pointerRegister, sizeof(pointerRegister));
+  );
+
+  uint8_t temperature[2] = {0, 0};
+  RETURN_IF_ERROR_CODE(
+    i2cReceiveFrom(devAddr, temperature, sizeof(temperature));
+  );
+
+  uint16_t finalTemp = 0;
+  finalTemp += (temperature[0] << 3);
+  finalTemp += (temperature[1] >> 5);
+
+  *temp = convertToCelcius(finalTemp);
   
   return ERR_CODE_SUCCESS;
 }
--- a/lm75bd/lm75bd.c
+++ b/lm75bd/lm75bd.c
@@ -46,6 +46,8 @@ error_code_t readTempLM75BD(uint8_t devAddr, float *temp) {
     i2cSendTo(devAddr, &pointerRegister, sizeof(pointerRegister));
   );
 
+  // temperature[0] stores MSByte
+  // temperature[1] stores LSByte
   uint8_t temperature[2] = {0, 0};
   RETURN_IF_ERROR_CODE(
     i2cReceiveFrom(devAddr, temperature, sizeof(temperature));
--- a/lm75bd/lm75bd.c
+++ b/lm75bd/lm75bd.c
@@ -25,16 +25,18 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
   return ERR_CODE_SUCCESS;
 }
 
-float convertToCelcius(int16_t finalTemp){
-    return (finalTemp & (1<<10))? 
-        -1*(-finalTemp) * 0.125f :
-        finalTemp * 0.125f;
+float convertToCelcius(int16_t tempData){
+  if (tempData & (1<<10)){           // if D10=1, make it -(2's complement of temp)
+      tempData |= (1<<16) - (1<<11); // make top 5 MSB's 1
+      tempData = -1*(-tempData);
+  }
+  return tempData * 0.125f;
 }
 
 error_code_t readTempLM75BD(uint8_t devAddr, float *temp) {
   /* Implement this driver function */
 
-  if (!temp) return ERR_CODE_INVALID_ARG;
+  if (temp == NULL) return ERR_CODE_INVALID_ARG;
 
   error_code_t errCode;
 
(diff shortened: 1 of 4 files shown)


services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +44,45 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+          addTemperatureTelemetry(tempC);
+      }
+      if (tme.type == THERMAL_MGR_EVENT_OS_HANDLER){
+          if (tempC > config.overTempThresholdCelsius){
+            overTemperatureDetected();
+          }
+          else if (tempC < config.hysteresisThresholdCelsius) {
+            safeOperatingConditions();
+          }
+      }
```
Adityya-K, line 99: This should print or log something if the neither of the event above were received. It's good practice and very crucial to debugging.

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +44,45 @@ void initThermalSystemManager(lm75bd_config_t *config) {
 
 error_code_t thermalMgrSendEvent(thermal_mgr_event_t *event) {
   /* Send an event to the thermal manager queue */
+  if (!event) return ERR_CODE_INVALID_ARG;
```
Adityya-K, line 48: Good work adding the NULL check! We just need to add one more check that checks `thermalMgrQueueHandle`. Again good practice and better to be safe than sorry.

lm75bd/lm75bd.c
```diff
@@ -25,8 +25,39 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
   return ERR_CODE_SUCCESS;
 }
 
+float convertToCelcius(uint16_t finalTemp){
+    finalTemp &= (1<<11) - 1; // keep lowest 11 bits
+
+    return (finalTemp & (1<<10))? 
+        -1*(~(finalTemp | ~(((1<<16) - 1)>>5)) + 1) * 0.125f :
+        finalTemp * 0.125f;
```
Adityya-K: You can actually simplify this a lot if you use C's signed int, `int16_t`. If we cast the bits to an `int16_t` they will be stored as 2's complement (effectively undoing the 2's complement from the temperature register). Of course, relying on `int16_t` being 2's complement in the cross-platform project wouldn't be a great idea, but, here we can be certain about it since we are only targeting one platform.

lm75bd/lm75bd.c
```diff
@@ -25,8 +25,39 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
...
+  );
+
+  // temperature[0] stores MSByte
+  // temperature[1] stores LSByte
+  uint8_t temperature[2] = {0, 0};
+  RETURN_IF_ERROR_CODE(
+    i2cReceiveFrom(devAddr, temperature, sizeof(temperature));
+  );
+
+  uint16_t finalTemp = 0;
```
Adityya-K: Again instead of storing this as a `uint16_t`, you can just cast it to an `int16_t` and store it as such (refer to previous comment).

lm75bd/lm75bd.c
```diff
@@ -25,8 +25,39 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
...
+
+    return (finalTemp & (1<<10))? 
+        -1*(~(finalTemp | ~(((1<<16) - 1)>>5)) + 1) * 0.125f :
+        finalTemp * 0.125f;
+}
+
 error_code_t readTempLM75BD(uint8_t devAddr, float *temp) {
   /* Implement this driver function */
+
+  if (!temp) return ERR_CODE_INVALID_ARG;
```
Adityya-K: To follow the codebase style I think you should explicitly check against NULL. It just improves readability and allows you to check against a macro that is used throughout to check for NULL.

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +44,45 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+  lm75bd_config_t config = *(lm75bd_config_t *)pvParameters;
   while (1) {
-    
+    thermal_mgr_event_t tme;
+    if (xQueueReceive(thermalMgrQueueHandle, &tme, (TickType_t) 0 ) == pdPASS){
+      float tempC = 0.0f;
+      error_code_t errCode;
+      LOG_IF_ERROR_CODE(
+        readTempLM75BD(config.devAddr, &tempC);
+      );
```
Adityya-K, line 77: This would continue through the loop even of the `readTempLM75BD()` function fails. Let's try checking for it's success and then continuing through the loop, otherwise, we should skip to the next iteration.

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +44,45 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+  thermalMgrSendEvent(&event); 
 }
 
 static void thermalMgr(void *pvParameters) {
   /* Implement this task */
+  lm75bd_config_t config = *(lm75bd_config_t *)pvParameters;
   while (1) {
-    
+    thermal_mgr_event_t tme;
+    if (xQueueReceive(thermalMgrQueueHandle, &tme, (TickType_t) 0 ) == pdPASS){
```
Adityya-K: The delay should be indefinite here as we want this to block until we receive an event.

Adityya-K (commented):
Great start! There are just some cases you need handle, some style nit-piks and possible simplifications you can make.

services/thermal_mgr/thermal_mgr.c
Adityya-K: Forgot to add a comment here but, as I will mention later, you should probably explicitly check against `NULL`.

lm75bd/lm75bd.c
```diff
@@ -48,12 +46,12 @@ error_code_t readTempLM75BD(uint8_t devAddr, float *temp) {
 
   // temperature[0] stores MSByte
   // temperature[1] stores LSByte
-  uint8_t temperature[2] = {0, 0};
+  int8_t temperature[2] = {0, 0};
```
Adityya-K: Try keeping this butter as an unsigned int so that C doesn't add any sort of interpretation to the data.

lm75bd/lm75bd.c
```diff
@@ -48,12 +46,12 @@ error_code_t readTempLM75BD(uint8_t devAddr, float *temp) {
...
   RETURN_IF_ERROR_CODE(
     i2cReceiveFrom(devAddr, temperature, sizeof(temperature));
   );
 
-  uint16_t finalTemp = 0;
+  int16_t finalTemp = 0;
```
Adityya-K: Try using the `<< >>` (bit shift) and `|` (or) operation to combine the first and second elements of the buffer, cast it and then use bit shift to discard the last five bytes. Then once you have the temperature as a `int16_t`, you can cast it to a float and assign it to `temp`.

services/thermal_mgr/thermal_mgr.c
Adityya-K: Remember to add the null check for the `thermalMgrQueueHandle`

Adityya-K (commented):
LGTM!

PR #234, opened 2025-08-27
Finished Onboarding

AnshRoutray:
Implemented the temperature drivers and added the thermal management task.


Testing/Temporary/CTestCostData.txt
```diff
@@ -0,0 +1 @@
+---
```
Adityya-K: Make sure to delete any testing files before pushing to the repo (we do not want any unwanted cluttering the repo!)

services/thermal_mgr/thermal_mgr.c
```diff
@@ -38,23 +38,45 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
-
 }
 
 error_code_t thermalMgrSendEvent(thermal_mgr_event_t *event) {
-  /* Send an event to the thermal manager queue */
-
+  if (event == NULL){
+    return ERR_CODE_INVALID_ARG;
+  }
+  if (xQueueSend(thermalMgrQueueHandle, event, (TickType_t)10) == errQUEUE_FULL){
```
Adityya-K: Instead of casting, you can use a macro within the codebase `pdMS_TO_TICKS()` which will convert milliseconds to the amount of ticks. This is for readability but, also functionality since the macro would have been defined with the specifications of the RTOS in mind.

services/thermal_mgr/thermal_mgr.c
```diff
@@ -38,23 +38,45 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
-  /* Implement this task */
+static void thermalMgr(void *pvParameters){
+  lm75bd_config_t config = *(lm75bd_config_t*)pvParameters;
   while (1) {
-    
+    thermal_mgr_event_t eventBuf;
+    if (xQueueReceive(thermalMgrQueueHandle, &eventBuf, 0) != errQUEUE_EMPTY){
+      if (eventBuf.type == THERMAL_MGR_EVENT_MEASURE_TEMP_CMD){
+        float temp;
+        readTempLM75BD(config.devAddr, &temp);
```
Adityya-K: You should add error checking here are this function can fail.

services/thermal_mgr/thermal_mgr.c
```diff
@@ -38,23 +38,45 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+        addTemperatureTelemetry(temp);
+      }
+      else if (eventBuf.type == THERMAL_MGR_EVENT_OS_INTERRUPT){
+        float temp;
+        readTempLM75BD(config.devAddr, &temp);
```
Adityya-K: Again make sure to handle the error this function might throw!

services/thermal_mgr/thermal_mgr.c
```diff
@@ -38,23 +38,45 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+        if (temp > config.hysteresisThresholdCelsius){
+          overTemperatureDetected();
+        }
+        else {
+          safeOperatingConditions();
+        }
+      }
```
Adityya-K, line 93: You should also account for if the `eventBuf.type` is none of the two conditions you've specified above. Logging an error will suffice!

Adityya-K (changes requested):
Great start! Just a few things to handle before your onboarding is approved.

services/thermal_mgr/thermal_mgr.c
Adityya-K: We should also check if `thermalMgrQueueHandle` is null before using the queue send function.

Adityya-K (commented):
LGTM!

PR #236, opened 2025-08-27
Onboarding Completed

vy038:


services/thermal_mgr/thermal_mgr.c
```diff
@@ -42,19 +42,46 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+  thermalMgrSendEvent(&event);
 }
 
 static void thermalMgr(void *pvParameters) {
-  /* Implement this task */
-  while (1) {
-    
+  lm75bd_config_t config = *(lm75bd_config_t *)pvParameters;
+  while (1) { // constant check
+    thermal_mgr_event_t *event;
```
Adityya-K: This does not need to be a pointer variable, just keep it normal (Remember pointers are just memory addresses but here we want to define an actual variable to store data for us). Be sure to update the rest of the file accordingly!

services/thermal_mgr/thermal_mgr.c
```diff
@@ -42,19 +42,46 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+    if (xQueueReceive(thermalMgrQueueHandle, &event, portMAX_DELAY) == pdPASS) { // upon successful event
+      float tempC;
+      if ((*event).type == THERMAL_MGR_EVENT_MEASURE_TEMP_CMD) {  // if measure temperature command
+        if (readTempLM75BD(LM75BD_OBC_I2C_ADDR, &tempC) == ERR_CODE_SUCCESS) { // if successful temperature value
```
Adityya-K: Let's log if this read is unsuccessful so that the board is easier to debug when things go wrong.

services/thermal_mgr/thermal_mgr.c
```diff
@@ -42,19 +42,46 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+          addTemperatureTelemetry(tempC);
+        }
+      } else {
+        if (readTempLM75BD(LM75BD_OBC_I2C_ADDR, &tempC) == ERR_CODE_SUCCESS) {
```
Adityya-K: Again be sure to log if this goes wrong and remember you can always use the error handling macros defined on the notion doc.

services/thermal_mgr/thermal_mgr.c
```diff
@@ -42,19 +42,46 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+          if (tempC >= config.overTempThresholdCelsius) { // if over temperature threshold
+            overTemperatureDetected();
+          } else if (tempC <= config.hysteresisThresholdCelsius) { // safe temperature threshold
+            safeOperatingConditions();
+          }
+        }
```
Adityya-K, line 91: Let's handle if the event type was none of the above as well by logging an error.

services/thermal_mgr/thermal_mgr.c
```diff
@@ -42,19 +42,46 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
 
 error_code_t thermalMgrSendEvent(thermal_mgr_event_t *event) {
-  /* Send an event to the thermal manager queue */
+  // Send an event to the thermal manager queue 
+
+  // if event is null 
+  if (event == NULL) return ERR_CODE_INVALID_ARG;
+
+  // if queue is full return error
+  if (xQueueSend(thermalMgrQueueHandle, ( void * ) &event, portMAX_DELAY) == errQUEUE_FULL) {
```
Adityya-K: We just need to pass in event here without any extra operation. The queue send function takes a pointer to the item to add to the queue. Since the function parameter, `event`, is a pointer, we can just use it without anything extra.

services/thermal_mgr/thermal_mgr.c
```diff
@@ -42,19 +42,46 @@ void initThermalSystemManager(lm75bd_config_t *config) {
 }
 
 error_code_t thermalMgrSendEvent(thermal_mgr_event_t *event) {
-  /* Send an event to the thermal manager queue */
+  // Send an event to the thermal manager queue 
+
+  // if event is null 
+  if (event == NULL) return ERR_CODE_INVALID_ARG;
```
Adityya-K, line 49: We should also check if the queue handle is none before using the queue send function.

Adityya-K (changes requested):
Great start, we just need to handle some errors. Additionally, make sure you are using pointers correctly (you can always ask if you need help)!

services/thermal_mgr/thermal_mgr.c
```diff
@@ -42,19 +43,58 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+
+      error_code_t readTemp = readTempLM75BD(LM75BD_OBC_I2C_ADDR, &tempC);
+
+      if (readTemp == ERR_CODE_SUCCESS) {
+
+        if (event.type == THERMAL_MGR_EVENT_MEASURE_TEMP_CMD) {  // if measure temperature command
+          addTemperatureTelemetry(tempC);
+        } else {
+
+          if (tempC >= config.overTempThresholdCelsius) { // if over temperature threshold
```
Adityya-K, line 83: Sorry I might have been confusing here. We want to check here if there is an event other than `THERMAL_MGR_EVENT_MEASURE_TEMP` where we execute the `addTemperatureTelemetry()` function and `THERMAL_MGR_EVENT_OS_INTERRUPT` (you should probably use an `else if` to check this) where we check for hysteresis. 

Currently in your logic the `else` is redundant as your checking if the temp is greater than a number, less than a number and they if it's anything else (which is impossible since the number can only be greater or equal OR less or equal).

services/thermal_mgr/thermal_mgr.c
```diff
@@ -42,19 +43,58 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+            overTemperatureDetected();
+          } else if (tempC <= config.hysteresisThresholdCelsius) { // safe temperature threshold
+            safeOperatingConditions();
+          } else {
+            error_code_t errCode = ERR_CODE_UNKNOWN;
```
Adityya-K: You don't have to declare a variable here, you can just directly `LOG_ERROR_CODE(ERR_CODE_UNKNOWN)`

Adityya-K (changes requested):
You have fixed most things, just a few things remain. You are so close!

services/thermal_mgr/thermal_mgr.c
```diff
@@ -42,19 +43,58 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
-  /* Send an event to the thermal manager queue */
+  // Send an event to the thermal manager queue 
+
+  // if event is null 
+  if (event == NULL) return ERR_CODE_INVALID_ARG;
+
+  if (thermalMgrQueueHandle == NULL) return ERR_CODE_INVALID_QUEUE_MSG;
+
+  // if queue is full return error
+  if (xQueueSend(thermalMgrQueueHandle, event, portMAX_DELAY) == errQUEUE_FULL) {
```
Adityya-K: Here let's not block. Since we have no reason to block when sending, let's keep the delay to a minimum (i.e. 0-10 ms). You can use the `pdMS_TO_TICKS()` macro instead of `portMAX_DELAY`. `pdMS_TO_TICSK()` takes milliseconds as an argument and converts it to the correct amount of ticks for us.

Adityya-K (commented):
The code looks good, just remove the unnecessary files (everything in the `Test/Temporary` directory) and then you are done!

Adityya-K (commented):
LGTM!

PR #244, opened 2025-10-05
OBC-Firmware-Onboarding By Yousef A.

Yousef-Heiba:
### Summary
- Implemented LM75BD temperature reading driver.
- Added thermal management task for event-based telemetry.
- Implemented OS interrupt handler with hysteresis logic (80°C / 75°C).

### Testing
- Verified correct output in provided test environment.
- Telemetry and OS handler output matched expected thresholds.

## Completed Challenges
- [x] LM75BD Driver Challenge

> [!NOTE]

## Testing Screenshots
image


lm75bd/lm75bd.c
```diff
@@ -27,7 +28,27 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
 
 error_code_t readTempLM75BD(uint8_t devAddr, float *temp) {
   /* Implement this driver function */
-  
+  error_code_t errCode;
+  uint8_t tempreg = LM75BD_REG_TEMP; //
+  errCode = i2cSendTo(devAddr, &tempreg, 1);
+  if (errCode != ERR_CODE_SUCCESS) return errCode;
```
camspec: You should also log these error codes, use the `RETURN_IF_ERROR_CODE` macro

lm75bd/lm75bd.h
```diff
@@ -5,7 +5,7 @@
 #include <stdint.h>
 
 /* LM75BD I2C Device Address */
-#define LM75BD_OBC_I2C_ADDR /* Define the address here */
+#define LM75BD_OBC_I2C_ADDR 0x004F/* Define the address here */
```
camspec: This device address is incorrect, did you run unit tests?

lm75bd/lm75bd.c
```diff
@@ -27,7 +28,27 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
 
 error_code_t readTempLM75BD(uint8_t devAddr, float *temp) {
```
camspec, line 28: What if `temp` is `NULL`?

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,19 +43,40 @@ void initThermalSystemManager(lm75bd_config_t *config) {
 
 error_code_t thermalMgrSendEvent(thermal_mgr_event_t *event) {
   /* Send an event to the thermal manager queue */
-
-  return ERR_CODE_SUCCESS;
+  event->type = THERMAL_MGR_EVENT_MEASURE_TEMP_CMD;
```
camspec: You should not change the event type here, since this function should just send whatever event was passed.

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,19 +43,40 @@ void initThermalSystemManager(lm75bd_config_t *config) {
 
 error_code_t thermalMgrSendEvent(thermal_mgr_event_t *event) {
```
camspec, line 46: What if `event` is `NULL`?

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,19 +43,40 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
   /* Send an event to the thermal manager queue */
-
-  return ERR_CODE_SUCCESS;
+  event->type = THERMAL_MGR_EVENT_MEASURE_TEMP_CMD;
+  if (xQueueSend(thermalMgrQueueHandle, event, 0) == pdPASS) {
```
camspec, line 56: What if the queue handle is null?

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,19 +43,40 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+  } else {
+    return ERR_CODE_QUEUE_FULL; 
+  }
 }
 
-void osHandlerLM75BD(void) {
+void osHandlerLM75BD(void) { //function only runs when an interrupt happens
   /* Implement this function */
+  float tempC = 0.0f;
+  if (readTempLM75BD(LM75BD_OBC_I2C_ADDR, &tempC) == ERR_CODE_SUCCESS) {
```
camspec: Notice how you don't use the `thermalMgrSendEvent` function that you made. Since this function runs in an ISR, we want it to execute as quickly as possible and not block, so reading from the temperature register in the os handler is not ideal

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,19 +43,40 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+    if (tempC >= 75.0f) { //if an interrupt happens and the temp is above 75you
```
camspec: This threshold value is hardcoded in the LM75BD config

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,19 +43,40 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
 
 static void thermalMgr(void *pvParameters) {
   /* Implement this task */
+  lm75bd_config_t *config = (lm75bd_config_t *)pvParameters;
+  thermal_mgr_event_t event;
+  float temperatureC = 0.0f;
   while (1) {
-    
-  }
+    if (xQueueReceive(thermalMgrQueueHandle, &event, portMAX_DELAY) == pdTRUE) {
```
camspec, line 79: What if the queue handle is null?

Testing/Temporary/CTestCostData.txt
```diff
@@ -0,0 +1 @@
+---
```
camspec, line 1: Avoid committing other files in the future, since it clutters the history and can make some prs harder to review

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,19 +43,40 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+      if (event.type == THERMAL_MGR_EVENT_MEASURE_TEMP_CMD) {
+        if (readTempLM75BD(config->devAddr, &temperatureC) == ERR_CODE_SUCCESS) {
```
camspec, line 81: If this fails you should log the error code

camspec (changes requested):
Overall well done! There are some comments that need to be addressed

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,7 +45,10 @@ void initThermalSystemManager(lm75bd_config_t *config) {
 
 error_code_t thermalMgrSendEvent(thermal_mgr_event_t *event) {
   /* Send an event to the thermal manager queue */
-  event->type = THERMAL_MGR_EVENT_MEASURE_TEMP_CMD;
+  if (event == NULL || thermalMgrQueueHandle == NULL) {
```
camspec: The queue handle isn't an arg, so you should handle that as an invalid state

services/thermal_mgr/thermal_mgr.c
```diff
@@ -53,28 +58,39 @@ error_code_t thermalMgrSendEvent(thermal_mgr_event_t *event) {
...
-  float tempC = 0.0f;
-  if (readTempLM75BD(LM75BD_OBC_I2C_ADDR, &tempC) == ERR_CODE_SUCCESS) {
-    if (tempC >= 75.0f) { //if an interrupt happens and the temp is above 75you
-      overTemperatureDetected();
-    }  else {
-      safeOperatingConditions();
-    }
-  }
+  thermal_mgr_event_t event;
+  event.type = THERMAL_MGR_EVENT_MEASURE_TEMP_CMD;
```
camspec: This is the event that is called periodically to get temp data from the sensor. The event to send when an os interrupt occurs should be separate from this, so you'll need to make another event (see the onboarding docs for more details)

services/thermal_mgr/thermal_mgr.c
```diff
@@ -53,28 +58,39 @@ error_code_t thermalMgrSendEvent(thermal_mgr_event_t *event) {
...
+  static float lastTempC = 0.0f;
+  if (thermalMgrQueueHandle == NULL) {
+    LOG_ERROR_CODE(ERR_CODE_INVALID_ARG);
+  }
   while (1) {
     if (xQueueReceive(thermalMgrQueueHandle, &event, portMAX_DELAY) == pdTRUE) {
       if (event.type == THERMAL_MGR_EVENT_MEASURE_TEMP_CMD) {
         if (readTempLM75BD(config->devAddr, &temperatureC) == ERR_CODE_SUCCESS) {
           addTemperatureTelemetry(temperatureC);
+          if (temperatureC >= config->overTempThresholdCelsius && lastTempC < config->overTempThresholdCelsius) {
```
camspec: You only want to do this if the interrupt event occurs, not every time we try to measure the temperature. Also, we don't need to track last temp

services/thermal_mgr/thermal_mgr.c
```diff
@@ -53,28 +58,39 @@ error_code_t thermalMgrSendEvent(thermal_mgr_event_t *event) {
...
+  thermalMgrSendEvent(&event);
 }
 
 static void thermalMgr(void *pvParameters) {
   /* Implement this task */
   lm75bd_config_t *config = (lm75bd_config_t *)pvParameters;
   thermal_mgr_event_t event;
   float temperatureC = 0.0f;
+  static float lastTempC = 0.0f;
+  if (thermalMgrQueueHandle == NULL) {
```
camspec: You should add this as a condition inside of the while loop. Right now it would just log it if it was null but then continue the task as normal

services/thermal_mgr/thermal_mgr.c
```diff
@@ -53,28 +58,39 @@ error_code_t thermalMgrSendEvent(thermal_mgr_event_t *event) {
...
+          }
+          lastTempC = temperatureC;
+        } else {
+          LOG_ERROR_CODE(ERR_CODE_INVALID_ARG);
         }
+      } else {
+        LOG_ERROR_CODE(ERR_CODE_INVALID_ARG);
       }
+    } else {
+      LOG_ERROR_CODE(ERR_CODE_INVALID_ARG);
```
camspec: You don't need to log these, because if our queue is empty (in-between events) we will be spammed with logs

camspec (changes requested):
Very good progress, a few more things need to be addressed

camspec (commented):
looks good :)

PR #245, opened 2025-10-07
complete onboarding challenge

EleKtr1X:

## Completed Challenges
- [x] LM75BD Driver Challenge

> [!NOTE]

## Testing Screenshots
image

### DIFF
--- a/services/thermal_mgr/thermal_mgr.c
+++ b/services/thermal_mgr/thermal_mgr.c
@@ -42,28 +42,38 @@ void initThermalSystemManager(lm75bd_config_t *config) {
 }
 
 error_code_t thermalMgrSendEvent(thermal_mgr_event_t *event) {
-  xQueueSend(thermalMgrQueueHandle, (void *) event, (TickType_t) 1);
-
+  if (event == NULL) return ERR_CODE_INVALID_ARG;
+  if (thermalMgrQueueHandle == NULL) return ERR_CODE_INVALID_STATE;
+  if (xQueueSend(thermalMgrQueueHandle, (void *) event, (TickType_t) 1) == errQUEUE_FULL)
+    return ERR_CODE_QUEUE_FULL;
   return ERR_CODE_SUCCESS;
 }
 
 void osHandlerLM75BD(void) {
-  float temp = 0;
-  readTempLM75BD(LM75BD_OBC_I2C_ADDR, &temp);
-  addTemperatureTelemetry(temp);
-  if (temp > LM75BD_DEFAULT_HYST_THRESH) overTemperatureDetected();
-  else safeOperatingConditions();
+  thermal_mgr_event_t event = { .type = THERMAL_MGR_EVENT_OS_INT };
+  thermalMgrSendEvent(&event);
 }
 
 static void thermalMgr(void *pvParameters) {
+  lm75bd_config_t* lm75bd_config = (lm75bd_config_t*) pvParameters;
   while (1) {
-    thermal_mgr_event_t xMessage;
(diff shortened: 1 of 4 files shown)


services/thermal_mgr/thermal_mgr.c
```diff
@@ -42,19 +42,29 @@ void initThermalSystemManager(lm75bd_config_t *config) {
 }
 
 error_code_t thermalMgrSendEvent(thermal_mgr_event_t *event) {
-  /* Send an event to the thermal manager queue */
+  xQueueSend(thermalMgrQueueHandle, (void *) event, (TickType_t) 1);
```
camspec: What if `event` is `NULL`? What about the queue handle?
If the queue is full, you won't be able to send anything to it. Best to return the respective error codes for each of these cases.

services/thermal_mgr/thermal_mgr.c
```diff
@@ -42,19 +42,29 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
 
   return ERR_CODE_SUCCESS;
 }
 
 void osHandlerLM75BD(void) {
-  /* Implement this function */
+  float temp = 0;
+  readTempLM75BD(LM75BD_OBC_I2C_ADDR, &temp);
```
camspec: It is unideal to read from the temperature register while in an ISR since we can be blocked and that can be critical for the system. The os handler should be as short as possible. Check the onboarding docs for implementation hints

services/thermal_mgr/thermal_mgr.c
```diff
@@ -42,19 +42,29 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+  addTemperatureTelemetry(temp);
+  if (temp > LM75BD_DEFAULT_HYST_THRESH) overTemperatureDetected();
```
camspec: As another hint, if you implement the above, you won't need to use a macro here, but you can use the LM75BD config as passed via parameters to the thermal management task

services/thermal_mgr/thermal_mgr.c
```diff
@@ -42,19 +42,29 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+  else safeOperatingConditions();
 }
 
 static void thermalMgr(void *pvParameters) {
-  /* Implement this task */
   while (1) {
-
```
camspec, line 57: If the queue handle is null, we should avoid attempting to receive from the queue

services/thermal_mgr/thermal_mgr.c
```diff
@@ -42,19 +42,29 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
 
 static void thermalMgr(void *pvParameters) {
-  /* Implement this task */
   while (1) {
-    
+    thermal_mgr_event_t xMessage;
+    if (xQueueReceive(thermalMgrQueueHandle, &xMessage, portMAX_DELAY) == pdPASS) {
+      if ((&xMessage)->type == THERMAL_MGR_EVENT_MEASURE_TEMP_CMD) {
+        float temp = 0.0f;
+        readTempLM75BD(LM75BD_OBC_I2C_ADDR, &temp);
```
camspec: What if this read fails? We would want to log the error code. We also don't want to add temperature telemetry if the read wasn't successful, because then we would be writing 0 degrees, which is erroneous data!

camspec (changes requested):
Good job overall, but there are some changes that need to be addressed

EleKtr1X:
alright, it took 3 commits but i think i got everything you asked me to change

camspec (commented):
lgtm, welcome to the team! :)

PR #249, opened 2025-11-09
Selvahini Kamalarajan Firmware Onboarding 

selvxhini-10:

## Completed Challenges
- [✓] LM75BD Driver Challenge
- [✓] Thermal Management Challenge
- [✓] OS Handler Task Challenge

> [!NOTE]

## Testing Screenshots
image


lm75bd/lm75bd.c
```diff
@@ -17,7 +17,7 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
 
   RETURN_IF_ERROR_CODE(writeConfigLM75BD(config->devAddr, config->osFaultQueueSize, config->osPolarity,
                                          config->osOperationMode, config->devOperationMode));
-
+
```
camspec: Like the gs onboarding, try to avoid random formatting changes

lm75bd/lm75bd.c
```diff
@@ -26,8 +26,29 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
 }
 
 error_code_t readTempLM75BD(uint8_t devAddr, float *temp) {
-  /* Implement this driver function */
-  
+   /* Implement this driver function */
+
+  if (temp == NULL) return ERR_CODE_INVALID_ARG;
```
camspec, line 31: Nice error checking!

lm75bd/lm75bd.c
```diff
@@ -26,8 +26,29 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
...
+
+  error_code_t errCode;
+  uint8_t tempRegAddr = 0x00U; // Temperature register address
+  uint8_t tempData[2];
+
+  // Step 1: Send the temperature register address to the device
+  errCode = i2cSendTo(devAddr, &tempRegAddr, 1);
+  if (errCode != ERR_CODE_SUCCESS) return errCode;
```
camspec: You also want to log the error code if either i2c function fails. In this case, you should use one of our pre-defined macros.

lm75bd/lm75bd.h
```diff
@@ -5,7 +5,7 @@
 #include <stdint.h>
 
 /* LM75BD I2C Device Address */
-#define LM75BD_OBC_I2C_ADDR /* Define the address here */
+#define LM75BD_OBC_I2C_ADDR 0b1001111
```
camspec: Correct address, we prefer to define these in hex for brevity, it's also consistent with the other values you see defined in this file

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +43,41 @@ void initThermalSystemManager(lm75bd_config_t *config) {
 
 error_code_t thermalMgrSendEvent(thermal_mgr_event_t *event) {
   /* Send an event to the thermal manager queue */
+  if (event == NULL) return ERR_CODE_INVALID_ARG;
 
+  BaseType_t result = xQueueSend(thermalMgrQueueHandle, event, portMAX_DELAY);
```
camspec: What if the queue handle is null?

camspec: We don't want to use the max timeout here, since that will indefinitely block until the queue send succeeds. Since you call this function in an ISR via the os handler, this scenario could cause our entire satellite to hang (very bad!)

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +43,41 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
 }
 
 static void thermalMgr(void *pvParameters) {
-  /* Implement this task */
+  /* Task Implemented */
+  lm75bd_config_t config = *(lm75bd_config_t *)pvParameters; /* Cast the parameter to the correct type */
+  thermal_mgr_event_t event;
+
   while (1) {
-
```
camspec, line 57: We also want to make sure we don't do anything if the queue handle is null

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +43,41 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+  lm75bd_config_t config = *(lm75bd_config_t *)pvParameters; /* Cast the parameter to the correct type */
+  thermal_mgr_event_t event;
+
   while (1) {
-    
+    if (xQueueReceive(thermalMgrQueueHandle, &event, portMAX_DELAY) == pdPASS) {
+      if (event.type == THERMAL_MGR_EVENT_MEASURE_TEMP_CMD) {
+        /* Measure temperature and handle accordingly */
+        float tempC;
+        readTempLM75BD(config.devAddr, &tempC);
```
camspec: If this temperature read fails (which it can in different ways), we'll be adding invalid temperature telemetry

camspec: We should also log the error code if not success

Testing/Temporary/CTestCostData.txt
```diff
@@ -0,0 +1 @@
+---
```
camspec: More files that shouldn't be committed to the repo

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +43,41 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+        addTemperatureTelemetry(tempC);
+        if (tempC >= config.overTempThresholdCelsius) {
```
camspec: We only want to call either of these functions if an os interrupt event occurs.

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +43,41 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+    return ERR_CODE_QUEUE_FULL;
+  }
   return ERR_CODE_SUCCESS;
 }
 
 void osHandlerLM75BD(void) {
-  /* Implement this function */
+  /* Function Implemented */
+  thermal_mgr_event_t event;
+  event.type = THERMAL_MGR_EVENT_MEASURE_TEMP_CMD;
```
camspec: This is the incorrect event type (this event just tells us that we want to capture temperature telemetry). What actually happens when the os handler is called?

services/thermal_mgr/thermal_mgr.c
camspec: You can just check if the temp at that point is greater than the hysteresis temperature, since the interrupt will only occur at certain points

camspec: If it's greater than the hysteresis temperature, then we know the interrupt must've been called because we went over temp, rather than going back to safe levels

camspec (changes requested):
Very good progress overall! There are some comments that need to be addressed

lm75bd/lm75bd.c
```diff
@@ -26,8 +26,37 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
...
+
+  error_code_t errCode;
+  uint8_t tempRegAddr = 0x00U; // Temperature register address
+  uint8_t tempData[2];
+
+  // Step 1: Send the temperature register address to the device
+  errCode = i2cSendTo(devAddr, &tempRegAddr, 1);
+
+  // Log error code if i2cSendTo fails
+  if (errCode != ERR_CODE_SUCCESS) {
```
camspec: The pre-defined macro `RETURN_IF_ERROR_CODE` achieves this functionality

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,20 +44,63 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
 static void thermalMgr(void *pvParameters) {
-  /* Implement this task */
+  /* Task Implemented */
+  lm75bd_config_t config = *(lm75bd_config_t *)pvParameters; /* Cast the parameter to the correct type */
+  thermal_mgr_event_t event;
+
+  /* Do nothing if queue is NULL*/
+  if (thermalMgrQueueHandle == NULL) {
+    LOG_ERROR_CODE(ERR_CODE_INVALID_STATE);
+    return;
```
camspec: You should not return from a FreeRTOS task; you should let the scheduler take care of it. Exiting prematurely can cause task corruption, memory leaks, etc. and should be avoided. You can make this check within the while loop and either run nothing if the handle is null (which is unideal, as it takes 100% process time for the task) or block the task indefinitely.

camspec (changes requested):
Looks great, just a couple more changes

camspec:
Great work!

PR #250, opened 2025-11-12
Complete firmware onboarding

ColeMeikle:

## Completed Challenges
- [x] LM75BD Driver Challenge

> [!NOTE]

## Testing Screenshots

Screenshot 2025-11-18 at 5 48 32 PM

### DIFF
--- a/lm75bd/lm75bd.c
+++ b/lm75bd/lm75bd.c
@@ -30,8 +30,8 @@ error_code_t readTempLM75BD(uint8_t devAddr, float *temp) {
     if (temp == NULL) return ERR_CODE_INVALID_ARG;
 
     error_code_t errCode;
-    u_int8_t tempRegAddr = 0x00U;
-    u_int8_t buff[2];
+    uint8_t tempRegAddr = 0x00U;
+    uint8_t buff[2];
 
     //Send address of temperature register to sensor
     errCode = i2cSendTo(devAddr, &tempRegAddr, 1);
@@ -42,7 +42,7 @@ error_code_t readTempLM75BD(uint8_t devAddr, float *temp) {
     errCode = i2cReceiveFrom(devAddr, buff, 2);
     if(errCode != ERR_CODE_SUCCESS) return errCode;
 
-    u_int16_t tempNum = (buff[0] << 8) + buff[1];
+    uint16_t tempNum = (buff[0] << 8) + buff[1];
     tempNum >>= 5;//Remove the 5 trailing zeros
 
     if(tempNum & 0x400){
--- a/lm75bd/lm75bd.c
+++ b/lm75bd/lm75bd.c
@@ -26,9 +26,35 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
 }
 
 error_code_t readTempLM75BD(uint8_t devAddr, float *temp) {
-  /* Implement this driver function */
-  
-  return ERR_CODE_SUCCESS;
+    /* Implement this driver function */
+    if (temp == NULL) return ERR_CODE_INVALID_ARG;
+
+    error_code_t errCode;
+    u_int8_t tempRegAddr = 0x00U;
+    u_int8_t buff[2];
+
+    //Send address of temperature register to sensor
+    errCode = i2cSendTo(devAddr, &tempRegAddr, 1);
+
+    if(errCode != ERR_CODE_SUCCESS) return errCode;
+
+    //buff[0] receives the MSB, buff[1] receives the LSB
+    errCode = i2cReceiveFrom(devAddr, buff, 2);
+    if(errCode != ERR_CODE_SUCCESS) return errCode;
+
+    u_int16_t tempNum = (buff[0] << 8) + buff[1];
+    tempNum >>= 5;//Remove the 5 trailing zeros
+
+    if(tempNum & 0x400){
+      tempNum = (~tempNum & 0x7FF) + 1;//Two's complement
+      *temp = tempNum * -0.125f;//Convert to C°
+    }
+    else{
+      //If D10 is 0, temp is positive
+      *temp = tempNum * 0.125f;//Convert to C°
+    }
+    
+    return ERR_CODE_SUCCESS;
 }
 
 #define CONF_WRITE_BUFF_SIZE 2U
--- a/lm75bd/lm75bd.c
+++ b/lm75bd/lm75bd.c
@@ -36,11 +36,11 @@ error_code_t readTempLM75BD(uint8_t devAddr, float *temp) {
     //Send address of temperature register to sensor
     errCode = i2cSendTo(devAddr, &tempRegAddr, 1);
 
-    if(errCode != ERR_CODE_SUCCESS) return errCode;
+    RETURN_IF_ERROR_CODE(errCode);
 
     //buff[0] receives the MSB, buff[1] receives the LSB
     errCode = i2cReceiveFrom(devAddr, buff, 2);
-    if(errCode != ERR_CODE_SUCCESS) return errCode;
+    RETURN_IF_ERROR_CODE(errCode);
 
     uint16_t tempNum = (buff[0] << 8) + buff[1];
     tempNum >>= 5;//Remove the 5 trailing zeros
--- a/lm75bd/lm75bd.h
+++ b/lm75bd/lm75bd.h
@@ -5,7 +5,7 @@
 #include <stdint.h>
 
 /* LM75BD I2C Device Address */
-#define LM75BD_OBC_I2C_ADDR /* Define the address here */
+#define LM75BD_OBC_I2C_ADDR 0x4FU /* Define the address here */
 
 /* LM75BD Configuration Values */
 #define LM75BD_DEV_OP_MODE_NORMAL 0x00U
--- a/services/thermal_mgr/thermal_mgr.c
+++ b/services/thermal_mgr/thermal_mgr.c
@@ -2,6 +2,7 @@
 #include "errors.h"
 #include "lm75bd.h"
 #include "console.h"
+#include "logging.h"
 
 #include <FreeRTOS.h>
 #include <os_task.h>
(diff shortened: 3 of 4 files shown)


lm75bd/lm75bd.c
```diff
@@ -26,9 +26,35 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
 }
 
 error_code_t readTempLM75BD(uint8_t devAddr, float *temp) {
-  /* Implement this driver function */
-  
-  return ERR_CODE_SUCCESS;
+    /* Implement this driver function */
+    if (temp == NULL) return ERR_CODE_INVALID_ARG;
```
camspec, line 30: Nice error checking!

lm75bd/lm75bd.c
```diff
@@ -26,9 +26,35 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
...
+
+    error_code_t errCode;
+    uint8_t tempRegAddr = 0x00U;
+    uint8_t buff[2];
+
+    //Send address of temperature register to sensor
+    errCode = i2cSendTo(devAddr, &tempRegAddr, 1);
+
+    if(errCode != ERR_CODE_SUCCESS) return errCode;
```
camspec: If either I2C function fails, we don't want to just return the error code, we want to log it. There are some pre-defined macros you can use for this purpose; look for one that both logs and returns the error code.

lm75bd/lm75bd.h
```diff
@@ -5,7 +5,7 @@
 #include <stdint.h>
 
 /* LM75BD I2C Device Address */
-#define LM75BD_OBC_I2C_ADDR /* Define the address here */
+#define LM75BD_OBC_I2C_ADDR 0x4FU /* Define the address here */
```
camspec, line 8: Well done on converting to hex!

services/thermal_mgr/thermal_mgr.c
```diff
@@ -42,20 +43,61 @@ void initThermalSystemManager(lm75bd_config_t *config) {
 }
 
 error_code_t thermalMgrSendEvent(thermal_mgr_event_t *event) {
-  /* Send an event to the thermal manager queue */
+  if(event == NULL) return ERR_CODE_INVALID_STATE;
```
camspec: In this case it should be invalid arg, as the event pointer is the arg

services/thermal_mgr/thermal_mgr.c
```diff
@@ -42,20 +43,61 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+
+  if(thermalMgrQueueHandle == NULL) return ERR_CODE_QUEUE_FULL;
```
camspec: If the queue handle is null, it doesn't mean the queue is full, it means that our data is in an invalid state (the queue handle is the pointer to our queue, and should not normally be null)

services/thermal_mgr/thermal_mgr.c
```diff
@@ -42,20 +43,61 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
 
 static void thermalMgr(void *pvParameters) {
-  /* Implement this task */
-  while (1) {
-    
-  }
+    if(pvParameters == NULL){
+      //Unable to read temp data without access to the sensor
+      LOG_ERROR_CODE(ERR_CODE_INVALID_ARG);
+      return;
```
camspec: You shouldn't be returning from a task function, and the reason is because this immediately ends the task prematurely and thus leaks resources. We don't want any task to do this, especially for our satellite.

services/thermal_mgr/thermal_mgr.c
```diff
@@ -42,20 +43,61 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+    }
+    if(thermalMgrQueueHandle == NULL){
```
camspec: Nice error checking, but again we shouldn't be returning.

services/thermal_mgr/thermal_mgr.c
```diff
@@ -42,20 +43,61 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+      //Queue was not succesfully created
+      LOG_ERROR_CODE(ERR_CODE_INVALID_QUEUE_MSG);
+      return;
+    }
+
+    lm75bd_config_t config = *(lm75bd_config_t *) pvParameters;
+    while (1) {
+        thermal_mgr_event_t event;
+        if(xQueueReceive(thermalMgrQueueHandle, &event, 0) == pdPASS){
```
camspec: Setting this timeout to 0 will immediately skip to the next iteration of the loop if the queue is empty. This means that lower priority tasks will be starved, as it effectively uses 100% of our resources running this loop constantly. We should instead block this task while waiting for the queue to receive an event.

services/thermal_mgr/thermal_mgr.c
```diff
@@ -42,20 +43,61 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+              if(errCode == ERR_CODE_SUCCESS){
+                  if(tempC > config.hysteresisThresholdCelsius){
+                    safeOperatingConditions();
+                  }
+              } else{
+                  LOG_ERROR_CODE(errCode);
+              }
+            }
+
+            if(event.type == THERMAL_MGR_EVENT_MEASURE_TEMP_CMD){
```
camspec: This should be an `else if` as no event will have two types at the same time.

services/thermal_mgr/thermal_mgr.c
```diff
@@ -42,20 +43,61 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+              float tempC = 0;
+              error_code_t errCode = readTempLM75BD(config.devAddr, &tempC);
+              if(errCode == ERR_CODE_SUCCESS){
+                  addTemperatureTelemetry(tempC);
+              } else{
+                  LOG_ERROR_CODE(errCode);
+              }
+            }
```
camspec: You should also log the erroneous case where the event received is neither of the above.

services/thermal_mgr/thermal_mgr.c
```diff
@@ -42,20 +43,61 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+
+            if(event.type == THERMAL_MGR_EVENT_MEASURE_TEMP_CMD){
+              float tempC = 0;
+              error_code_t errCode = readTempLM75BD(config.devAddr, &tempC);
+              if(errCode == ERR_CODE_SUCCESS){
+                  addTemperatureTelemetry(tempC);
```
camspec: Nice error checking!

services/thermal_mgr/thermal_mgr.c
```diff
@@ -42,20 +43,61 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+    lm75bd_config_t config = *(lm75bd_config_t *) pvParameters;
+    while (1) {
+        thermal_mgr_event_t event;
+        if(xQueueReceive(thermalMgrQueueHandle, &event, 0) == pdPASS){
+            if(event.type == THERMAL_MGR_EVENT_OS){
+              float tempC = 0;
+              error_code_t errCode = readTempLM75BD(config.devAddr, &tempC);
+              if(errCode == ERR_CODE_SUCCESS){
+                  if(tempC > config.hysteresisThresholdCelsius){
+                    safeOperatingConditions();
```
camspec: I think you mean to call `overTemperatureDetected()` here

services/thermal_mgr/thermal_mgr.h
```diff
@@ -8,6 +8,10 @@ typedef enum {
   
 } thermal_mgr_event_type_t;
 
+typedef enum {
+  THERMAL_MGR_EVENT_OS,
+} thermal_mgr_event_os;
```
camspec: This creates a type mismatch when setting an event's type to os. Add it to the enum above

camspec (changes requested):
Nice work overall! There are some comments that need to be addressed

services/thermal_mgr/thermal_mgr.c
```diff
@@ -42,19 +43,65 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+    /* Implement this function */
+    thermal_mgr_event_t event;
+    event.type = THERMAL_MGR_EVENT_OS;
+    thermalMgrSendEvent(&event);
 }
 
 static void thermalMgr(void *pvParameters) {
-  /* Implement this task */
-  while (1) {
+  if(pvParameters && thermalMgrQueueHandle != NULL){
```
camspec: In practice, FreeRTOS tasks should never exit naturally. Here you never reach the while loop if either are NULL, which can cause undefined behaviour or memory leaks after you log the error code (especially bad for OTA satellite)

camspec: If either are null, don't undergo normal task behaviour, but we should still be running the task loop. Don't log error codes in this case as otherwise they'll spam

camspec (changes requested):
Awesome progress, just one more small issue

services/thermal_mgr/thermal_mgr.c
ColeMeikle: Am I intended to just move the if statement inside of the while loop? I'm concerned that this is just going to waste a bunch of resources if either are null as it just checks the if statement repeatedly.

camspec: Ideally you can use a task delay to effectively block the task indefinitely if it's assumed these issues can't be resolved normally (if a queue handle or task parameters are ever null, it's a critical situation). However, in this case it's okay to have a tight loop because it's much better than exiting the task prematurely (and thus leaking resources, causing task/scheduler corruption, etc. instead of full CPU time at that priority level).

camspec:
Could you also upload an updated screenshot of the test results?

ColeMeikle:
I've blocked the task if there's invalid params and uploaded a new screenshot.

camspec:
great job, congrats on completing the firmware onboarding :)

PR #252, opened 2026-01-08
Firmware Onboarding

Nickolas-Barandich:

## Completed Challenges
- [ ✅] LM75BD Driver Challenge
- [✅ ] Thermal Management Challenge
- [ ✅] OS Handler Task Challenge

> [!NOTE]

## Testing Screenshots
image


services/thermal_mgr/thermal_mgr.c
```diff
@@ -42,19 +43,56 @@ void initThermalSystemManager(lm75bd_config_t *config) {
 }
 
 error_code_t thermalMgrSendEvent(thermal_mgr_event_t *event) {
-  /* Send an event to the thermal manager queue */
-
+  xQueueSend(thermalMgrQueueHandle, event, 0);
```
camspec: Some more error checking needs to be done; what if event is NULL? What about the queue handle? What if the queue is full when we try to send an event?

services/thermal_mgr/thermal_mgr.c
```diff
@@ -42,19 +43,56 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
   return ERR_CODE_SUCCESS;
 }
 
 void osHandlerLM75BD(void) {
-  /* Implement this function */
+  //sends event to thermal manager that there was an interrupt
+  thermal_mgr_event_t osEvent = {.type = OS_HANDLER_INTERRUPT};
+  xQueueSendFromISR(thermalMgrQueueHandle, &osEvent, 0);
```
camspec: If you want to use NULL here it's better to be explicit rather than using the value 0. But ideally, you would request a context switch after this interrupt happens.

services/thermal_mgr/thermal_mgr.c
```diff
@@ -42,19 +43,56 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+  //while loop that runs indefintely checking for events
   while (1) {
-    
+    //checks to see if an event was received
+    if (xQueueReceive(thermalMgrQueueHandle, &event ,portMAX_DELAY) == pdPASS){
+      //reads temperature
+      if(event.type == THERMAL_MGR_EVENT_MEASURE_TEMP_CMD){
+        float temp = 0.0f;
+        LOG_IF_ERROR_CODE(readTempLM75BD(config.devAddr, &temp));
+        addTemperatureTelemetry(temp);
```
camspec: We don't want to add telemetry data if there was an error; the data could be erroneous

services/thermal_mgr/thermal_mgr.c
```diff
@@ -42,19 +43,56 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+      }
+      //reads temperature and checks if overtemperature or safe
+      else if(event.type == OS_HANDLER_INTERRUPT){
+        float temp = 0.0f;
+        LOG_IF_ERROR_CODE(readTempLM75BD(config.devAddr, &temp));
+        if(temp >= config.overTempThresholdCelsius){
```
camspec: We should only check the temperature if we know that the temp we read in is correct. If an error happened we still compare the value to the thresholds.

Testing/Temporary/CTestCostData.txt
```diff
@@ -0,0 +1 @@
+---
```
camspec: Avoid committing extra files

camspec (changes requested):
Looks really good overall! Just a few comments that need to be addressed

camspec (commented):
Looks great, welcome to the team!

PR #259, opened 2026-05-19
Level3

Adrian-kwok:

## Completed Challenges
- [ yep] LM75BD Driver Challenge
- [ yup] Thermal Management Challenge
- [ yes] OS Handler Task Challenge

> [!NOTE]

## Testing Screenshots

### DIFF
--- a/lm75bd/lm75bd.c
+++ b/lm75bd/lm75bd.c
@@ -30,27 +30,38 @@ error_code_t lm75bdInit(lm75bd_config_t* config) {
 error_code_t readTempLM75BD(uint8_t devAddr, float* temp) {
   /* Implement this driver function */
 
-  // select pointer register  (temperature)
-  uint8_t select_temp_reg = 0b00000000;
-  i2cSendTo(devAddr, &select_temp_reg, 1);
-  uint8_t temperature_data[2] = {0, 0};
-  i2cReceiveFrom(devAddr, temperature_data, 2);
-
-  if (temperature_data[0] >> 7 == 0) {  // positive case
-    *temp = temperature_data[0] * (1 << 3);
-    temperature_data[1] = temperature_data[1] >> 5;
-    *temp += temperature_data[1];
-  } else {  // negative case
-    temperature_data[0] = ~temperature_data[0];
-    temperature_data[1] = ~temperature_data[1];
-    temperature_data[1] = temperature_data[1] >> 5;
-    temperature_data[1]++;
-    *temp = -(temperature_data[0] * (1 << 3) + temperature_data[1]);
+  if (temp == NULL) { // check if temp is NULL
+    return ERR_CODE_INVALID_ARG;
   }
 
-  *temp *= LM75BD_TEMP_RESOLUTION;
+  // select pointer register  (temperature)
+  uint8_t select_temp_reg = 0b00000000;
+  error_code_t send_outcome = i2cSendTo(devAddr, &select_temp_reg, 1);
+  if ( send_outcome == ERR_CODE_SUCCESS) {
+    uint8_t temperature_data[2] = {0, 0};
+    error_code_t receive_outcome = i2cReceiveFrom(devAddr, temperature_data, 2);
+    if (receive_outcome == ERR_CODE_SUCCESS) {
+      if (temperature_data[0] >> 7 == 0) {  // positive case
+        *temp = temperature_data[0] * (1 << 3);
+        temperature_data[1] = temperature_data[1] >> 5;
+        *temp += temperature_data[1];
+      } else {  // negative case
+        temperature_data[0] = ~temperature_data[0];
+        temperature_data[1] = ~temperature_data[1];
+        temperature_data[1] = temperature_data[1] >> 5;
+        temperature_data[1]++;
+        *temp = -(temperature_data[0] * (1 << 3) + temperature_data[1]);
+      }
 
-  return ERR_CODE_SUCCESS;
+      *temp *= LM75BD_TEMP_RESOLUTION;
+
+      return ERR_CODE_SUCCESS;
+    } else { // first call suceeded but second failed, probably invalid state
+      return ERR_CODE_INVALID_STATE;
+    }
+  } else { // relay error
+    return send_outcome;
+  }
 }
 
 #define CONF_WRITE_BUFF_SIZE 2U
--- a/lm75bd/lm75bd.c
+++ b/lm75bd/lm75bd.c
@@ -1,22 +1,24 @@
 #include "lm75bd.h"
-#include "i2c_io.h"
-#include "errors.h"
-#include "logging.h"
 
+#include <math.h>
 #include <stdint.h>
 #include <string.h>
-#include <math.h>
+
+#include "errors.h"
+#include "i2c_io.h"
+#include "logging.h"
 
 /* LM75BD Registers (p.8) */
-#define LM75BD_REG_CONF 0x01U  /* Configuration Register (R/W) */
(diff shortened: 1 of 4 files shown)


lm75bd/lm75bd.c
```diff
@@ -1,22 +1,24 @@
 #include "lm75bd.h"
-#include "i2c_io.h"
-#include "errors.h"
-#include "logging.h"
 
+#include <math.h>
```
camspec, line 3: For the future try to avoid formatting existing files

lm75bd/lm75bd.c
```diff
@@ -25,15 +27,36 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
...
-  
+
+  // select pointer register  (temperature)
+  uint8_t select_temp_reg = 0b00000000;
+  i2cSendTo(devAddr, &select_temp_reg, 1);
+  uint8_t temperature_data[2] = {0, 0};
+  i2cReceiveFrom(devAddr, temperature_data, 2);
+
+  if (temperature_data[0] >> 7 == 0) {  // positive case
+    *temp = temperature_data[0] * (1 << 3);
```
camspec: What if temp is NULL?

lm75bd/lm75bd.c
```diff
@@ -25,15 +27,36 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
...
 }
 
-error_code_t readTempLM75BD(uint8_t devAddr, float *temp) {
+error_code_t readTempLM75BD(uint8_t devAddr, float* temp) {
   /* Implement this driver function */
-  
+
+  // select pointer register  (temperature)
+  uint8_t select_temp_reg = 0b00000000;
+  i2cSendTo(devAddr, &select_temp_reg, 1);
```
camspec: What if this function fails (i.e. returns an error code)?

lm75bd/lm75bd.c
```diff
@@ -25,15 +27,36 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
...
+  uint8_t temperature_data[2] = {0, 0};
+  i2cReceiveFrom(devAddr, temperature_data, 2);
```
camspec: This function can also fail, and now the data received is erroneous, but we don't know

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +43,32 @@ void initThermalSystemManager(lm75bd_config_t *config) {
 
 error_code_t thermalMgrSendEvent(thermal_mgr_event_t *event) {
```
camspec: What if the event pointer is NULL?

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +43,32 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
   /* Send an event to the thermal manager queue */
+  if (thermalMgrQueueHandle != 0) {
+    xQueueSend(thermalMgrQueueHandle,event,(TickType_t) 10);
+  }
 
   return ERR_CODE_SUCCESS;
```
camspec: If the queue handle is NULL you still return success, but if the queue handle is NULL something has gone terribly wrong

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +43,32 @@ void initThermalSystemManager(lm75bd_config_t *config) {
 
 error_code_t thermalMgrSendEvent(thermal_mgr_event_t *event) {
   /* Send an event to the thermal manager queue */
+  if (thermalMgrQueueHandle != 0) {
```
camspec: To be explicit, compare with NULL

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +43,32 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+    xQueueSend(thermalMgrQueueHandle,event,(TickType_t) 10);
+  }
 
   return ERR_CODE_SUCCESS;
 }
 
 void osHandlerLM75BD(void) {
   /* Implement this function */
+  float temp;
+  readTempLM75BD(LM75BD_OBC_I2C_ADDR,&temp);
```
camspec: We do not want to call driver functions within an ISR, there are quite a few reasons why we don't want to, but short answer is that they can potentially block, error out, and are lengthy while we want to keep ISRs as short as possible
Also, if this were to fail, the temp value we would use would be wrong

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +43,32 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+  if (temp >= LM75BD_DEFAULT_HYST_THRESH) {
+    overTemperatureDetected();
```
camspec: This check shouldn't be in the ISR handler, it should be handled by the task itself

services/thermal_mgr/thermal_mgr.c
camspec: You want to use the value set in the device config rather than the macro here, but this function shouldn't be used in the handler at all

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +43,32 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+    safeOperatingConditions();
+  }
 }
 
 static void thermalMgr(void *pvParameters) {
   /* Implement this task */
+  float temp;
   while (1) {
-    
+    if (pdTRUE == xQueueReceive(thermalMgrQueueHandle,thermalMgrQueueStorageArea,(TickType_t) 10)) {
```
camspec: Delay of 10 ticks is arbitrary, ideally we would stall the task for as long as possible while an event is not in the queue. Also, a hint is that the type of the event matters
The function call should be on the left of the equality for consistency

services/thermal_mgr/thermal_mgr.c
```diff
@@ -43,18 +43,32 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+      readTempLM75BD(LM75BD_OBC_I2C_ADDR, &temp);
```
camspec: If this driver function fails, we don't know that it did (no logging), and then we add erroneous telemetry

services/thermal_mgr/thermal_mgr.c
camspec: We don't check if the queue handle is null within the task itself

camspec: You're writing to the queue storage area, which is the queue's internal storage to avoid dynamic allocation, writing there corrupts elements in the queue

camspec (changes requested):
Great progress overall, but there are some issues that need to be addressed

lm75bd/lm75bd.c
```diff
@@ -25,15 +27,47 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
...
+        temperature_data[1] = temperature_data[1] >> 5;
+        temperature_data[1]++;
+        *temp = -(temperature_data[0] * (1 << 3) + temperature_data[1]);
+      }
+
+      *temp *= LM75BD_TEMP_RESOLUTION;
+
+      return ERR_CODE_SUCCESS;
+    } else { // first call suceeded but second failed, probably invalid state
+      return ERR_CODE_INVALID_STATE;
```
camspec, line 60: You should also relay the second error if it occurs from the receive function

camspec (commented):
Great work, welcome to the team!

PR #264, opened 2026-08-17
Completed Challenges

sugo14:

## Completed Challenges
- [✓] LM75BD Driver Challenge
- [✓] Thermal Management Challenge
- [✓] OS Handler Task Challenge

> [!NOTE]

## Testing Screenshots
Screenshot from 2026-08-16 19-03-27

### DIFF
--- a/lm75bd/lm75bd.c
+++ b/lm75bd/lm75bd.c
@@ -34,7 +34,11 @@ error_code_t readTempLM75BD(uint8_t devAddr, float *temp) {
   uint8_t tempBuf[2];
   RETURN_IF_ERROR_CODE(i2cReceiveFrom(devAddr, tempBuf, 2));
 
-  int16_t tempIntermediate = tempBuf[0] << 3 && tempBuf[1] >> 5;
+  // convert from 11 bit fixed point to float
+  int16_t tempIntermediate = ((int16_t)tempBuf[0] & 0b01111111) << 3 | tempBuf[1] >> 5;
+  if (tempBuf[0] & 0b10000000) {
+    tempIntermediate = tempIntermediate - 0b0000010000000000;
+  }
   *temp = (float)tempIntermediate / 8.0;
   
   return ERR_CODE_SUCCESS;
--- a/lm75bd/lm75bd.c
+++ b/lm75bd/lm75bd.c
@@ -26,7 +26,16 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
 }
 
 error_code_t readTempLM75BD(uint8_t devAddr, float *temp) {
-  /* Implement this driver function */
+  error_code_t errCode;
+
+  uint8_t ptrByte = 0b00000000;
+  RETURN_IF_ERROR_CODE(i2cSendTo(devAddr, &ptrByte, 1)); // ? should i be using the macro here
+
+  uint8_t tempBuf[2];
+  RETURN_IF_ERROR_CODE(i2cReceiveFrom(devAddr, tempBuf, 2));
+
+  int16_t tempIntermediate = tempBuf[0] << 3 && tempBuf[1] >> 5;
+  *temp = (float)tempIntermediate / 8.0;
   
   return ERR_CODE_SUCCESS;
 }
--- a/services/thermal_mgr/thermal_mgr.c
+++ b/services/thermal_mgr/thermal_mgr.c
@@ -42,19 +42,37 @@ void initThermalSystemManager(lm75bd_config_t *config) {
 }
 
 error_code_t thermalMgrSendEvent(thermal_mgr_event_t *event) {
-  /* Send an event to the thermal manager queue */
+  xQueueSend(thermalMgrQueueHandle, event, 1000); // ? ticks to wait
 
   return ERR_CODE_SUCCESS;
 }
 
+// this has to be non blocking i think so im not running the reading thing within this?
+// not sure if this actually makes sense
 void osHandlerLM75BD(void) {
-  /* Implement this function */
+  thermalMgrSendEvent(THERMAL_MGR_EVENT_OS);
 }
 
 static void thermalMgr(void *pvParameters) {
-  /* Implement this task */
+  lm75bd_config_t data = *(lm75bd_config_t *) pvParameters;
+  thermal_mgr_event_t event;
+  float temp;
+
   while (1) {
-    
+    if (xQueueReceive(thermalMgrQueueHandle, &event, 10)) { // ? how many ticks should i wait? idk
+      readTempLM75BD(data.devAddr, &temp); // ? error handling?
+      if (event.type == THERMAL_MGR_EVENT_MEASURE_TEMP_CMD) {
+        addTemperatureTelemetry(temp);
+      }
+      else if (event.type == THERMAL_MGR_EVENT_OS) {
+        if (temp > data.hysteresisThresholdCelsius) {
+          overTemperatureDetected();
+        }
+        else {
+          safeOperatingConditions();
+        }
+      }
+    }
   }
 }
 
--- a/lm75bd/lm75bd.h
+++ b/lm75bd/lm75bd.h
@@ -5,7 +5,7 @@
 #include <stdint.h>
 
 /* LM75BD I2C Device Address */
-#define LM75BD_OBC_I2C_ADDR /* Define the address here */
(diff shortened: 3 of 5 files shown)


lm75bd/lm75bd.c
```diff
@@ -26,7 +26,20 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
 }
 
 error_code_t readTempLM75BD(uint8_t devAddr, float *temp) {
-  /* Implement this driver function */
+  error_code_t errCode;
+
+  uint8_t ptrByte = 0b00000000;
```
kepler452b123: Our convention is to use hex to represent the value throughout the codebase

kepler452b123: also just write 0, no need to write the full byte

lm75bd/lm75bd.c
```diff
@@ -26,7 +26,20 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
...
+  RETURN_IF_ERROR_CODE(i2cSendTo(devAddr, &ptrByte, 1)); // ? should i be using the macro here
+
+  uint8_t tempBuf[2];
```
kepler452b123: it's good practice to initialize arrays to 0, otherwise the memory is just whatever was in the location before (garbage)

services/thermal_mgr/thermal_mgr.c
```diff
@@ -42,19 +42,37 @@ void initThermalSystemManager(lm75bd_config_t *config) {
 }
 
 error_code_t thermalMgrSendEvent(thermal_mgr_event_t *event) {
-  /* Send an event to the thermal manager queue */
+  xQueueSend(thermalMgrQueueHandle, event, 1000); // ? ticks to wait
```
kepler452b123: do a time of 0 as this is called from an ISR

services/thermal_mgr/thermal_mgr.c
```diff
@@ -42,19 +42,37 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
```
kepler452b123: also always check if pointer arguments are NULL (if they are then return invalid argument error). dereferencing a NULL pointer is dangerous

services/thermal_mgr/thermal_mgr.c
kepler452b123: error handling: check if the xQueueSend passes successfully. if not, return an error code

services/thermal_mgr/thermal_mgr.c
```diff
@@ -42,19 +42,37 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
 
 static void thermalMgr(void *pvParameters) {
-  /* Implement this task */
+  lm75bd_config_t data = *(lm75bd_config_t *) pvParameters;
+  thermal_mgr_event_t event;
+  float temp;
+
   while (1) {
-    
+    if (xQueueReceive(thermalMgrQueueHandle, &event, 10)) { // ? how many ticks should i wait? idk
```
kepler452b123: we want this task to block until an event is in the queue, so use the max delay time possible (portMAX_delay)

services/thermal_mgr/thermal_mgr.c
```diff
@@ -42,19 +42,37 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+      readTempLM75BD(data.devAddr, &temp); // ? error handling?
```
kepler452b123: log if error code here. don't continue with the logic if the driver function fails

services/thermal_mgr/thermal_mgr.c
```diff
@@ -42,19 +42,37 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+      if (event.type == THERMAL_MGR_EVENT_MEASURE_TEMP_CMD) {
+        addTemperatureTelemetry(temp);
+      }
+      else if (event.type == THERMAL_MGR_EVENT_OS) {
+        if (temp > data.hysteresisThresholdCelsius) {
+          overTemperatureDetected();
+        }
+        else {
+          safeOperatingConditions();
```
kepler452b123: nit: it's possible that the error code is nether one of these, also the event is an enum so it's cleaner to use a switch statement here instead of an if block

services/thermal_mgr/thermal_mgr.c
```diff
@@ -42,19 +42,37 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
 void osHandlerLM75BD(void) {
-  /* Implement this function */
+  thermalMgrSendEvent(THERMAL_MGR_EVENT_OS);
 }
 
 static void thermalMgr(void *pvParameters) {
-  /* Implement this task */
+  lm75bd_config_t data = *(lm75bd_config_t *) pvParameters;
+  thermal_mgr_event_t event;
+  float temp;
```
kepler452b123: initialize this to 0

kepler452b123 (changes requested):
Overall good, some fixes

lm75bd/lm75bd.c
```diff
@@ -26,7 +26,20 @@ error_code_t lm75bdInit(lm75bd_config_t *config) {
 }
 
 error_code_t readTempLM75BD(uint8_t devAddr, float *temp) {
```
panthpatel2016: we usually want to check parameters like `temp` which are pointers to make sure they are not `NULL` to prevent against segfaults

services/thermal_mgr/thermal_mgr.c
```diff
@@ -42,19 +42,37 @@ void initThermalSystemManager(lm75bd_config_t *config) {
 }
 
 error_code_t thermalMgrSendEvent(thermal_mgr_event_t *event) {
```
panthpatel2016: same thing here - should check if `event` is `NULL` to prevent against segfaults

services/thermal_mgr/thermal_mgr.c
panthpatel2016: as we are sending an event to a queue, it should be non blocking so we can just do `0` here. Also if we do ever decide to use a blocking delay, we should use the `pdMS_TO_TICKS()` ticks macro.

services/thermal_mgr/thermal_mgr.c
```diff
@@ -42,19 +42,37 @@ void initThermalSystemManager(lm75bd_config_t *config) {
...
+  xQueueSend(thermalMgrQueueHandle, event, 1000); // ? ticks to wait
 
   return ERR_CODE_SUCCESS;
 }
 
+// this has to be non blocking i think so im not running the reading thing within this?
+// not sure if this actually makes sense
 void osHandlerLM75BD(void) {
-  /* Implement this function */
+  thermalMgrSendEvent(THERMAL_MGR_EVENT_OS);
```
panthpatel2016: You are passing the `THERMAL_MGR_EVENT_OS `enum value directly into a function that expects a pointer to a struct. Instead declare a local event variable and set the type of the event to be `THERMAL_MGR_EVENT_OS` and then pass its address

panthpatel2016 (changes requested):
Initial Review Looks great - left some comments that need to be addressed before moving further!

services/thermal_mgr/thermal_mgr.c
panthpatel2016: more on this comment - after looking at the build it looks like you crashed because what happened is that the function was expected a pointer (but instead an enum of 1 was passed) so it tried dereferencing the address 0x1 which makes it essentially a null pointer as there could be anything at that memory location.

panthpatel2016: Image

panthpatel2016: image

panthpatel2016: command dumped core means that we crashed runtime when tried to access memory that shouldnt be accessed causing it to hardfault

sugo14:
afaik i've addressed all the comments, lmk if theres anything else!

services/thermal_mgr/thermal_mgr.c
sugo14: smh i didnt even notice this. im confused about how my code continued to pass the tests despite this this seems pretty major

kepler452b123 (commented):
LGTM

